PR #3, opened 2017-10-12
Adding Search Page and Buy Page

wanderingstan:
Search page is minimal but functioning. Buy page (and buy contract) still under development.


README.md
```diff
@@ -41,38 +41,41 @@ If you're new to the space, it may be helpful to first familiarize yourself with
...
-```
-testrpc
-```
-  This will output credentials for several test wallets with 100 Eth.
-  
-7. In [Metamask](https://metamask.io/), configure RPC to be private network (localhost 8545) and import the private key for one of your test wallets from previous step. **DO NOT GET YOUR MAIN NET WALLET MIXED UP**.
-  * This needs to be done every single time a new testrpc instance is run.
-  * After awhile, your Metamask will be cluttered with invalid addresses. [Here is how you clear it](https://ethereum.stackexchange.com/questions/21422/how-to-remove-unused-test-accounts-from-metamask/21468)
-  * While we dev, I would actually not recommend having real Eth in a Main Net wallet so we can dump it whenever
+
```
UnicycleMatt: Not sure we should state the expected private key. For example, I use a different mnemonic.

contracts/Listing.sol
```diff
@@ -1,35 +1,54 @@
...
   function Listing() {
     origin = msg.sender;
   }
 
-  // Create a new listing
-  function create(string ipfsHash) {
-    // Check for IPFS hash length
-    
-    require(bytes(ipfsHash).length == 46);
+  struct listingStruct {
```
UnicycleMatt: Add note that this is an assumption that may change in the future, but we're doing it this way to conserve space/computation costs for now.

contracts/Listing.sol
```diff
@@ -1,35 +1,54 @@
...
+      listings[index].ipfsHash,
+      listings[index].price,
+      listings[index].unitsAvailable
+    );
   }
-  
-  // Change this to return multiple listings later
-  function listingForAddress(address _account) public constant returns (string) {
-    return listings[_account];
+
```
UnicycleMatt: We haven't set a standard on how to organize smart contracts, but we should probably have some sort of hierarchy for public/private variables, functions, and events. Perhaps move all events above functions rather than interspersing them amongst more complicated code. Thoughts on this?

README.md
joshfraser: I actually like including the expected private key. How about we compromise and add: 
"if you used '0rigin Demo' as your mnemonic."

src/components/demo-app.js
```diff
@@ -1,5 +1,6 @@
 import React, { Component } from 'react'
 
+import DemoStep0 from './demo-step-0'
```
UnicycleMatt: Totally a function of my poor way of denoting steps here, but we should be consistent about either numbers or descriptive names for these steps. I think either is fine for now.

src/components/demo-step-0.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    let that = this
+
+    // Test getting listings from chain
+    setTimeout(function() {
+      // TODO: Remove hacky 2s delay and correctly determine when contractService
+      // is ready
+      contractService.getAllListings().then((allContractResults) => {
+        var resultIndex;
```
UnicycleMatt: Should this be a let?

src/components/demo-step-0.js
```diff
@@ -0,0 +1,107 @@
...
+                'ipfs': JSON.parse(listingJson).data
+              }
+              console.log(listingData)
+              // Append our new result to state. For now we don't care about ordering.
+              that.setState({
+                listingsResults: that.state.listingsResults.concat(listingData)
+              })
+            })
+            .catch((error) => {
+              alert(error)
```
UnicycleMatt: Fine for now. Future request for the future is some nicer dialog or toast error message. :)

src/components/demo-step-0.js
```diff
@@ -0,0 +1,107 @@
...
+    this.state = {
+      listingsResults: [],
+      contractListingsCount: -1
+    }
+
+    let that = this
+
+    // Test getting listings from chain
+    setTimeout(function() {
+      // TODO: Remove hacky 2s delay and correctly determine when contractService
```
UnicycleMatt: Yea this is fine for now. 

Borrowing from Goldman's modification of the standard TODO, but can we put TODO (name of engineer): .... so that in the future it's really easy for any given engineer to search for their own TODOs.

src/components/demo-step-0.js
```diff
@@ -0,0 +1,107 @@
...
+          }}>
+            Create Listing
+          </button>
+        </div>
+
+         <div>
+            <img
+              src='ajax-loader.gif'
+              style={{
+                display: ((this.state.listingsResults.length == 0 && this.state.listingsResults.contractListingsCount != 0) ||
```
UnicycleMatt: When would contractListingsCount be less than 0?

src/components/demo-step-0.js
```diff
@@ -0,0 +1,107 @@
...
+            {this.state.listingsResults.map(listing => (
+              <div className="listing" key={listing.contract.ipfsHash}>
+                <hr/>
+                <h3>{listing.ipfs.name}</h3>
+                <img
+                  height="200"
+                  src={
+                    (listing.ipfs.pictures && listing.ipfs.pictures.length>0) ?
+                    listing.ipfs.pictures[0] :
+                    'http://www.lackuna.com/wp-content/themes/fearless/images/missing-image-640x360.png'
```
UnicycleMatt: Nitpick but let's put this in the images folder so we don't have to link to this site.

src/components/demo-step-0.js
```diff
@@ -0,0 +1,107 @@
...
+                'block' : 'none'
+              }}
+            />
+            {(this.state.contractListingsCount == 0) ? "No listings" : ""}
+            {this.state.listingsResults.map(listing => (
+              <div className="listing" key={listing.contract.ipfsHash}>
+                <hr/>
+                <h3>{listing.ipfs.name}</h3>
+                <img
+                  height="200"
```
UnicycleMatt: Add a class to the image and CSS vs. inline height please. I can style it later.

src/services/contract-service.js
```diff
@@ -39,15 +63,56 @@ class ContractService {
       this.listingContract.setProvider(web3Service.web3.currentProvider)
       web3Service.web3.eth.getAccounts((error, accounts) => {
         this.listingContract.deployed().then((instance) => {
-          return instance.create(ipfsListing, {from: accounts[0]})
+          return instance.create(this.getBytes32FromIpfsHash(ipfsListing), 1.0, 1.0, {from: accounts[0]})
```
UnicycleMatt: Add a comment that we're inserting defaults of 1.0 for now.

src/services/contract-service.js
```diff
@@ -39,15 +63,56 @@ class ContractService {
...
+            let listings = []
+            let getListingPromises = []
+            // TODO: Paging over listings to get only subsets
+            for (var i = 0; i < listingsLength; i++) {
+              getListingPromises[i] = new Promise((resolve) => {
+                instance.getListing.call(i).then((listing)  => {
+                  // Listing is returned as array of properties.
+                  // IPFS hash (as bytes32 hex string) is in results[2]
+                  // Convert it to regular IPFS base-58 encoded hash
+                  var listingObject = {
```
UnicycleMatt: const?

src/services/contract-service.js
```diff
@@ -39,15 +63,56 @@ class ContractService {
...
+    return new Promise((resolve, reject) => {
+      this.listingContract.setProvider(web3Service.web3.currentProvider)
+        this.listingContract.deployed().then((instance) => {
+          // Get total number of listings
+          instance.listingsLength.call().then((listingsLength) => {
+            let that = this
+            let listings = []
+            let getListingPromises = []
+            // TODO: Paging over listings to get only subsets
+            for (var i = 0; i < listingsLength; i++) {
```
UnicycleMatt: let?

src/services/ipfs-service.js
```diff
@@ -31,22 +37,46 @@ class IpfsService {
...
+  getListing(ipfsHashStr) {
+    return new Promise((resolve, reject) => {
+
+      this.ipfs.files.cat(ipfsHashStr, function (err, stream) {
+        if (err) {
+          console.error(err)
+          reject("Got ipfs cat err:" + err)
+        }
+
+        var res = ''
```
UnicycleMatt: const?

contracts/Listing.sol
```diff
@@ -1,35 +1,54 @@
...
+  listingStruct[] public listings;
+
+  // Return number of listings
+  function listingsLength() public constant returns (uint) {
+      return listings.length;
+  }
+
+  // Return listing info
+  function getListing(uint index) public constant returns (uint, address, bytes32, uint, uint) {
+    // TODO: Determine if less gas to do one array lookup into var, and return var struct parts
```
UnicycleMatt: Same comment about TODO

contracts/Listing.sol
wanderingstan: I could go either way, no strong opinion. If events end up pretty tightly bound to functions, then could make sense to define them together. If events are getting triggered from multiple places, would make sense to group them at the top.

src/components/demo-app.js
wanderingstan: Totally. The numbered steps made sense for the (literally) step-by-step demo, but I think best to re-write to function names. Or (for now) function + step, e.g. `create-listing-step-1`.

src/components/demo-step-0.js
wanderingstan: YES!

src/components/demo-step-0.js
wanderingstan: It's initialized to -1, indicating that haven't gotten a count returned from the contract yet. Maybe better to go with `undefined` or a separate `isContractInfoLoaded` bool.

src/components/demo-step-0.js
wanderingstan: Agree! This was a 5pm friday "make it work" addition. :)

src/components/demo-step-0.js
wanderingstan: Yeah, I should make a quick similar-ish image in photoshop to make it license-free.

src/services/contract-service.js
wanderingstan: YES! Damn my pre-ES6 muscle memory.

src/services/contract-service.js
wanderingstan: always yes

src/services/ipfs-service.js
wanderingstan: yes again

src/components/demo-step-0.js
wanderingstan: ..in my defense, realized they actually use `var` in the official docs.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in
I'm changing it to `.forEach()` which, also solves the problem of keeping reference to the result object. Cleaner code all around!
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

src/components/demo-step-0.js
wanderingstan: toast

src/services/ipfs-service.js
wanderingstan: (actually needs to be `let`, as we append to it as chunks come in from ipfs.)

wanderingstan:
Going to re-submit pull request with these fixes.

PR #8, opened 2017-11-15
Add ipfs daemon steps to readme

royalharsh:


joshfraser:
Whoops. Good catch @royalharsh

PR #42, opened 2018-01-07
Listing contract and ipfs refactor

kkomaz:
- Related to https://github.com/OriginProtocol/demo-dapp/issues/39
- Not exactly sure if this is the type of refactoring you were envisioning.  I can adjust as needed 😄

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,16 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "JSONStream": {
-      "version": "1.3.2",
-      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.2.tgz",
-      "integrity": "sha1-wQI3G27Dp887hHygDCC7D85Mbeo=",
-      "dev": true,
-      "requires": {
-        "jsonparse": "1.3.1",
-        "through": "2.3.8"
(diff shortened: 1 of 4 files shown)


wanderingstan:
Hey @kkomaz that's looking good.

Feels a little weird to pass in a `self` object, and require a specific structure of props (has to have self.props.listingId). This requires `origin-service.js` to know about reactjs, and I feel it should be more generic. 

Just thinking out loud about a possible better way: Maybe `getListing()` accepts a listingID, and returns two promises: one for the contract data, and one for the ipfs data. Then `listing-card` and `listing-detail` do the work of taking the results of those promises and putting them into `setState()`

(Not a critique of your PR: I'm realizing now that its a trickier thing to do "right" than on first glance.)

kkomaz:
Yeah definitely agree with that one.  It felt very weird passing the object of `this` in that way and handling all the `setState` logic in the function itself.  Based on the original issue I wasn't sure to what extent you wanted to abstract that away from the component level.  Will take another stab at it.

kkomaz:
@wanderingstan 

Maybe another attempt based on our conversation.  Let me know what you think.

nick:
I'm going to close this as the demo-dapp has gone through some fairly major changes since January and these changes don't really apply anymore.

Issue #24, opened 2017-12-17
Add support for loading sub-schemas by URL to the React JSONSchema library


joshfraser:
We're using this awesome React library to load & validate our JSONSchema forms:
https://github.com/mozilla-services/react-jsonschema-form

Unfortunately, this library does not currently support loading sub-schemas by URL. This functionality is supported in the JSONSchema spec but it's not in the React implementation that we're using. We're going to need this to be able to have schemas inherit common reusable components like mailing addresses or phone numbers.

Read up on $refs and the id properties of JSONSchema for more context:
https://spacetelescope.github.io/understanding-json-schema/structuring.html

garrisons:
So it seems like the way we would want to implement a solution is by making a PR for the `react-jsonschema-form` framework? Or is there an easier way to implement a solution that I'm missing?

joshfraser:
we should do some outreach to their team & see how receptive they would be

garrisons:
Yeah that seems like a no go. I missed the attached screenshot from first look. I have some ideas on how to make it work though.

![screen shot 2018-01-22 at 7 12 27 pm](https://user-images.githubusercontent.com/5100597/35252929-54d7e98a-ffa8-11e7-9acd-194c3e14cbd6.png)

epicfaace:
@joshfraser @garrisons I wouldn't mind a PR for this functionality, especially given that it is supported on the JSON Schema spec.

It would be related to https://github.com/mozilla-services/react-jsonschema-form/issues/323

micahalcorn:
Closing this as we no longer use [react-jsonschema-form](https://github.com/mozilla-services/react-jsonschema-form) 😐

PR #68, opened 2018-01-23
Issue #24 - Reusable Schema Definitions

garrisons:
This is in response to issue #24. 

So I did some research around if Mozilla would add in support for fetching sub schemas by url and discovered that they won't. They mention it very explicitly, but kind of hard to find in their readme [here](https://github.com/mozilla-services/react-jsonschema-form).

So then I started to look at other ways for us to accomplish our wildest dreams of having reusable schemas and came up with something a little different. I removed the fetch completely, because it seemed clunky to me and moved the form schemas inside the react app as opposed to in the public folder.

I understand this is a pretty big change and that this is a new eco-system for me so please give me feedback on this! Especially @wanderingstan as I think I'm re-working something that you put together and it may just be personal preference and in which case I can change it back in a heart beat!

Cheers!

### DIFF
--- a/src/components/listing-create.js
+++ b/src/components/listing-create.js
@@ -6,6 +6,8 @@ import contractService from '../services/contract-service'
 import ListingDetail from './listing-detail'
 import Form from 'react-jsonschema-form'
 import Overlay from './overlay'
+import _ from 'lodash'
+import Schemas from '../schemas/index'
 
 class ListingCreate extends Component {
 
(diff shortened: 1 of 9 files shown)


micahalcorn:
Thanks @garrisons, and sorry it's been so long. We are going to be discussing a proposal for our schemas this Wednesday during [our open engineering call](https://www.originprotocol.com/en/developers#call). We would love for you to join us!

Issue #72, opened 2018-01-31
Handle scrolling correctly


wanderingstan:
Right now when you move between "pages" of the DApp, your scrolling position is left the same even when it would make more sense to move to the top. 

E.g. you scroll to bottom of grid page and then click on listing, you are then taken to the BOTTOM of the listing page. You would expect to be at the top.

wanderingstan:
Fixed in a005f6355f2fe656cf0bfc0779d120fb4142f3b4
Thanks @endlist !

Issue #64, opened 2018-01-10
Add paging to URLs / Router


wanderingstan:
Back button behavior is broken on `listings-grid.js` (homepage) because we don't change the URL. 

We need to use react-router `Route` and `Link` to do it right. I think the right way is: 
`/` = homepage
`/page/1` = page 1
`/page/2` = page 2
etc...
Where the page corresponds with `activePage` from `Pagination` component.

wanderingstan:
@tyleryasaka did this ages ago. Closing.

PR #76, opened 2018-02-02
add ScrollToTop component for routing

endlist:
Fix for #72 Handle scrolling correctly will always scroll to top when there is a routing page change.

Does not fix scroll to top on the grid, as that requires the routing changes in #64 Add paging to URLs / Router that I am looking into separately.

### DIFF
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -6,6 +6,7 @@ import {
 import { Web3Provider } from 'react-web3'
 
 // Components
+import ScrollToTop from './scroll-to-top.js';
 import Listings from './listings-grid.js'
 import ListingDetail from './listing-detail.js'
 import ListingCreate from './listing-create.js'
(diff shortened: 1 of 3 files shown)


wanderingstan:
This approach works, but seems overly aggressive in cases like selecting type of listing. E.g. changing from 'Announcements' to 'For Sale' also triggers a scroll to top. 
http://localhost:3000/create

Is there a way we could only to the page scroll when the url/route changes? (I'll look into this too)

endlist:
Hm, it really should only be scrolling on url/route changes.  It's definitely not triggering for the grid without additional changes to make routing happen.  I'll look into it further.

wanderingstan:
Ah, that old `window.scrollTo(0, 0)` was still in there!

Looks good, I'll merge. 

As someone new to React, how does your `ScrollToTop` component limit itself to url/route changes in this line? 
```
    if (this.props.location !== prevProps.location) {
```

endlist:
React's router seems to pass along `location` (including the pathname) in the props, and the previous properties can be accessed through `componentDidUpdate` which seems to be a lifecycle hook triggered on render changes.  So you can check if the current page location data matches the previous and act if it matches or doesn't as suits the need.  In this case I only want to scroll if the pathname changed which meant we routed to a new page, not just rerendered something on the page.

wanderingstan:
> Oh, I see now why that was probably included originally was to bump the page up when going from one view to the next inside Create-- should there be subroutes there (create/category vs create/details or something) or should I make it so the component just scrolls to top when the view changes?

And now I'm just noticing your comment, and realizing that we just "broke" scroll to top within the creation steps. :)

I'm open to ideas here. Subroutes would work, though it would allow links to directly to a later step. (Would we just kick them back to the start?) Or if there's a way to use the `window.scrollTo(0, 0)` in `render()` but limit it to when `state.STEP` changes?

endlist:
`window.scrollTo(0, 0)` works pretty much anywhere in the component, so I was thinking we could just put it after the relevant `setState` calls if you don't want to make them actual routes.  I tested that with the category > detail and it seemed to work, so I could just add another for the preview step.

PR #82, opened 2018-02-09
issue/81 use a sticky footer

### LINKED ISSUE #81: Use a sticky footer
Issue is particularly noticeable when you load the page and there are 0 listings & especially if you are on a larger monitor.

micahalcorn:
This resolves #81. It also made sense to add a shared layout to the components in an effort to preserve the footer implementation. This involved adding an extra wrapper div or two, but that is common in React projects. The commit also adds the new partners link to the footer.


wanderingstan:
For others curious what a "Sticky footer is": https://css-tricks.com/couple-takes-sticky-footer/

PR #88, opened 2018-02-23
json updates to 6 files in public/schemas

nsimon:
1) Updated $schema":"http://json-schema.org/draft-04/schema#" to 
"$schema":"http://json-schema.org/draft-06/schema#"

2) Changed types on all min/max field values from string to integer.

Original draft-04 was throwing an error in my linux Firefox (v58.0.2) dev console:
"Error: no schema with key or ref".

I changed the source json $schema values to use draft-06, and then saw a more useful error message. Can't recall the exact error syntax, but it referred to all of the min/max fields being invalid string types, which I then fixed to be integers. That solved the issue.

I suspect the original draft-04 schemas were working in other os/browser environments, due to a less stringent JVM running in those browsers. Perhaps the linux Firefox JVM was less forgiving?

Neil

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -106,10 +106,9 @@
       "dev": true
     },
     "ajv": {
-      "version": "4.11.8",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.8.tgz",
-      "integrity": "sha1-gv+wKynmYq5TvcIK8VlHcGc5xTY=",
-      "dev": true,
+      "version": "4.11.7",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.7.tgz",
+      "integrity": "sha1-hlWl2G0IJJhcxHGh2RP7Zymg7Eg=",
       "requires": {
         "co": "4.6.0",
         "json-stable-stringify": "1.0.1"
@@ -2128,7 +2127,7 @@
       "requires": {
         "anymatch": "1.3.2",
         "async-each": "1.0.1",
-        "fsevents": "1.1.2",
+        "fsevents": "1.1.3",
         "glob-parent": "2.0.0",
         "inherits": "2.0.3",
         "is-binary-path": "1.0.1",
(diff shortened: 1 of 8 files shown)


nsimon:
More (possibly helpful) info.

"Error: no schema with key or ref" -- was thrown when I clicked the Continue button to add an item.

ryana:
Hey @wanderingstan just FYI I helped Neil put this PR together. I'm not sure on the content of this PR. It seems right on the surface but I don't know enough about the JSON schema stuff just yet to be sure. I mostly just wanted to help Neil to go through the process of submitting a PR. Let us both know what you think when you get around to it.

tyleryasaka:
To shed some more light on why this error was happening, FWIW:

If you look at our [package-lock.json](https://github.com/OriginProtocol/demo-dapp/blob/master/package-lock.json), you'll see that we have a couple packages that have the `ajv` package as a dependency. [ethereumjs-testrpc](https://github.com/OriginProtocol/demo-dapp/blob/master/package-lock.json#L3812) has version `5.5.2` as a dependency, and [react-jsonschema-form](https://github.com/OriginProtocol/demo-dapp/blob/master/package-lock.json#L10188) and [fsevents](https://github.com/OriginProtocol/demo-dapp/blob/master/package-lock.json#L4798) have version `4.11.8` as a dependency. I think due to this ambiguity different versions were getting loaded in different environments (which is why only some people had this issue). If version 5 was loaded, this causes an error (see the [note in ajv's readme](https://github.com/epoberezkin/ajv#using-version-6)). That note has instructions to support draft-04 schemas, but unfortunately there's no way to do that with our usage through react-jsonschema-form at the moment (see [this PR](https://github.com/mozilla-services/react-jsonschema-form/pull/794)).

All that to say, this fix looks good. 😃

wanderingstan:
Thanks for this, @nsimon !

nsimon:
Hey @wanderingstan glad I could help!

PR #100, opened 2018-03-18
Issue/99 web3 account flickering

### LINKED ISSUE #99: Web3 Account Flickering
In my experience (on probably a slower machine), I see a split second of "You are not signed in to MetaMask." when I first load the DApp _while already signed in to MetaMask_.

dapp-web3-flicker

I believe this is attributable to `react-web3` relying on the synchronous [`web3.eth.accounts`](https://github.com/coopermaruyama/react-web3/blob/71b0950132d729f06461d0dba537cae49047f70a/src/Web3Provider.jsx#L196) before using the asynchronous [`web3.eth.getAccounts`](https://github.com/coopermaruyama/react-web3/blob/71b0950132d729f06461d0dba537cae49047f70a/src/Web3Provider.jsx#L104). There is apparently [a bug in MetaMask, which causes the synchronous version to be unreliable](https://github.com/MetaMask/metamask-extension/issues/2268).

I have a commit in the works that will bring the web3 provider component into our DApp code in lieu of using the React package (of which [we are already maintaining a fork](https://github.com/OriginProtocol/demo-dapp/blob/master/package.json#L66)).

micahalcorn:
This resolves #99 and simplifies the web3 integration. This removes an external package dependency, which, among other things:

**Previously**
- required the maintenance of an external fork
- included unnecessary Redux integration
- injected unused styles
- depended on Lodash
- used React [context](https://reactjs.org/docs/context.html#why-not-to-use-context)

**Considerations**

1. We were already _blocking the UI_ when (a) web3 was unavailable, (b) the user was logged out, or (c) the selected network was unsupported. However, we were only _preventing the route-specific rendering_ (listings, forms, etc) in cases (a) and (b). So if the selected network was unsupported, a user could easily remove the overlay elements and engage the web3-dependent functions. If it is preferable for all of that route-specific UI to exist and be visible behind the overlay when an unsupported network is selected, then a different rendering strategy should be used rather than the one that this PR inherits from `react-web3` (multiple conditioned return statements). Aside from this consideration, 45e30d584ae2bcb498a25c6d2a5e2b27208bc2ec at least moves that network logic out of the navbar component and into the`Web3Provider` component.
2. One consequence of this PR is that it makes the provider component code more specific to our use case (and UI), which is less modular. I think the long-term goal is to make the DApp code more modular and perhaps separate UI concerns from others. But I think the appropriate first step is to simplify the code and remove the external package dependency, as this PR does.
3. At some point in the relatively near future, [web3.js 1.0](https://github.com/ethereum/web3.js/pull/558) will be released. This may be yet another reason to not deal with an external package dependency and rather implement our own provider component.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4020,6 +4020,14 @@
             }
           }
         },
+        "string_decoder": {
+          "version": "1.0.1",
+          "bundled": true,
+          "dev": true,
+          "requires": {
+            "safe-buffer": "5.0.1"
+          }
+        },
         "string-width": {
           "version": "1.0.2",
           "bundled": true,
@@ -4028,14 +4036,6 @@
             "code-point-at": "1.1.0",
             "is-fullwidth-code-point": "1.0.0",
             "strip-ansi": "3.0.1"
-          }
-        },
-        "string_decoder": {
-          "version": "1.0.1",
-          "bundled": true,
-          "dev": true,
-          "requires": {
-            "safe-buffer": "5.0.1"
           }
         },
         "stringstream": {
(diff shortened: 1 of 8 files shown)


joshfraser:
Great PR description @micahalcorn. Nice explanation on the purpose of the pull request and the impact it will have.

PR #112, opened 2018-04-18
Fix npm test script

tyleryasaka:
### Checklist:

- [x] Code contains relevant tests for the problem you are solving
- [x] Submit to the `develop` branch instead of `master`

### Description:

Before this was failing on jest errors. This fixes the script so that it doesn't fail. (There are still no tests.)

We were getting this error: https://github.com/facebook/create-react-app/issues/1319


wanderingstan:
Hey @tyleryasaka is this one still good to go? Looks pretty minor and could be included in 0.5.9 release.

tyleryasaka:
@wanderingstan Just resolved the conflicts - this is good to go.

Kinda wondering if we should just remove the test script altogether since it's not being used, unless we plan to actually use it soon. Either way, we can go ahead and merge this and think about that later.

PR #119, opened 2018-04-20
Issue/83 change recommended wallet based on browser

ambertch:
### Checklist:

- [X] Test your work and double check you didn't break anything
- [X] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs)

### Description:

MetaMask is not supported on all platforms, and modals which refer to it should be updated.

It was decided that Cipher, Toshi, and Trust Browser should be the recommended wallet-enabled browsers for iOS.

Notes:
- "Full Instructions For Demo" link is for v0.1, remember to update before v0.2 release
- Two other modals (account unavailable, unsupported network) also updated
- I saw that MetaMask supports Android, so browsers with Android UAs are directed there as well
- Apparently matching "Android" in the UA string does not detect the platform with 100% accuracy, although a recommended library to do so(https://github.com/hgoebl/mobile-detect.js) just looks for that substring as well


micahalcorn:
**Note**: will be a fix for #83

micahalcorn:
@ambertch I've updated this to include changes from #206, so hopefully we can get this merged soon. Thanks again!

joshfraser:
@micahalcorn is this ready to be merged?

micahalcorn:
@joshfraser I think so, but this one isn’t critical for the deployment and it could use some more testing.

wanderingstan:
I'm in middle of a release now

PR #133, opened 2018-04-30
fix for missing form-data lib

inkredabull:
https://github.com/OriginProtocol/demo-dapp/issues/132


micahalcorn:
Is this just a matter of the current `develop` branch not playing well with the published npm package? I think that lib is [required by origin-js](https://github.com/OriginProtocol/origin-js/blob/develop/src/ipfs-service.js#L7) only and maybe should not be installed in this app.

inkredabull:
No longer a problem; see thread in Issue.

PR #160, opened 2018-05-10
Wire up users

micahalcorn:
This adds real identity data to ~the public profile~ most elements that represent a user. I'm pretty rusty with Redux, so feel free to offer critiques. ~I also added support for [DevTools](http://extension.remotedev.io/).~ After getting your approval or revisions on this structure, I will go back and update all of the other mock user elements in this or another branch.


src/reducers/Users.js
```diff
@@ -0,0 +1,19 @@
...
+
+    case UserConstants.FETCH_ERROR:
+      return state
+
+    case UserConstants.FETCH_SUCCESS:
+      const { user } = action
+      const users = [...state]
+      const i = users.findIndex(u => u.address === user.address)
+
+      return i === -1 ? [...users, user] : users.map(u => u.address === user.address ? user : u)
```
micahalcorn: Ternary happy 😬

src/Store.js
```diff
@@ -13,14 +14,18 @@ if (process.env.NODE_ENV !== 'production') {
   middlewares.push(logger)
 }
 
+const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
```
nick: I'm not familiar with this tool... is it worth introducing to the codebase?

src/actions/User.js
```diff
@@ -0,0 +1,26 @@
...
+)
+
+export function fetchUser(address) {
+  return async function(dispatch) {
+    try {
+      const user = await origin.users.get(address)
+
+      dispatch({
+        type: UserConstants.FETCH_SUCCESS,
+        user: Object.assign({}, user, { address }),
```
nick: A shorthand for this is:

`user: { ...user, address }`

src/reducers/Users.js
nick: I think this is OK for now but I foresee turning the state into an object at some point so we can store other state, such as loading status, error codes etc.

src/actions/User.js
micahalcorn: I forgot about that one. 👍

src/Store.js
micahalcorn: I actually didn't think twice about this since it doesn't actually add the dependency, it just exposes to the window. I'm happy to remove it and I can just add it locally if I need to debug.

micahalcorn:
@nick I'll go ahead and wire up the user in a few more places, so consider this a work in progress until later today.

micahalcorn:
@nick I think this is ready to test. The only fake user data should now be limited to notifications.

Issue #84, opened 2018-02-16
Detect wallet(provider) in use and change name/icon 


wanderingstan:
Detect which wallet user is using (Metamask or whatever) and change terminology in the dapp. 

So instead of saying "Complete transaction in Metamask" it should say, "Complete transaction in Trust Wallet". 

See  here for exact location of name in `web3` object: https://github.com/TrustWallet/trust-web3-provider#how-to-identify-trust-provider

cc: @vikmeup of Trust Browser

micahalcorn:
@jordajm I think that this might be covered by #83, which is blocked. What do you think?

jordajm:
Hmmm - I'm not sure - seems like this one is slightly different b/c it refers to changing name/icon throughout the app, but since @ambertch is working on #83, it probably makes sense for him to grab this one too..?

micahalcorn:
@jordajm ah, I see the difference. Go ahead and take this one if you want.

wanderingstan:
This SO question seems to have some good code on detecting what provider is being used: https://ethereum.stackexchange.com/questions/24266/elegant-way-to-detect-current-provider-int-web3-js

jordajm:
PR submitted here: https://github.com/OriginProtocol/demo-dapp/pull/178

The title of this issue references icons that need to be changed, but I wasn't able to locate any icons. I did a "find all" on the phrase "MetaMask" and replaced those hard-coded strings with dynamic provider names. 

Let me know if I missed anything. 

Also, this may be blocked by the pending updated way to reference `web3` in demo-dapp as noted here: https://github.com/OriginProtocol/demo-dapp/issues/126

micahalcorn:
Done via #178 ✅

PR #178, opened 2018-05-14
Issue #84 - Detect provider and change provider name in user alerts

jordajm:
### Checklist:

### Description:

This PR adds a util called `getCurrentProvider` and implements it in four files where the provider name `MetaMask` was hard-coded. Now, the provider name is determined dynamically so that users that are on providers other than MetaMask will see their provider's name in the alerts.

Also, this may be blocked by the pending updated way to reference `web3` in demo-dapp as noted here: https://github.com/OriginProtocol/demo-dapp/issues/126


src/components/web3-provider.js
```diff
@@ -147,7 +150,7 @@ class Web3Provider extends Component {
     curr = curr && curr.toLowerCase()
 
     if (curr !== next) {
-      curr && Store.dispatch(showAlert('MetaMask account has changed.'))
+      curr && Store.dispatch(showAlert('{this.state.currentProvider} account has changed.'))
```
micahalcorn: Looks like this is supposed to be a template literal.

src/utils/getCurrentProvider.js
```diff
@@ -0,0 +1,24 @@
+export default function getCurrentProvider(web3) {
+    if (!web3 || !web3.version) {
```
micahalcorn: Please also convert this to two-space tabs instead of four-space. ✌️

jordajm:
@micahalcorn good catches - thanks!

jordajm:
Fixed and ready for another look @micahalcorn

micahalcorn:
Thanks, @jordajm. Out of curiosity were you able to test this with any providers other than MetaMask? We could do it on mobile by deploying this branch to Heroku, but that seems like overkill.

micahalcorn:
@jordajm I haven't forgotten about this. I was just trying to test on mobile and found other issues.

jordajm:
@micahalcorn no worries - how are you doing mobile testing? I usually just use BrowserStack but I don't think I'd be able to test wallet scenarios there... do you have an easy way to deploy this branch to Heroku or are you running localhost on your phone?

micahalcorn:
@jordajm currently Heroku, which seems to be the easiest way to test across multiple browsers. But I've yet to find a mobile wallet/browser that will result in the appropriate scenario to test one of the messages. I'm distracted with other (unrelated) issues for now. 😫

jordajm:
@micahalcorn do you have this branch deployed to Heroku?

micahalcorn:
@jordajm I honestly don't remember what is deployed. 😭 I'll let you know when I do.

jordajm:
Ok let me know if I can help with testing - but without being able to deploy the branch to Heroku I don't think I'll be much help.

micahalcorn:
This is great @jordajm 👍 I have tested it on all of the known providers except for Parity, which I can't make sense of. I suggest that we merge this and then worry about what recommendations to make based on which browser a person is using when there is no given provider (i.e., don't tell them to install MetaMask when on Safari).

  | Not Web3 Enabled | Not Signed In To Wallet | Unsupported Network
-- | -- | -- | --
MetaMask | N/A | ✅ | ✅
Trust | N/A | N/A | ✅
Toshi | N/A | N/A | ✅
Cipher | N/A | N/A | ✅
Mist | N/A | ✅ | ✅
Parity | ❔ | ❔ | ❔
Infura | N/A | N/A | ✅
Localhost | N/A | N/A | ✅

Issue #62, opened 2018-01-09
Handle case where use rejects/fails to complete metamask transaction


wanderingstan:
For example:
![image](https://user-images.githubusercontent.com/673455/34702340-14348566-f4a2-11e7-9a9c-aaf0c5b9173d.png)

At the moment, all we do is `alert()` the error, here:
https://github.com/OriginProtocol/demo-dapp/blob/master/src/components/listing-detail.js#L76

We should give message and reset page back to normal.

wanderingstan:
Note victor from Trust Browser notified that this is big problem in his browser.

ambertch:
It doesn't look like Trust Browser can connect to a local instance, although it looks like this was added at the beginning of March (this issue is from January):

<img width="389" alt="screenshot 2018-04-16 18 35 55" src="https://user-images.githubusercontent.com/156263/38843916-0f0d9004-41a6-11e8-946f-753d0c6e684e.png">

<img width="825" alt="screenshot 2018-04-16 18 42 03" src="https://user-images.githubusercontent.com/156263/38843918-125e1972-41a6-11e8-8e20-c2117b78efd0.png">

ambertch:
Sorry, by local instance I mean that there doesn't seem to be a way to point Trust Browser's wallet to the truffle development instance, although I hadn't realized at the time that I could access my local dev instance of the demo app using Trust Browser connected to Rinkby.

![img_1083](https://user-images.githubusercontent.com/156263/38900154-48d2b0e0-424f-11e8-97f5-b846427422e1.PNG)

For clarification, does this issue pertain to desktop + Trust Browser, or just the latter? 

On desktop (Metamask), 

- if I press the "reject" button, the component state resets
- if I cancel out of the transaction by closing the popup or window, the component state gets stuck, but this is because doing so does not change Metamask's state (the transaction remains in a pending state) 

<img width="816" alt="screenshot 2018-04-17 14 21 09" src="https://user-images.githubusercontent.com/156263/38900493-71ffa4f4-4250-11e8-8296-5aa36fd25dcb.png">

jordajm:
@wanderingstan @micahalcorn @ambertch  I'm thinking of grabbing this one next - any objections?

wanderingstan:
No objections at all. This one has been lingering around a while.

nick:
Just a heads up that the user profile does handle errors, in case you wanted to use the same technique elsewhere in the code:

[...]

PR #181, opened 2018-05-15
Issue #62 - Handle MetaMask errors and allow user to reset state

jordajm:
### Checklist:

### Description:

When a user rejects a transaction via the provider's popup window, demo-dapp needs to handle that gracefully and allow the user to reset the state of the app to remove the "waiting" modal. 

This PR uses the error modal styling from the Profile page and implements it on the "Create Listing" and "Listing Detail" pages.

### DIFF
--- a/src/components/listing-create.js
+++ b/src/components/listing-create.js
@@ -25,7 +25,8 @@ class ListingCreate extends Component {
       PREVIEW: 3,
       METAMASK: 4,
       PROCESSING: 5,
-      SUCCESS: 6
+      SUCCESS: 6,
+      ERROR: 7
     }
 
     this.schemaList = [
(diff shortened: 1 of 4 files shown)


src/components/listing-detail.js
```diff
@@ -111,18 +112,21 @@ class ListingsDetail extends Component {
...
+  }
+
 
   render() {
     const buyersReviews = this.state.reviews.filter(r => r.revieweeRole === 'SELLER')
 
     return (
       <div className="listing-detail">
-        {this.state.step===this.STEP.METAMASK &&
+        {this.state.step === this.STEP.METAMASK &&
```
micahalcorn: 😊

src/components/listing-create.js
```diff
@@ -218,6 +222,23 @@ class ListingCreate extends Component {
...
+            {this.state.step === this.STEP.ERROR && (
+              <Modal backdrop="static" isOpen={true}>
+                <div className="image-container">
+                  <img src="images/flat_cross_icon.svg" role="presentation" />
+                </div>
+                There was a problem creating this listing.<br />See the console for more details.<br />
+                <a
+                  href="#"
+                  onClick={e => {
+                    e.preventDefault()
```
micahalcorn: Style/preference question: Do you prefer for this to be handled down here in the render function rather than inside of `resetToPreview`? I would have done as follows but am wondering if there is a reason not to...

```
onClick={this.resetToPreview}
```
```
resetToPreview(e) {
  e.preventDefault()

  this.setState({ step: this.STEP.PREVIEW })
}
```

micahalcorn:
Sneaking a couple of unrelated fixes into this PR.

src/components/listing-create.js
micahalcorn: I see now that you followed the convention that was already there. 👍

PR #188, opened 2018-05-17
Issue #36 : add approx price

inkredabull:
### Checklist:

- [x] No Tests
- [x] No README change

### Description:

 - opted for least-effort inclusion of approximate price with Cryptonator API
 - no tests; [maybe someday?](https://github.com/OriginProtocol/demo-dapp/issues/149)

screen shot 2018-05-16 at 5 50 29 pm

Ref: https://github.com/OriginProtocol/demo-dapp/pull/130

### DIFF
--- a/src/components/listing-card-prices.js
+++ b/src/components/listing-card-prices.js
@@ -1,0 +1,86 @@
+import React, { Component } from 'react'
+
+const baseCurrencyCode = 'ETH';
+
+class ListingCardPrices extends Component {
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      price: "Loading...",
+      exchangeRate: null,
+      approxPrice: "Loading...",
+      currencyCode: "USD",
+      defaultDecimalPlaces: 5,
+      exchangeBaseURL: 'https://api.cryptonator.com/api/ticker/'
+    }
+  }
+
+  componentWillReceiveProps(nextProps){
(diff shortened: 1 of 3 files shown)


DanielVF:
Micah, maybe Aure could take a pass at how we want to display alternate prices, both on the search page, and the listing detail pages.

I changed this to use two decimal places.

screen shot 2018-05-17 at 3 38 30 pm

wanderingstan:
For reference, this is how MetaMask shows ETH/USD combinations: 
image 
image



Not that MetaMask is a great design inspiration, but it is what a lot of our users will see as the final price that they have to sign off on.

DanielVF:
Here's another design:

screen shot 2018-05-18 at 12 47 39 pm

auregimon:
I like the first option @DanielVF 

ETH and USD to the left of it, same font size and USD in lighter gray. Try putting a pipe | in between the two if it looks weird.

DanielVF:
And here's the final design:

image

wanderingstan:
@matthewliu , am I correct in understanding that you would prefer (A) to have the eth/token price first, with approx dollar afterward (with dollar more clearly marked as appprox), or (B) take out dollars entirely, and maybe we'll rethink the design?

matthewliu:
A. I like the dollars but having it as primary is misleading because it would be changing minute-by-minute.

PR #206, opened 2018-05-22
Pass web3 to Origin.js

micahalcorn:
**_This depends on a merge of https://github.com/OriginProtocol/origin-js/pull/202._**

If MetaMask is not installed and a native `web3` object is not provided by the browser, then the DApp will fail to render due to the exception thrown [here](https://github.com/OriginProtocol/origin-js/blob/develop/src/services/contract-service.js#L17). This is the critical step to satisfy https://github.com/OriginProtocol/demo-dapp/issues/128, which should probably be broken up into several issues.


tyleryasaka:
I'll approve after https://github.com/OriginProtocol/origin-js/pull/202 is merged

PR #212, opened 2018-05-24
hotfix/Pagination

micahalcorn:
☑️ _This PR is intentionally pointed to `master` since it is a hotfix and we don't want to confuse this branch's package.json with `develop`'s (among other things)._

Pagination was not working because the [`wrapRouter` and `connect` wrappers were in the wrong order](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/redux.md#blocked-updates).


micahalcorn:
Double-approved ✅

PR #216, opened 2018-05-25
Hotfix/zeros

micahalcorn:
☑️ _This PR is intentionally pointed to master since it is a hotfix and we don't want to confuse this branch's package.json with develop's (among other things)._

As a result of https://github.com/OriginProtocol/demo-dapp/pull/199/commits/3bd9cd7523e800df7215b06dcc45c237067847ca#diff-a9103ca3b51d952e7ee2970ed96fbec9, listings without an image resulted in being prepended by 0️⃣s in the listing card. Zeros no more.

screen shot 2018-05-24 at 7 27 45 pm 2

### DIFF
--- a/src/components/listing-card.js
+++ b/src/components/listing-card.js
@@ -34,7 +34,7 @@ class ListingCard extends Component {
     return (
       <div className={`col-12 col-md-6 col-lg-4 listing-card${loading ? ' loading' : ''}`}>
         <Link to={`/listing/${address}`}>
-          {photo &&
+          {!!photo &&
             <div className="photo" style={{ backgroundImage: `url("${photo}")` }}></div>
           }
           {!photo &&


tyleryasaka:
@micahalcorn I'm not quite familiar with our release process, but I feel like in this case since we're planning to do a new release soon anyway, it would keep things simpler just to merge this into `develop`. Whereas a hotfix is something where we have to get something fixed *today* or all hell will break loose. Just my 2 cents.

micahalcorn:
@tyleryasaka agreed now that we know that the plan has changed a bit. I went into hotfix mode last night with #212 because we had already deployed everything and stumbled across DApp issues that needed immediate attention. I would vote for merging this one to `master` (particularly since that is what I branched from), and I'll use develop in the future if it is a last-minute part of a release.

Issue #209, opened 2018-05-23
Cannot create a pet announcement listing on demo.originprotocol.com 


gaboesquivel:
It seems that for some reason it is expecting a price 
Metamask window is not being openened.

<img width="1385" alt="submit-announcement" src="https://user-images.githubusercontent.com/391270/40438193-a9dd2924-5e74-11e8-9b6f-aab83b3a206f.png">

micahalcorn:
@wanderingstan @tyleryasaka @DanielVF it looks like at one time we anticipated some listings that would not include a price? Should this be dropped since [`origin.js` seems to require a price](https://github.com/OriginProtocol/origin-js/blob/master/src/resources/listings.js#L71)? Or should we support a price of `0` and include that behind the scenes in the DApp?

joshfraser:
Free listings are pretty common. Think about the  free section on Craigslist or announcements on Nextdoor. Probably something we would support.

micahalcorn:
Done via #225 ✅

PR #225, opened 2018-05-27
📣Repair announcement listings

micahalcorn:
This addresses #209 by modifying the announcement listing schema to only allow a single enumerated value (`0`), which is also set as the `default`. The form component then detects this "auto value" and hides the price field.


micahalcorn:
@joshfraser this should be included in today's deployment, if possible.

PR #227, opened 2018-05-28
🎉 Walletless browsing

micahalcorn:
Another commit or two might trickle in, but I'm not planning any more significant changes here.

**Summary**
- Allows an override of the production URL with `PRODUCTION_DOMAIN` env variable
- Requires a `PROVIDER_URL` env variable for a web3 network node
- Creates a web3 provider if none is available in the browser
- Creates a reactive single source of truth for current ETH account in the Redux store
- Moves browser/device detection to the top-level component and stores the result in Redux
- Allows ui to render behind blocking web3 alert modals
- Simulates delayed connections (per @joshfraser)
- Auto-shows and auto-hides the connectivity dropdown when loaded without web3
- Improves modal styling
- Renders modal alerts based on web3 intent

... and other things 🤷‍♂️ 

Live: http://demo-micah.herokuapp.com/


micahalcorn:
@DanielVF this branch is currently deployed to http://demo-micah.herokuapp.com/ for testing 😬

DanielVF:
Great! This just works.

PR #233, opened 2018-05-28
Add component for displaying Etherscan links

joshfraser:
Adds a universal component for displaying Etherscan links. You simply pass in a hash for an address or transaction and it will create an Etherscan link that points to the right network (rinkeby.etherscan.io, etc) and the right URL path (/address or /tx).

### DIFF
--- a/src/css/app.css
+++ b/src/css/app.css
@@ -208,7 +208,7 @@ a:hover {
 }
 
 .identification a {
-  color: white;
+  color: var(--dusk);
 }
 
 .user-card .identification > div {
--- a/src/css/app.css
+++ b/src/css/app.css
@@ -204,6 +204,7 @@ a:hover {
 }
 
 .identification .name {
+  color: var(--dark);
   font-weight: bold;
 }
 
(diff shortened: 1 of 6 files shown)


src/css/app.css
```diff
@@ -208,7 +208,7 @@ a:hover {
 }
 
 .identification a {
-  color: white;
+  color: var(--dusk);
```
micahalcorn: @joshfraser was this intended to dark the user name here or something somewhere else?

usercard

joshfraser: Oops my bad. I changed that class because the buyer & seller names on the purchase view were invisible because they got turned white when linked. Didn't realize that class was used in multiple places.

tyleryasaka: @micahalcorn No, that's here:

screen shot 2018-05-28 at 2 40 02 pm

micahalcorn: @tyleryasaka poor naming practice on my part, maybe. I've moved your target link up a couple of generations in the tree. I'll change this back and they will play nice.

tyleryasaka:
Please don't merge this yet - I'm getting DOM validation errors. Looking into those now.

Issue #63, opened 2018-01-10
Localization


wanderingstan:
We can't be English/America-only. Let's get this DApp global! 

Overall plan is: 

Step 1: Choose localization library. Default choice for React would seem to be: [react-localization](https://github.com/stefalda/react-localization)

Step 2: Break apart current english strings and replace with appropriate l10n calls.

Step 3: Get translations for different languages. 

If you have an opinion on 1, or want to do some research, that'd be very useful. Step 2 will be the bulk of this issue. 

Sub-tasks created:

- [x] https://github.com/OriginProtocol/demo-dapp/issues/245
- [x] https://github.com/OriginProtocol/demo-dapp/issues/246
- [x] https://github.com/OriginProtocol/demo-dapp/issues/247
- ~https://github.com/OriginProtocol/demo-dapp/issues/248~
- [x] https://github.com/OriginProtocol/demo-dapp/issues/249
- [x] https://github.com/OriginProtocol/demo-dapp/issues/255
- [ ] https://github.com/OriginProtocol/demo-dapp/issues/263

micahalcorn:
@wanderingstan do you want to link us here to the implementation(s) that you saw for us to reference?

wanderingstan:
John J and I had discord convo on this, with me forgetting g we already had an open issue. 

His comment there: 
> cool - I'll look at swarm's stuff but I think this is the winner in terms of react i18n stuff:
> https://www.npmjs.com/package/react-intl

jordajm:
`react-i18next` looks cool too though - worth checkout out - and @wanderingstan found this for translating between the json files it uses and `.po/.pot` files of `gettext`
https://www.npmjs.com/package/react-intl-po

wanderingstan:
For reference: Here is our translation files & workflow or the website: 
https://github.com/OriginProtocol/company-website/tree/master/translations

jordajm:
WIP PR:
https://github.com/OriginProtocol/demo-dapp/pull/238

micahalcorn:
Considered done with https://github.com/OriginProtocol/origin-dapp/issues/263 split into sub-tasks ✅

PR #238, opened 2018-05-29
Issue #63 - Localization of the demo-dapp - PART 1

jordajm:
WIP - Not ready for thorough review, but feedback is welcome. I'll continue working on getting the rest of the strings/dates/numbers/etc wrapped and ready to be translated.

### Checklist:

- [x] Test your work and double check you didn't break anything *(as far as I can tell, but this will need a lot of testing...)*

### Description:
- This is a ton of stuff to review... but the meat of it happens in [index.js](https://github.com/OriginProtocol/demo-dapp/compare/issue_63?expand=1#diff-1fdf421c05c1140f6d71444ea2b27638) and [package.json](https://github.com/OriginProtocol/demo-dapp/compare/issue_63?expand=1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) and is explained in a new [README](https://github.com/OriginProtocol/demo-dapp/blob/6a702d50b5a44aa2afd4bb7ed6958a033936b220/translations/README.md) in the `/translations` directory
- I'll fix conflicts with the `develop` branch all at once in the next day or two 😅 and I won't check in the change to `origin-js` version currently showing in `package.json`

### Open questions / TODO's:
- Can we localize `react-jsonschema-form` used on `listing-creste.js`?
- Not sure if sentence construction logic in `notification.js` is going to work in other languages
- Look at using `moment-locale` or other localization solution for `<Timelapse>` components
- move language/messages variables in `index.js` to redux so we can allow user to change it via dropdown menu
- What CSS might we need for right-to-left languages?


wanderingstan:
Can't wait to check this out. 

Quick skim of code and I agree that our sentence construction logic is going to have to change to be localizable. Saw it in my-listing-card too.

micahalcorn:
Thanks @jordajm 🌐 Yes, the human-readable notifications mess was created before I realized that we were planning to translate the DApp. In the company website, we always strived to have people translate entire sentences because we can't rely on a single sentence structure pattern. But in this case, we're surely not going to have people translating every user input. Perhaps the notification messages should just be generic (i.e., "someone has purchased your listing") and then point to the listing in question. I'll put this on the list of product questions to tackle. cc @wanderingstan

wanderingstan:
Ouch! Yeah, I just saw the human-readable code. For others' reference it's here:
https://github.com/OriginProtocol/demo-dapp/blob/96911aec45d5646a1ee04ee5da95171458196250/src/components/notification.js#L8-L44

That's a textbook example of hard-coding English into an app. I think we can actually pull off a localized version of what that code is trying to, but it'll take some work. It will mean breaking it out into all different possible sentences with variables substituted it, rather than assembling parts and putting them together with `{subject} {presPerf} {verb}`.

jordajm:
@joshfraser - I fixed some bugs in the translation scripts and resolved the merge conflicts... can we merge this quick before there are more commits that cause conflicts? 🙏 😁 

@wanderingstan @micahalcorn I'll create additional issues for the rest of the localization work and break future localization PRs smaller and quicker to prevent conflicts since they touch so many files...

README.md
```diff
@@ -108,6 +111,7 @@ Heroku will prompt you to set some config variables. You will likely want to sti
 Since this app is just a bunch of HTML and JavaScript, you can also deploy and use it directly from IPFS. Just run:
 
 `./scripts/deploy.sh`
+>>>>>>> develop
```
joshfraser: Looks like a mistake

joshfraser:
This is a big PR (not your fault, goes with the territory).

@wanderingstan @micahalcorn or @tyleryasaka would appreciate your eyes & sign off on this.

micahalcorn:
@wanderingstan @tyleryasaka @jordajm I've committed a small fix here and may have more. Looking great so far! 👍

jordajm:
Thanks @tyleryasaka - nice work 👍
I've done a lot of experimentation with different options for the `react-intl-po` commands, and I ended up using the `message-context` option because it adds the message ID to the `.po` file, so that when it gets re-converted to a JSON file, it looks at IDs instead of default strings to do the matching, which sounds like a good idea for long-term IMO. If we don't use the IDs, I think it's confusing b/c we're assigning unique IDs but actually matching on default text, not ID. 
Not sure why the command was failing for you but I'll investigate and report back.

tyleryasaka:
@jordajm Gotcha. Yeah if you can get the command working with the option I'm all for it.

jordajm:
I'll create issues to track work on these localization-related tasks that are still unfinished:
- Whitepaper and product brief links in footer should be language-aware
- Add a language select component
- Move language config to Redux
- Localize web3 modals
- Replace sentence concatenation with full sentences that can be translated - `notification.js`and maybe other components
- Timelapse component needs a localization strategy - maybe moment-locale

Issue #249, opened 2018-06-02
Complete wrapping of strings with <FormattedMessage> component


jordajm:
Subtask of: https://github.com/OriginProtocol/demo-dapp/pull/238

A number of JSX files still need to be refactored so that all strings are wrapped in a `<FormattedMessage>` component and given a unique `id` and `defaultMessage`.

micahalcorn:
I have a question about the unique id. It appears that your convention is to use the naming convention `{component-file-name}.{whatever}`. As file names change, that won't cause any problems, will it?

If in one file (say purchase-detail.js) we have a message id `purchase-detail.seller` with a default message of "seller" and then in another file/component we want to translate the string "seller", then would we re-use the message id `purchase-detail.seller` so that we don't have to translate the same string twice with two different message ids?

jordajm:
@micahalcorn - correct, if we change the file name, the ID could either stay the same or change depending on how much we care about naming convention.

Regarding duplicate IDs, at first I tried having a "shared" namespace but `react-intl` threw errors b/c evidently it's not a "best practice" to use the same ID across multiple files. Plus, once there were over a hundred IDs I couldn't keep track of all of what was shared vs unique, so it makes sense to me to keep everything namespaced and unique. 

The exception to that rule is if you have the same string multiple times in the same file, you can re-use the same ID as long as the `defaultMessage` is the same in each case.

jordajm:
Completed by https://github.com/OriginProtocol/demo-dapp/pull/252

PR #252, opened 2018-06-02
Issue #249 - Wrapping the remaining strings in <FormattedMessage> component

jordajm:
### Checklist:

### Description:
- Hopefully this is the last of the giant internationalization PRs... they should be smaller after this.
- Issue: https://github.com/OriginProtocol/demo-dapp/issues/249
- Unless I missed some strings, this should complete the wrapping of strings throughout the app
- Fixed a minor bug in web3-provider.js related to undefined `props`
- Added `rel="noopener noreferrer"` to some links with `target="_blank"`
- Also added updated messages JSON and `.po`

### DIFF
--- a/src/components/web3-provider.js
+++ b/src/components/web3-provider.js
@@ -148,7 +148,7 @@ const NoWeb3Account = ({ currentProvider, storeWeb3Intent, web3Intent }) => (
     <a
       className="close"
       aria-label="Close"
-      onClick={() => props.storeWeb3Intent(null)}
+      onClick={() => storeWeb3Intent(null)}
     >
       <span aria-hidden="true">&times;</span>
     </a>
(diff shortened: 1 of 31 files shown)


micahalcorn:
Thanks @jordajm 🇨🇳 Will you add the additional machine translations for `zh`? I tried doing it but I keep getting empty strings when I run `npm run translate:build-translated`. I'm not sure if I'm missing a step in the readme or doing something wrong on the Google Translate side.

jordajm:
Good call @micahalcorn - I updated the machine translations of Chinese so those console error messages would go away. I was holding off b/c those translations were really just for the demo last Wednesday and we don't have a process for generating real translations for the DApp yet but I'm glad you suggested that b/c it made me figure out the fix for the translation build scripts (and get rid of the errors) :-)

jordajm:
@micahalcorn I think I have wrapped all the strings aside from the ones that need refactoring or the ones in schemas, which are covered in other issues. 

Would you mind giving this PR a last look?

jordajm:
@micahalcorn Did transactions dropdown, unnamed user strings, country options, as well as some stuff I missed in the user, notification, and connectivity files. 
Also updated Chinese translations for testing.

PR #257, opened 2018-06-05
Transactions dropdown

micahalcorn:
This adds a dropdown to the navbar containing the [simulated] progress for pending blockchain confirmations. This will need to be wired up before merging to master and deploying, but I would like to get it merged into develop for two reasons:

1. There are a ton of CSS changes resulting from previously using the word "transaction" as a class name when it should have been "purchase".
2. The notifications dropdown is reintroduced to allow for the appropriate positioning of it and several other dropdowns.

Doing this now will help avoid merge conflicts.


micahalcorn:
@tyleryasaka I don't think that this one is using `line-height`. I tried changing my font size, zoom, etc, and couldn't reproduce it. Any chance that a browser extension is causing it?

Issue #246, opened 2018-06-02
Add language selection dropdown menu


jordajm:
Sub-task of: https://github.com/OriginProtocol/demo-dapp/pull/238

As a user, I want to change the language of the DApp using a dropdown menu that shows all available languages. 

According to Aure, there is no design for this menu 

Let's put it in the footer and make it match other footer link styling as much as possible. 

This task will require moving language config currently found in `index.js` to Redux

micahalcorn:
Let's actually put this in the footer, especially if we are auto-detecting the preferred language in most cases. cc @wanderingstan

jordajm:
Good call @micahalcorn - description udpated.

micahalcorn:
Done via #260 ✅

PR #260, opened 2018-06-08
Issue #246 - Add language selection dropdown menu

jordajm:
### Checklist:

### Description:
- Move the `react-intl` logic and settings into redux
- Add dropdown to footer
- Make a way to mount and unmount the whole app so it fully re-renders after user selects a new langauge
- Consolidate all localization util functions into a single util file


jordajm:
@micahalcorn - per our discord chat, this PR now stores user's selected lang in local storage, and if user changes it via the footer dropdown menu, the page refreshes.

Issue #163, opened 2018-05-10
Wire up notifications


micahalcorn:
Once https://github.com/OriginProtocol/bridge-server/issues/6 is completed, we will need to remove the existing mock data and wire up the real data.

- [ ] navbar dropdown
- [ ] notifications page

PR #261, opened 2018-06-09
Wire up notifications

micahalcorn:
This addresses #163 and simplifies the logic for rendering the notification content. Several changes are made to the localization ids and default strings. There are also a few other UX improvements.


src/components/dropdowns/notifications.js
```diff
@@ -1,25 +1,47 @@
...
+      const { sellerAddress } = n.resources.listing
+
+      return { ...n, perspective: web3Account === sellerAddress ? 'seller' : 'buyer' }
+    })
+    const filteredNotifications = notificationsWithPerspective.filter(n => n.status === 'unread')
     // avoid integers greater than two digits
-    const notificationCount = exampleCounts < 100 ? Number(exampleCounts).toLocaleString() : `${Number(99).toLocaleString()}+`
+    const notificationCount = filteredNotifications.length < 100 ?
+                              Number(filteredNotifications.length).toLocaleString() :
+                              `${Number(99).toLocaleString()}+`
```
micahalcorn: @jordajm I think I saw that react-intl has some number localization utilities. Is it appropriate to replace this native localization method?

src/components/notification.js
```diff
@@ -1,132 +1,174 @@
...
 
-/*
-  This subcomponent turns structured notification objects into dom elements,
-  which wrap an anchor if there is a relevant listing.
-*/
-class HumanReadableNotification extends Component {
+// support other derived notifications in the future
+const NON_PURCHASE_RELATED_MESSAGE = 'A message from Origin that does not involve a listing'
+
+class NotificationMessage extends Component {
```
jordajm: This file seems like it would be cleaner if it just had one class - what would you think about moving this `NotificationMessage` class to its own file?

src/components/purchase-detail.js
```diff
@@ -175,6 +177,13 @@ class PurchaseDetail extends Component {
     if (prevState.listing.sellerAddress !== sellerAddress) {
       this.loadSeller(sellerAddress)
     }
+
+    // detect route param change and reload data
+    if (address && address !== this.props.purchaseAddress) {
+      this.setState(defaultState)
+
+      this.loadPurchase()
+    }
```
jordajm: IMO this would be cleaner as:
```
componentWillUnmount() {
    this.setState(defaultState)
}
```
This will prevent the previously viewed purchase data from showing for a second before the new data loads, and the purchase data will be loaded on `componentWillMount()`.
I don't think there's a way this page would need to update without the component being unmounted/remounted, correct?

src/components/dropdowns/notifications.js
micahalcorn: Cool, I'm not aware of a compelling reason to use [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat).

src/components/notification.js
micahalcorn: @jordajm done ✅

Issue #248, opened 2018-06-02
Localize <Timelapse> component


jordajm:
Sub-task of: https://github.com/OriginProtocol/demo-dapp/pull/238

The `<Timelapse>` component is used to say how long ago an event occurred. 

It is currently english-only and needs to be internationalized. 

One possible strategy may be to use `moment-locale` in addition to or instead of the standard `moment` library, but more research is required.

micahalcorn:
No longer necessary thanks to https://github.com/OriginProtocol/origin-dapp/pull/271 🎉

PR #271, opened 2018-06-16
Issue #248 - replace <timelapse> component with locale-aware moment.js

jordajm:
### Checklist:

### Description:
- Instead of a roll-your-own `<Timelapse>` component, we can simply set the locale in `moment.js` and rely on their handy [`timeFromNow`](https://momentjs.com/docs/#/displaying/fromnow/) method to display how long ago an event occurred. 

(For screen shots, ignore the hard-coded notification messages and just look at the beautiful "how long ago" text)

#### English:
screenshot 2018-06-16 00 50 36

#### Chinese:
screenshot 2018-06-16 00 51 30


micahalcorn:
How did I not know about this method? 🙉 Can we still abstract away the reactivity (updating the count) with appropriate intervals instead of arbitrarily choosing 30 seconds? I think that these are still subject to possible memory leaks if the components get unmounted.

jordajm:
Glad you mentioned the memory leaks - I thought about that as I was falling asleep last night but had forgotten again by this morning 😬 I'll fix those later today.

Since the interval can be set in each component, is that configurable enough or do you have an idea for another way to do it?

micahalcorn:
In the prior, homegrown, experiment, the component accepted a prop for reactivity (`true`|`false`). If `true`, it would adjust the interval over time based on the amount of time that had lapsed, i.e., no need to update every 30 seconds if it's been a year since the review was left.

This isn't really critical, I'm just a fan of reactivity, and not relying future-me to set the right interval when calling `timeFromNow`.

jordajm:
@micahalcorn per our discussion today, this now only has a `setInterval` on the profile page b/c the other places didn't seem like they'd need them. 
The transactions dropdown menu will probably need them but I think it'll be a lot easier once they're wired up to real data b/c the loop makes it tricky the way it is now.

PR #272, opened 2018-06-18
Crowdin integration - fix error in Crowdin console 

jordajm:
### Checklist:

### Description:
Fixing the below error that appeared in the Crowdin console. This change will just affect the directory name where Crowdin will place the translated files when it creates PRs. The script that relies on this hasn't been written yet, so there is no change needed outside of this one file. 

screenshot 2018-06-17 23 36 14


jordajm:
@micahalcorn one small fix here, then we should be able to get the translators going on the DApp (for real this time ;-)

PR #278, opened 2018-06-22
Airbnb attestation

sparrowDom:
### Checklist:

- [X] Test your work and double check you didn't break anything

### Description:
[Issue: Verify Airbnb identity for ERC725](https://github.com/OriginProtocol/origin-js/issues/226)

- adding Airbnb attestation to profile screen. 
- see user flow in below screenshots:
  - user is prompted to input the link to their Airbnb profile
  - dapp via origin-js calls bridge server to generate verification code
  - user edits verification code in their Airbnb's profile description
  - bridge server fetches Airbnb profile view to confirm change
- if user inputs non existing Airbnb profile or verification code is not found an error dialog is returned

### Dependant Pull Requests: 
- [Bridge-server Airbnb attestation](https://github.com/OriginProtocol/origin-bridge/pull/93)
- [Origin-js Airbnb attestation](https://github.com/OriginProtocol/origin-js/pull/252)
- [Origin-doc Airbnb attestation](https://github.com/OriginProtocol/origin-docs/pull/37)
- [Origin-bridge Airbnb attestation fix](https://github.com/OriginProtocol/origin-bridge/pull/97)

screenshot 2018-06-22 07 44 44screenshot 2018-06-22 07 46 10

screenshot-2018-06-26-15 39 31screenshot 2018-06-22 07 46 30

screenshot 2018-06-22 07 46 57


tyleryasaka:
When I reload the page the attestation is gone.

screen shot 2018-06-25 at 1 51 05 pm

micahalcorn:
@tyleryasaka I actually just had a similar experience with phone this morning. I remember that happening a couple of months ago and I can't remember how we resolved it.

tyleryasaka:
@micahalcorn Yeah, there can be a couple reasons for it.

(1) The attestation may not be getting saved onto the identity
(2) The attestation is getting saved, but is not passing our client-side validation for some reason.

tyleryasaka:
Can you troubleshoot this @sparrowDom ?

micahalcorn:
@tyleryasaka my issue was probably a missing environment variable since I was doing it on my new machine. Last time that we both experienced this I think it had something to do with a private key missing on the bridge server.

tyleryasaka:
Oh! I think you're right

tyleryasaka:
Still not having any luck

sparrowDom:
sorry for missing that 🙈 and  @tyleryasaka tnx for finding the bug 

Funny thing... turns out that if claim data contains only numbers (even though they are of type string) the function [isValidAttestation origin-js/user.js](https://github.com/OriginProtocol/origin-js/blob/master/src/resources/users.js#L192-L203) fails to validate claim. There is a separate [Pull Request](https://github.com/OriginProtocol/origin-bridge/pull/97) on bridge-server repository that fixes the problem.

I didn't yet get to the bottom of it, so I can't say what the exact problem is. Will investigate tomorrow. 

Oh and one more thing, since `origin-js` hashes attestation data before handing it forward we can't access the users airbnb userId property from the blockchain. Which was the goal of the [ticket](https://github.com/OriginProtocol/origin-js/issues/226). We should probably alter that?

tyleryasaka:
@sparrowDom Good point about the AirBnB username being private. This has been our design for the other fields like facebook/twitter/phone/email (for now) because we wanted to err on the side of caution as far as privacy for the first iteration.

Before doing anything, I would verify with @micahalcorn that we want to make the AirBnB username public. This would be a deviation from how we are handling the other attestations so I would just double check.

That said, the way I would implement this would be to store the AirBnB username on IPFS. Then the `data` field on ERC735 would be the IPFS hash of the AirBnB username. This would let us store an arbitrary length username in a way that could be looked up on IPFS from the hash that we store on the blockchain. This will require some refactoring most likely and careful thought around how to engineer this.

For example, you might want to store the username in IPFS *from origin-bridge* and simply send the hash back to origin-js. However, origin-js would need to treat this hash differently than the normal `data` response, because it would *already* be hashed (and should not be hashed again). On the other hand, the username could be sent to IPFS from origin-js itself. In that case, origin-js would just need to know that for AirBnB data fields, we should send the data to IPFS and store the IPFS hash, rather than simply doing a sha256 hash on the data (which is what we do for facebook/twitter/email/phone).

All that to say it could be quite complex - probably better to deal with that as a separate task after this is complete.

micahalcorn:
Thanks @tyleryasaka. Perhaps the question is what we're trying to accomplish on the product side by enabling Airbnb attestation. If users are expecting to "bring their reputation" from Airbnb to Origin, then they would expect people to be able to see who they are, their reviews, etc on that platform. In that case, exposing their username would be necessary. But of course, we're not going to account for their reputation since it's not on the blockchain and conforming to our standards.

For that matter, people might be disappointed that their Facebook and Twitter accounts are NOT linked to their social profiles. @wanderingstan ?

wanderingstan:
Well shoot. Yes, this is of course very similar the questions we faced around Facebook and email.

Somehow it seems less dangerous/onerous to tie one's AirBnB username to a blockchain identity, especially for sellers/hosts: Any homesharing host is effectively a mini-business who wants customers, so why would you not want people to find what you have to sell?

For guests/buyers the situation is a little trickier, but still an AirBnB profile is a lot less revealing that a Facebook id or email address. 

I think I'll have to call in @joshfraser and @matthewliu to weigh in on this one. :)

tyleryasaka:
Let's continue the discussion on a separate issue: https://github.com/OriginProtocol/origin-js/issues/257

In the meantime, how do we feel about merging the current state of this PR? (The username is currently private; it works the same as the other attestations.)

micahalcorn:
@tyleryasaka I'll take a look today. 👍 I suspect that tomorrow's engineering call might be a decent opportunity to discuss.

micahalcorn:
I've removed the placeholder for the Address service and moved the Airbnb service button into its spot as approved by Matt during yesterday's engineering call. Thanks again @sparrowDom for all your efforts!

PR #339, opened 2018-08-03
Deal with block time in the future

micahalcorn:
I don't know if this is only a local blockchain quirk or not, but I know that we can't entirely rely on block times since they can be manipulated by miners. This adds a safeguard so that [`fromNow`](https://momentjs.com/docs/#/displaying/fromnow/) does not produce "in a few seconds" or whatever.


micahalcorn:
I wonder if this somehow has something to do with my local machine's clock. 🤔 Either way, we don't want the UI to say that something happened in the future.

PR #352, opened 2018-08-06
Localization improvements - Issue 340

jordajm:

### Checklist:

### Description:

- Use locale codes instead of language names identify languages 
- Support both Traditional and Simplified Chinese
- Use `locale-code` NPM module which relies on official language names and locale codes instead of our own JSON file to track language names/codes (one less thing to maintain)
- Update the script that turns Crowdin output into `react-intl` consumable translations


micahalcorn:
@wanderingstan this appears to "work". I'll defer to you on whether or not it resolves your requests.

wanderingstan:
Few notes: 

• @micahalcorn For future we should think about how to do QA on things like this. Maybe something like `testing.demo.originprotocol.com` where we first deploy versions of the dapp before replacing `demo.originprotocol.com`, and then we can point translators and testing volunteers there? 

• Looks like we have at least 3 arabic variations on there, but we're not supporting RTL languages. We should remove those. Is that something we on on crowdin? 

• Pedantic Nitpic: I'm not a fan of using "Abbrev" for what really are language codes, e.g. in variables like `langAbbrev`. It's not really an _abbreviation_ to have `zh-Hant` mean "Simplified Chinese". Could we `s/Abbrev/Code/g` those? 

• The capitalization of the names is inconsistent. Had the same problem on the website, [which we fixed](https://github.com/OriginProtocol/origin-website/blob/develop/util/misc.py#L16) with python's `.capitalize()` method. Is there a similar node method? 

image

wanderingstan:
Also: Just realizing that many of the languages on Crowdin have <20% completion (many 0%).

How best do we remove languages? On crowdin side, or is there a master list somewhere of what gets shown on the dapp? 

(I'll start going through code myself now too...)

jordajm:
@wanderingstan - the languages in the footer menu are now capitalized and the `...Abberv` variables are renamed.

I think I removed all of the right to left languages from the DApp. We can talk to @coleman415 about removing them from Crowdin until we want to support them.

As for the languages with a low percentage translated, I think @joshfraser wanted those included with machine translations for now in hopes that it would prompt translators to provide better translations. 

I think Crowdin has a way to run machine translations for all of the untranslated strings, and so we can talk to Coleman about that too. If they can be done by tomorrow, I can probably merge them into the app in time for the release tomorrow.

micahalcorn:
screen shot 2018-08-09 at 3 08 27 pm

PR #353, opened 2018-08-06
Sparrow dom/calendar prototype fixes

sparrowDom:
### Description:
- resolve some minor bugs
- modify listing reservation to use Redux like in other parts of application

### Dependant Pull Request: 
- https://github.com/OriginProtocol/origin-js/pull/303 [add confirmation callback to origin-js request listing]

Solves issue: https://github.com/OriginProtocol/origin-dapp/issues/341

### DIFF
--- a/src/components/listing-detail.js
+++ b/src/components/listing-detail.js
@@ -164,10 +164,10 @@ class ListingsDetail extends Component {
       const totalPrice = slotsToReserve.reduce((totalPrice, nextPrice) => totalPrice + nextPrice.priceWei, 0)
       this.setState({step: this.STEP.METAMASK})
       try {
-        const transactionReceipt = await origin.listings.request(this.state.address, slotsToReserve, totalPrice)
+        const listingRequest = await origin.listings.request(this.state.address, slotsToReserve, totalPrice)
         console.log('Reservation request sent.')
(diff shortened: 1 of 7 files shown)


sparrowDom:
@micahalcorn I think I wrapped my head around Redux. Could you please throw your eye on this PR and see if I got it right. 

From what I understand this is also one of the transactions we want to have present in the `TransactionsDropdown` component in a same manner as it is implemented elsewhere in the codebase.

sparrowDom:
@jordajm this PR should resolve the problems when trying to reserve a `housing` or `services` listing. I didn't find any branches on `origin-js` for the calendar prototype so I opened a separate Pull Request (link above in the description), which this branch depends on. 

One thing I noticed (that perhaps caused some of the contract errors) is that the same user that created a listing can also try to reserve it. And in the final step when `reserve` button is clicked the contract throws an error.

micahalcorn:
@sparrowDom great, thanks 👍 @jordajm do you want to confirm that this works as expected? It all looks good to me but I haven't run it against your stuff.

sparrowDom:
:+1:
@jordajm I will let you merge this PR so it doesn't impede your work on the other branch.

PR #365, opened 2018-08-08
Fix that some css properties apply only to modal content

sparrowDom:
### Description:

In https://github.com/OriginProtocol/origin-dapp/pull/303 there was an error where some of the css changes affected too many form-controls.

### DIFF
--- a/src/css/app.css
+++ b/src/css/app.css
@@ -1207,7 +1207,7 @@ a:hover {
   height: 108px;
 }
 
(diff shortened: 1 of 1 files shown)


sparrowDom:
@micahalcorn this should resolve the problems introduced with https://github.com/OriginProtocol/origin-dapp/pull/303. Sorry for causing them in the first place 😇

PR #371, opened 2018-08-09
Improve success experiences

micahalcorn:
This automatically shows the transactions dropdown when a new confirmation is underway. It also improves the probability of causing the UI to react when a listing is closed, and provides a disclaimer message to encourage a delay after creating or purchasing a listing.


src/reducers/Transactions.js
```diff
@@ -3,33 +3,32 @@ import { TransactionConstants } from '../actions/Transaction'
 const initialState = []
 
 export default function Transactions(state = initialState, action = {}) {
-    switch (action.type) {
+  switch (action.type) {
```
micahalcorn: Just indentation fixes here 👌

sparrowDom: Yeah I figured it. Github show some weird code diff here

PR #378, opened 2018-08-10
322 compact messages

Natasha08:
### Checklist:

### Description:
[Issue#322](https://github.com/OriginProtocol/origin-dapp/issues/322)

Currently, every time a user sends a message the content is displayed with the avatar, name & timestamp resulting in repetitive visual clutter. 

If the latest message is from the same sender as the previous message and the time between these messages is less than 10 minutes, then we only display the message content. Otherwise we display the message content with the other information. I created a separate component to handle compacting the messages.
￼
image


micahalcorn:
Thanks, @Natasha08 🙏 I'll take a look now.

src/css/app.css
```diff
@@ -2482,6 +2482,10 @@ form .stars img {
   margin-bottom: 20px;
 }
 
+.conversation .compact-message {
+  margin: -20px auto 20px 25px
```
micahalcorn: Missing a ; 😁

src/components/message.js
```diff
@@ -37,7 +37,7 @@ class Message extends Component {
               {moment(created).format('MMM Do h:mm a')}
             </div>
           </div>
-          <div className="message-content">
+          <div>
```
micahalcorn: We're now using this `message-content` class, so we'll need it back 😬

Natasha08:
@micahalcorn let me know what you think. I'm still working on manual re-testing -- the dApp is in a weird state where I need to enable messaging, but I am not presented with the option. I'll let you know when I have manually tested with the new changes.

micahalcorn:
As I test more, I see the reasoning for negative `margin-top`. This is tricky depending on length of the non-compact message, presence of line breaks, etc. 😐

micahalcorn:
I've made some additional tweaks but the vertical spacing isn't perfect. I think a case could be made that we should have a logical unit for the group of closely timed messages and then actually render all of those content strings inside of a parent container - for example, all within the `<div className="message-content">`. So then instead of a Message component, we would have something like a MessageGroup component containing one or more MessageContent components. But then we would have to figure out the CSS for letting the second line of message content wrap under the timestamp. Since this UI is still in flux, I think we've done all that we should do for now. Thanks again @Natasha08 for the contribution! 👍

PR #401, opened 2018-08-14
Fixes #208: Update the progress on route changes

### LINKED ISSUE #208: Existing progress bar does not start with correct state
There seems to be a minor issue where the existing profile progress bar does not fill on route changes although it does fill on browser refreshes. The colored progress bar width should always match the strength number (40% in this example).

profile-progress

anshumanv:
Fixes #208 

### Checklist:

### Description:

- The progress loader didn't start on route changes and only worked on page refreshes.

- My suggested solution: Set the progress in the state everytime the component is mounted.

- Please suggest if a better solution comes to your mind. :pray:

### DIFF
--- a/src/pages/profile/Profile.js
+++ b/src/pages/profile/Profile.js
@@ -73,6 +73,10 @@ class Profile extends Component {
 
   componentDidMount() {
     this.props.getBalance()
+    this.setProgress({
+      provisional: this.props.provisionalProgress,
+      published: this.props.publishedProgress
+    })
   }
 
   componentDidUpdate(prevProps) {


anshumanv:
Ah, I believe I should've sent the PR to the develop branch.

micahalcorn:
Thanks @anshumanv 🙏 Yeah, I tried changing the branch, but you created your branch from master. So you will need to either fast-forward your branch to be equal to develop or just start over. It may be easier to just branch from our develop branch and then push your branch directly to our repo instead of using a fork. We're planning to make this all much easier very soon.

anshumanv:
Okay sure thing, I'll just branch from the develop branch and send a new PR asap.

I hope the approach looks fine tho. :smile:

PR #430, opened 2018-08-23
Adding dependencies from calendar_prototype branch

jordajm:
This will save me many headaches as I switch back and forth to the `calendar_prototype` branch. Each time I switch branches, I have to delete and re-install my entire origin-box setup to get the dependencies to be found by webpack because of this bug:
https://github.com/OriginProtocol/origin-box/issues/34

If these dependencies (which will eventually be merged to master anyway) are here now, I can spend more time coding and less time cursing whoever invented docker...

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "popper.js": "^1.14.3",
     "react": "^16.3.2",
     "react-avatar-editor": "^11.0.2",
+    "react-big-calendar": "^0.19.2",
     "react-dom": "^16.3.2",
     "react-intl": "^2.4.0",
     "react-js-pagination": "^3.0.2",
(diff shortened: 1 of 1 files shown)


micahalcorn:
Sorry, @jordajm, I know this is frustrating. 😔 There has been some sensitivity in the past about introducing dependencies, so give me a little time to figure this out. Are you not able to just stash your package-lock.json when you switch branches? Also, it's not a great solution, but you can probably run your branch outside of the box just fine since you don't need to test the bridge integration (I don't think). That's actually what I've continued to do for the time being.

wanderingstan:
Another idea: could you have 2 origin-box directories, and just activate whichever is needed at the time?

jordajm:
Both good ideas - thanks. 
It's true that I'm not currently working in origin-js and the dapp simultaneously like I was before, so I can just run the dapp by itself until this is sorted out.

PR #442, opened 2018-08-27
Mitch/token page

mek32390:
### Checklist:

### Description:

Added token page, which still needs copy.


src/components/about-token.js
```diff
@@ -0,0 +1,135 @@
...
+              </li>
+              <li>
+                <FormattedMessage
+                  id={'about-token.earn-step-3'}
+                  defaultMessage={'Do this to earn tokens'}
+                />
+              </li>
+            </ul>
+          </div>
+          <div className="col-md-6">
```
micahalcorn: @mek32390 this should have a class with some `padding-top` instead of using line breaks.

src/components/about-token.js
```diff
@@ -0,0 +1,135 @@
...
+                defaultMessage={'About Origin Tokens'}
+              />
+            </h1>
+            <h3 className="lead">
+              <FormattedMessage
+                id={'about-token.use'}
+                defaultMessage={'What are Origin Tokens used for?'}
+              />
+            </h3>
+            <h6><strong>
```
micahalcorn: @mek32390 these opening and closing `strong` tags should be nested equally for readability.

mek32390:
@micahalcorn made some tweaks

micahalcorn:
@mek32390 same here. It looks like the mockup might have seen some iteration since you originally built this out. Do you mind revisiting it and polishing up some of the styles? There are also still some `br`s that should be replaced with padding or margin.

PR #447, opened 2018-08-29
Fix Notifications

micahalcorn:
This addresses part of https://github.com/OriginProtocol/origin-dapp/issues/445 by repairing the listing id reference and picture shown in the notification component. It also removes some old code and repairs the notifications poll.

### DIFF
--- a/src/components/messaging-provider.js
+++ b/src/components/messaging-provider.js
@@ -58,6 +58,8 @@ class MessagingProvider extends Component {
         ),
       ONE_SECOND
     )
+
+    this.notificationsInterval = null
   }
 
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


src/components/messaging-provider.js
```diff
@@ -98,6 +94,13 @@ class MessagingProvider extends Component {
       web3Account
     } = this.props
```
sparrowDom: Nitpick: this will work well since `MessagingProvider` never un-mounts when users navigate to different pages. From purely puristic view if this component were ever to be used differently where it un-mounts each initialisation of the component would trigger another interval and not clean it up when it un-mounts. One way to fix this: 
```
componentWillUnmount() {
    if (this.notificationsInterval) {
      clearInterval(this.notificationsInterval)
    }
  }
```

micahalcorn: 😳 @sparrowDom I thought about adding that, but didn't think it would be necessary. It's fair to point out that others may not recognize the need for it in the future.

PR #471, opened 2018-09-06
Fetch amounts from renamed component

micahalcorn:
This moves the balance fetching from various parent components and consolidates it to just the wallet component, which I've renamed to be consistent with some of our others.


src/components/wallet-card.js
```diff
@@ -33,6 +35,9 @@ class Wallet extends Component {
   }
 
   componentDidMount() {
+    this.props.getEthBalance()
+    this.props.getOgnBalance()
+
     this.convertEthToUsd()
```
jordajm: You may have to `await` the first two async functions because I think the `convertEthToUsd` depends on the ETH balance already being set, correct? Thinking about a race condition where sometimes it tries to convert ETH to USD before the ETH balance has been set... ?

PR #474, opened 2018-09-07
Wire up boost value to origin.js commissions

jordajm:
### Checklist:

### Description:

This wires up the boost slider to origin.js so that boost value gets stored as commission price in IPFS. 

This can be merged when approved. There is still be additional work to do in origin.js to enable payment of commissions, etc.

### DIFF
--- a/src/utils/boostUtils.js
+++ b/src/utils/boostUtils.js
@@ -1,5 +1,5 @@
 const min = 0
-const max = 100
+const max = 1000
 const range = max - min
 
 export const defaultBoostValue = 50
--- a/src/utils/boostUtils.js
+++ b/src/utils/boostUtils.js
@@ -8,27 +8,27 @@ export const boostLevels = {
   None: {
     min: 0,
     max: 0,
-    desc: 'Your listing will get very low visibility and no seller guarantee'
+    desc: 'Your listing will get very low visibility.'
   },
   Low: {
     min: min + 1,
     max: (range / 4),
-    desc: 'Your listing will get below average visibility and low seller guarantee'
+    desc: 'Your listing will get below-average visibility.'
   },
   Medium: {
     min: range / 4 + 1,
     max: (range / 2),
-    desc: 'Your listing will get average visibility and good seller guarantee'
+    desc: 'Your listing will get average visibility.'
   },
   High: {
     min: range / 2 + 1,
     max: ((range / 4) * 3),
-    desc: 'Your listing will get above average visibility and excellent seller guarantee'
+    desc: 'Your listing will get above-average visibility.'
   },
   Premium: {
     min: (range / 4) * 3 + 1,
     max: max,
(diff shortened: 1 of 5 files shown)


micahalcorn:
@jordajm I made some other changes here including connecting the wallet balance to the boost step. I noticed that the max boost level is set in both the props and the util, so we should probably consolidate those.

jordajm:
Good catch @micahalcorn - pushed a fix

src/utils/listing.js
```diff
@@ -60,6 +65,8 @@ export function originToDAppListing(originListing) {
     name: originListing.title,
     pictures: originListing.media ? originListing.media.map(medium => medium.url) : [],
     price: originListing.price.amount,
+    boostValue: originListing.commission && originListing.commission.amount,
```
franckc: If originListing.commission is defined, the data coming from originListing.commission.amount is guaranteed to be a string. Can we convert to float here so that the getBoostLevel method can always assume it gets a float ?
Maybe something like:
`boostValue: originListing.commisson ? parseFloat(originListing.commisson.amount) : 0.0`

And same for line below.

jordajm: Good call - updated

src/utils/boostUtils.js
```diff
@@ -0,0 +1,58 @@
...
+
+/*
+ * getBoostLevel
+ *
+ * @description determines the name of the boost level when given a boost value
+ * @param {number} value - the number of OGN that has is being offered for the boost
+ * @return {string} - the name of the boost level (Low, Medium, Premium, etc.)
+ */
+
+export const getBoostLevel = (value) => {
```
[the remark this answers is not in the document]
jordajm: No difference in this case... I like arrow functions in general b/c I don't have to worry about `this` referring to the scope of the function instead of the parent class, but I agree that using the `function` keyword is cleaner looking. I pushed a change b/c I like happy Frenchmen :-)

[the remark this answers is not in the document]
micahalcorn: +1 for arrows 🏹

src/utils/boostUtils.js
```diff
@@ -0,0 +1,53 @@
+export const defaultBoostValue = 500
+export const minBoostValue = 0
+export const maxBoostValue = 1000
```
cuongdo: I'm curious what this max should be. 1,000 OGN could be a lot or a little. We only have 200 OGN in our test accounts. @matthewliu do you have an opinion on this?

cuongdo: To be clear, I don't think we should block merging on this comment. I think we should figure out the max and make any necessary changes in another PR.

micahalcorn: @cuongdo we've talked about making it dynamic based on the price and/or type of listing but I don't know that we've settled on an algorithm yet. I just set it to 1,000 for testing so that we could verify the "you don't have enough token" functionality with 200 in the wallet.

PR #483, opened 2018-09-11
📙 Discovery - Use discovery server for My Offers

DanielVF:
Use discovery server for My Offers. Makes a single query to the discovery server for ids, then directly fetches the listings and offers needed.

### DIFF
--- a/src/components/my-purchases.js
+++ b/src/components/my-purchases.js
@@ -22,29 +22,28 @@ class MyPurchases extends Component {
   }
 
   async componentWillMount() {
-    const purchasesFor = await origin.contractService.currentAccount()
-    const listingIds = await origin.marketplace.getListings({ purchasesFor })
-    const listingPromises = listingIds.map(listingId => {
+    const userAddress = await origin.contractService.currentAccount()
+    const offerResp = await origin.discovery.query(`{
+      user(walletAddress:${JSON.stringify(userAddress)}){
+        offers{
+          nodes{
+            id,
+            listing{
+              id
+            }
+          }
+        }
+      }
+    }`)
+    const offerNodes = offerResp.data.user.offers.nodes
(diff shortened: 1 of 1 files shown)


franckc:
Woohoo. Nice ! Code looks good and is much simpler (and hopefully will be lot faster :)

A few questions:
1. Strategy for rolling this out smoothly.
In dev environment, this change will require developers to have ElasticSearch + EventListener + Apollo GraphQL running. Those are not started by default in origin-box yet. Same thing in prod, we'll need to have those services running in our backend.
I'd be inclined to keep the old code for a bit and have a switch based on an ENV variable that decides whether to use old or new way ? So that we can make sure everything is ok. Then we can delete old code.
2. In general I think this is an example where we could make use of a simple framework to wire-on/off features in DApp and/or Origin. Thoughts ?
3. Another consideration is whether or not we want to keep the old code as a way to run the DApp in a truly decentralized way. We had discussed the idea of having the option to configure origin js to either use a centralized back-end vs decentralized.

DanielVF:
Good points!

The way we are doing the decentralized way right before this PR is much better than the way we were doing it before that. By using the logs, we only need to load all the listings the user has ever made, rather than all the listings in the system ever. This lets it scale much further. One concern is that our distributed code is only checking the latest marketplace contract. Whenever we do a marketplace contract upgrade, all the old listings would disappear off this page. (At least, I think that's what I'm reading the code doing)

We'd need to figure out the UI / rules / customer experience for when to use the distributed method.

I'm fine holding off on this PR.

franckc:
@DanielVF Tyler is actually fixing this so once his PR #444 gets merged in we should be looping thru all the contract versions and get logs from all of them.

I'd love to get the option of fetching listings from discovery back-end before our launch. Let's discuss our strategy - sounds like a good topic we could discuss during today's weekly engineering call so I have added it to the agenda.

PR #485, opened 2018-09-11
DApp was not making valid new-style offers.

DanielVF:
DApp was not making valid new-style offers.

I've updated the fields sent to IPFS.


DanielVF:
@franckc Already did this. Removing.

PR #486, opened 2018-09-11
Redux Logger

micahalcorn:
The log messages from Redux are out of control and cannot be filtered out of the console. This disables them by default and supports a `REDUX_LOGGER` environment variable for enabling them rather than binding the condition to the environment (dev vs prod).


franckc:
You are my hero. Merci.

PR #488, opened 2018-09-12
Changed unnecessary absolute paths

balibone:
Since `WORKDIR` has already been set to `/src`, it is unnecessary to use absolute paths to target locations within this working directory. Using `./` is enough as commands following `WORKDIR` will take `/src` as the current location. 
Only when a different path outside of `/src` needs to be accessed, like `/bin/docker-truffle.sh` in `COPY bin/docker-truffle.sh /bin/docker-truffle.sh`, would we need to use absolute paths.

### Checklist:

### Description:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 FROM node:8
 ENV NPM_CONFIG_LOGLEVEL warn
 WORKDIR /src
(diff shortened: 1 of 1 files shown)


micahalcorn:
🤷‍♂️ @tyleryasaka

PR #491, opened 2018-09-12
Wiring up boost approval workflow

jordajm:
### Checklist:

### Description:

This is a change to the listing creation workflow that allows the user to:
- See the "Boost Listing" info box if they have 0 OGN
- ~Click "Start Boosting" or "Skip this step"~
- ~"Start Boosting" shows a "Get OGN" CTA~
- Once user has created a listing or expanded the tutorial, she sees the boost slider
- ~They can click an "Approve" button to approve the selected boost for transfer by the marketplace contract~
- After boost amount is selected, they can see the listing preview page


jordajm:
Working on merge conflicts now

jordajm:
merge conflicts resolved - still some formatting and stuff to fix but leaving as-is till review is complete

src/components/boost-slider.js
```diff
@@ -32,11 +38,42 @@ class BoostSlider extends Component {
...
+    const allowanceRemaining = await origin.token.getAllowance(userAddress)
+    console.log('Allowance Remaining: ', allowanceRemaining)
+    this.setState({
+      allowanceRemaining
+    })
+  }
+
+  async approveOgnTransfer() {
+    const ognAmount = this.state.selectedBoostAmount
+    const weiAmount = (ognAmount * 10 ** 18).toString()
```
franckc: Let's us BigNumber or [web3.utils.toWei](https://web3js.readthedocs.io/en/1.0/web3-utils.html#towei).
Also nit: from a crypto linguistic standpoint it is not elegant to call this **wei** since we are not dealing with ETH here but with OGN tokens... I think the crypo party recommends to call this "[natural unit](https://theethereum.wiki/w/index.php/Natural_Units)"... :)

micahalcorn: @franckc @jordajm there was some discussion in April about whether or not DApp developers should have to deal with BigNumbers and these calculations. As I recall, we intentionally abstracted all of this wei conversion to origin-js and changed all of the return values to normal integers. [I documented this at the time](https://github.com/OriginProtocol/origin-docs/commit/897c84ba35727177b3e4555bc899ee2d029c5322) but I'm not sure where the discussion happened.

micahalcorn: I understand that we don't want to use floating point, but is there any other way that the DApp can avoid having to know about wei calculations?

jordajm: I'll take a look at what it would take to move this conversion into origin-js

micahalcorn: Thanks @jordajm I didn't dig very deep, but I don't see any uses of BigNumber or toWei in origin-dapp right now. So hopefully we can handle OGN similarly to however we're handling ETH right now. 🙏

micahalcorn: Ah, there is a `fromWei` in the ETH wallet action and some math in the OGN wallet action. cc @cuongdo

cuongdo: Yeah. the OGN math should probably be in the Token resource.

micahalcorn:
This now depends on (and works with) https://github.com/OriginProtocol/origin-js/tree/dev/approveAndCallWithSender. 😁

PR #500, opened 2018-09-13
Prevent wrapping of price on listing detail

wanderingstan:
Minor change to prevent line break between eth logo and our price. 

This:

screen shot 2018-09-13 at 2 18 06 pm

becomes:

screen shot 2018-09-13 at 2 18 18 pm

downside is that when browser gets narrower, the "ETH" gets cut off. This already existed before, but is worse here:

screen shot 2018-09-13 at 2 18 26 pm


wanderingstan:
Aure gave his sign-off in #design channel.

Issue #497, opened 2018-09-13
Get started selling origin modal issues


sparrowDom:
The modal dialog saying `Get started selling origin` prevents user input in the following cases: 
- as long as dialog is shown clicking on `general search input` does not work. 
- if user opens a webpage and clicks on `notification bell icon` -> `view all button` -> `back` using browser navigation. The modal blocks clicking anywhere on the page and shows a semi transparent white backgorund (see image below)
<img width="1389" alt="screenshot 2018-09-13 15 07 42" src="https://user-images.githubusercontent.com/579910/45490784-55470680-b768-11e8-98b6-b88ed1c368a4.png">

micahalcorn:
Resolved by #502 ✅

PR #502, opened 2018-09-14
Natasha/modal blocking ui

Natasha08:
### Checklist:

- ~~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs)~~

### Description:
[Issue #497](https://github.com/OriginProtocol/origin-dapp/issues/497)

Bugs:
- The default modal component blocks a user from interacting with input fields
  - the attribute `tabIndex='-1'` was the source of this behavior
- Navigating back to the Listings page using browser navigation => the getting started modal blocked the UI completely while open.

Fixes:
- refactor the modal component to take a `tabIndex` prop && default this to `-1`
- update the onboarding index component:
  - to wait for a wallet to load if it exists
  - to no longer show the modals if the user has completed all the steps
    - this means that currently a user can 'go through all the steps', not have a wallet and the modals will not show. Once the functionality of the modals have been implemented this should work as intended.
    - to reset the behavior, delete any onboarding data in localStorage
  - to no longer show the modals if the user has a `wallet.address`

image

### DIFF
--- a/src/components/onboarding-modal/index.js
+++ b/src/components/onboarding-modal/index.js
@@ -19,13 +19,13 @@ class OnboardingModal extends Component {
     this.closeModal = this.closeModal.bind(this)
   }
 
-  async componentWillMount() {
-    const { fetchSteps, onboarding, wallet } = this.props
+  async componentDidMount() {
+    const { fetchSteps } = this.props
     await fetchSteps()
 
-    if (!onboarding.stepsCompleted && !wallet.address) {
+    window.setTimeout(() => {
       this.userProgress()
-    }
+    }, 500)
   }
 
   componentWillUpdate(nextProps) {
@@ -36,7 +36,9 @@ class OnboardingModal extends Component {
     if (splitPanel) {
       this.addModalClass()
     } else {
-      this.removeModalClasses()
+      window.setTimeout(() => {
+        this.userProgress()
+      }, 500)
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/onboarding-modal/index.js
```diff
@@ -19,24 +19,26 @@ class OnboardingModal extends Component {
...
-    const { fetchSteps, onboarding, wallet } = this.props
+  async componentDidMount() {
+    const { fetchSteps } = this.props
     await fetchSteps()
 
-    if (!onboarding.stepsCompleted && !wallet.address) {
+    window.setTimeout(() => {
       this.userProgress()
-    }
+    }, 500)
```
sparrowDom: Half a second wait is probably required for the Dapp to register if wallet is present and `wallet` property is available through redux. On some machines it might take less than 500ms, on the other hand on others(that are really slow) it might take more. 

Another approach would be to mark when a wallet is initialised ( https://github.com/OriginProtocol/origin-dapp/blob/master/src/reducers/Wallet.js ): 
```
const initialState = {
  address: undefined,
  ethBalance: '0',
  ognBalance: '0',
  initialised: false
}

export default function Wallet(state = initialState, action = {}) {
  switch (action.type) {
  case WalletConstants.INIT_SUCCESS:
    return { ...state, address: action.wallet, initialised: true}
...
```

and call `userProgress` from `componentWillUpdate` only once `wallet.initilised === true`

Natasha08:
@sparrowDom Thanks 😄  and thank you for the feedback! I have updated the PR, let me know what you think.

Issue #393, opened 2018-08-14
Messaging: Disable If Web3 Is Not Present


micahalcorn:
Similar to the way that we handle the buy button in Safari, messaging features should be disabled if an ETH account cannot be detected.

Natasha08:
@micahalcorn can you assign this to me?

micahalcorn:
@Natasha08 done thx 🙏

Natasha08:
@micahalcorn is this the solution you're referring to for Safari? ![image](https://user-images.githubusercontent.com/17957496/45569782-d6aca080-b81d-11e8-92de-c9df9fd67e90.png)

micahalcorn:
Resolved by #506 ✅

PR #506, opened 2018-09-15
Natasha/disable messaging

Natasha08:
### Checklist:

- ~~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs)~~

### Description:
[Issue #393](https://github.com/OriginProtocol/origin-dapp/issues/393)

Problem:
- Disable messaging when a user does not have an account/not logged in
- [ ] feedback about messaging

Solution:
- Dispatch `storeWeb3Intent`:
   - when a user tries to access messages from the navbar
      - also disable the messages dropdown if they are don't have an account
  - when a user tries to contact a seller about a listing

**Message from the navbar (when no web3 account exists)**

image

**Message when contacting a seller (not logged in)**

image


micahalcorn:
Thanks @Natasha08 💬 As previously discussed, we don't have a good solution to those partial strings that [won't really work for localization](https://github.com/OriginProtocol/origin-dapp/issues/372).  This PR is at least a step in the right direction for mobile and Safari users. I made some changes, and we will have to come back to this when one of my other commits gets merged and exposes "contact" links more frequently.

PR #518, opened 2018-09-18
OGN tooltip

mek32390:

### Checklist:

- [ ] ~~Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs)~~

### Description:

This is a new clean PR for the OGN tooltip. OGN balance still needs to be passed in to the tooltip from Redux. cc: @micahalcorn


micahalcorn:
Mockup: https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5b7538aa755a0b576b86c981

src/components/wallet-card.js
```diff
@@ -23,6 +24,30 @@ class WalletCard extends Component {
...
+        id: '_wallet-card.balanceText',
+        defaultMessage: 'Having OGN is not required but will allow you \
+        to create a listing that will be more visible to buyers.'
+      },
+      getOGN: {
+        id: '_wallet-card.getOgn',
+        defaultMessage: 'Get OGN'
+      },
+      reccommendation: {
+        id: '_wallet-card.reccommendation',
```
Natasha08: `recommendation` here and line above

src/components/wallet-card.js
```diff
@@ -282,4 +360,4 @@ const matchDispatchToProps = dispatch => ({
 export default connect(
   mapStateToProps,
   matchDispatchToProps
-)(WalletCard)
+)(injectIntl(WalletCard))
```
Natasha08: needs an extra line here

micahalcorn:
@Natasha08 I haven't taken the time to look at this closely. Is there a more React-y way of handling the events rather than such heavy jQuery?

Natasha08:
@mek32390 
Manually testing from the listings page, when I click on a listing => mouseover the tooltip balance heading I get this 
cursor_and_origin_protocol_demo_dapp

micahalcorn:
> @mek32390
> Manually testing from the listings page, when I click on a listing => mouseover the tooltip balance heading I get this
> cursor_and_origin_protocol_demo_dapp

Sounds like we need to condition the rendering based on the parent component?

Natasha08:
@micahalcorn I think that [React's Event System](https://reactjs.org/docs/events.html) could be used here instead of JQuery, and the component can determine whether or not to render things (like the tooltip)

micahalcorn:
This still needs some fixes for the UX but I went ahead and merged it. We at least need to fix the hide delay so that the "Learn More" can be clicked. We probably also want it to appear programmatically and remain visible until dismissed by the user. @Natasha08 does it make sense to just go ahead and move all our tooltips away from jQuery via some established library?

Issue #400, opened 2018-08-14
Create "supported languages" list for translations script


jordajm:
The script that we use to process translated string translated string is here:
https://github.com/OriginProtocol/origin-dapp/blob/develop/translations/scripts/processTranslatedMessages.js

We need to modify that script so that that it will only include supported languages in `translated-messages.json` instead of just pulling in whatever Crowdin sends in the auto-generated PRs. 

This will prevent us from having to manually delete unsupported or incomplete languages before merging the Crowdin PRs. After this chance, we should just be able to merge the Crowdin PRs, run the script, and commit the results with no manual intervention.

List of supported languages:
```
const supportedLocales = [
  'de-DE',
  'el-GR',
  'es-ES',
  'fil-PH',
  'fr-FR',
  'hr-HR',
  'id-ID',
  'it-IT',
  'ja-JP',
  'ko-KR',
  'nl-NL',
  'pt-PT',
  'ro-RO',
  'ru-RU',
  'th-TH',
  'tr-TR',
  'uk-UA',
  'vi-VN',
  'zh-CN',
  'zh-TW'
]
```

PR #521, opened 2018-09-19
Issue #400 - Only process strings from supported locales

jordajm:
### Checklist:

### Description:

Updates the script we use to import new translations from Crowdin so it only pulls in strings from supported locales instead of just pulling in whatever Crowdin has in its auto-generated PR. 

This should speed up releases by preventing some/most of the cleanup work that was required to merge new strings from Crowdin.


wanderingstan:
@jordajm  Note on l18n for future reference: Learned that the setting we want from CrowdIn is actually the poorly-named `%osx_locale%`, which uses the BCP 47 language format used by Javascript, HTML, web browsers, and our own listing JSON. BCP 47 only includes the locale modified when needed, e.g. English is simply `en`, but Simplified Chinese is `zh_Hans`. Then we wouldn't need that extra logic to first try the full code, then without locale.

jordajm:
Good find @wanderingstan - I'll make an issue for that. CC @micahalcorn

jordajm:
Created: https://github.com/OriginProtocol/origin-dapp/issues/526

Issue #54, opened 2018-01-08
Submitting listings on testnet are appearing in wrong order


wanderingstan:
May be related to the earlier bug where it reverse direction on each reload.

PR #523, opened 2018-09-19
Public Profile Reviews

micahalcorn:
This re-implements the reviews on the public profile page. It also makes some improvements to the my-listings, my-purchases, and my-sales components.

### DIFF
--- a/src/pages/user/User.js
+++ b/src/pages/user/User.js
@@ -48,7 +48,6 @@ class User extends Component {
 
   render() {
     const { user, userAddress } = this.props
-    const { reviews } = this.state
     const { attestations, fullName, profile } = user
     const description =
       (profile && profile.description) || 'An Origin user without a description'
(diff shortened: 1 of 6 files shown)


src/pages/user/User.js
```diff
@@ -22,22 +24,34 @@ class User extends Component {
...
-      this.props.userAddress,
-      this.props.intl.formatMessage(this.intlMessages.unnamedUser)
-    )
-    // TODO: User reviews
+  async componentDidMount() {
+    const { fetchUser, intl, userAddress } = this.props
+
+    fetchUser(userAddress, intl.formatMessage(this.intlMessages.unnamedUser))
+
+    const listingIdsAsSeller = await origin.marketplace.getListings({
```
franckc: Why fetching those seller side reviews since we then filter them on line #54 ?

micahalcorn: @franckc unless I misunderstood how this is working, we need to fetch all the reviews from listings where the user was buyer or seller. For each listing, that will return up to two reviews. Then we need to filter them for basically every other review. In other words, just because the user was a seller on a particular listing doesn’t mean that the review will get filtered out. 🧐

franckc: Yes, you make sense and I don't !

PR #525, opened 2018-09-19
Fix subcategory translations.

DanielVF:
The dapp is trying to translate the subcategory string when displaying it, but only part of the subcategory string is being saved. After discussion here, I think we need to save the full subcategory translation field in the string. https://github.com/OriginProtocol/origin-js/pull/472

### DIFF
--- a/src/utils/listing.js
+++ b/src/utils/listing.js
@@ -12,7 +12,10 @@ import origin from '../services/origin'
 export function dappFormDataToOriginListing(formData) {
   // formData.category data format is "schema.<category>.<subCategory>".
   const subCategory = formData.category
-  const category = formData.category.split(".").slice(0,2).join('.')
+  const category = formData.category
+    .split('.')
+    .slice(0, 2)
+    .join('.')
 
   const listingData = {
     category: category,
@@ -87,7 +90,7 @@ export function originToDAppListing(originListing) {
 export async function getListing(id, translate = false) {
   const originListing = await origin.marketplace.getListing(id)
   const dappListing = originToDAppListing(originListing)
-  if(translate) {
+  if (translate) {
     dappListing.category = translateListingCategory(dappListing.category)
   }
   return dappListing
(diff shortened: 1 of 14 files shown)


src/utils/listing.js
```diff
@@ -11,7 +11,8 @@ import origin from '../services/origin'
  */
 export function dappFormDataToOriginListing(formData) {
   // formData.category data format is "schema.<category>.<subCategory>".
-  const [category, subCategory] = formData.category.split('.').slice(1)
+  const subCategory = formData.category
+  const category = formData.category.split('.')[1]
```
sparrowDom: These are translation strings Dapp uses for categories: https://github.com/OriginProtocol/origin-dapp/blob/master/src/utils/listingSchemaMetadata.js 

Might be enough to lookup the object with the matching `type` property.

franckc: Shouldn't category be `${splits[0]}.${splits[1]}` ?
So that for example if subcategory is schema.forSale.antiques then category will be schema.forSale

franckc:
Darn linter. Build is failing due to this:
/home/travis/build/OriginProtocol/origin-dapp/src/utils/listing.js
  15:44  error  Strings must use singlequote  quotes

DanielVF:
I've increased the scope of this PR to make the my-listings page and my-purchases page show the translated DAPP category/IPFS subcategory.

src/utils/translationUtils.js
```diff
@@ -291,22 +291,15 @@ export function translateSchema(schemaJson, schemaType) {
...
-      }
-    }
+export function translateListingCategory(rawCategory) {
+  const match = rawCategory.match(/^schema\.([^.]+)\.([^.]+)$/)
+  if (match === null) {
+    return rawCategory
   }
-
-  return listing
+  const [_, schemaType, category] = match
```
franckc: I'm with you, it is so irritating we can't use _ for unused variables.
But the build is failing due to eslint complaining:
  299:10  error  '_' is assigned a value but never used         no-unused-vars
  299:25  error  'category' is assigned a value but never used  no-unused-vars

DanielVF: I didn't realize at first that the dapp had the formatting/lint check under npm test. I'll run that in the future.

PR #544, opened 2018-09-21
display error in console when deploying profile fails

sparrowDom:
### Description:

On https://demo.staging.originprotocol.com/#/profile deploying the profile fails and we do not know why. Lets add an error log too see what the problem is


sparrowDom:
Ahh my bad closing this and making the fix to master. (saw @tomlinton 's message on Discord)

PR #546, opened 2018-09-21
Change claimType to topic

sparrowDom:
### Description:
Recently `claimType` attestation variable in `origin-js` has been renamed to `topic` ( https://github.com/OriginProtocol/origin-js/blob/47efcc0bffd57954cf945ba3fdfd9a1f6bf2eb5c/src/resources/attestations.js#L23 ). This prevents attestation from becoming provisional and being published on the blockchain. This PR reflects the change in `origin-js`.


franckc:
Great catch ! @sparrowDom as a separate PR could you add a unit test so that we catch this automatically if it breaks again in future ?

franckc:
oh darn. this is DApp... we don't have unit tests yet :( nevermind regarding my previous comment @sparrowDom  !

tyleryasaka:
Thanks for catching this @sparrowDom . My bad for forgetting this.

PR #556, opened 2018-09-23
Added page for developer information about DApp.

wanderingstan:
Simple page to output info for developers. Will be useful when debugging, especially around releases. For now just outputs ENV vars and origin.js version, but there are surely many more things we could add. 

Screenshot:

image


franckc:
This is SO helpful ! Genius idea @wanderingstan :)

PR #561, opened 2018-09-23
Wallet, Status, Card, et al, Improvements

micahalcorn:
- updates the `WalletCard` to reference `user` rather than `profile`
- moves `'Unnamed User'` out of Redux and consolidates various components into one
- replaces clock emoji with animated loading indicator
- accounts for offer status in listing cards and details
- updates the "my _____" sections with better copy and timestamps


wanderingstan:
Some formatting fixes needed, i'm working on it now.

src/components/my-sale-card.js
```diff
@@ -104,7 +104,7 @@ class MySaleCard extends Component {
               </h2>
               <p className="address text-muted">{user.address}</p>
               <div className="d-flex">
-                <p className="price">
+                <p className="price" data-createdAt={ createdAt }>
```
micahalcorn: @wanderingstan why dis? 😕

wanderingstan: Linter bitches about `createdAt` not being used. My limited javascript and a quick search couldn't find a way to throw away the value retrieved for `createdAt` at line 53. Do you know a way? Figured we might want to show the created date at some other point, so no harm in putting it in a element prop.

micahalcorn: Well, the short answer is just to [remove it](https://github.com/OriginProtocol/origin-dapp/commit/53484a4385c28931e0c2410bf2aff7ab47a96fab) if it's not being used. But it begs the question whether or not we're supposed to be showing a timestamp on that cart. I'll confirm.

PR #577, opened 2018-09-25
Show modal confirmation on transactionHash event instead of transaction receipt

sparrowDom:
### Checklist:

- [ ] Needs testing on Rinkeby

**Connected PR:** https://github.com/OriginProtocol/origin-js/pull/546
**Solves Issue:** https://github.com/OriginProtocol/origin-dapp/issues/571#event-1864295404

### Description:
Deploying profile changes takes too long on Rinkeby (~40 seconds) and it might take longer on mainnet. 
For that reason these are the proposed changes to the Dapp:
When deploying a profile Dapp now waits for transactionHash instead of transaction receipt. This way the user doesn't need to wait for the transaction to be mined in order to continue using the Dapp. On deployment of profile a new dialog is opened that lets a user know a transaction is waiting to be mined and provides a link to Etherscan:
screenshot 2018-09-25 16 33 01


micahalcorn:
Thanks @sparrowDom, is it too much trouble to just wire this up to our blockchain transactions dropdown? It may not make sense if origin-js isn't prepared to handle the confirmation callback like we do in marketplace contract-service.

sparrowDom:
@micahalcorn if I understand correctly, currently we show transactions in the transaction dropdown that have already been mined, and would be cool, to also be able to show transactions that have not yet been mined instead of directing users to etherscan?

sparrowDom:
@micahalcorn Aha I see. I will change the behaviour so that the dropdown gets populated with transaction information as soon as we receive the hash of the transaction (before the receipt).

The problem with waiting was that by my measurement the DApp hangs on this window for 40 seconds (on Rinkeby) waiting for transaction receipt after the user clicks Confirm in MetaMask:
screenshot 2018-09-25 18 06 15

This might take longer on Mainnet.

micahalcorn:
Thanks @sparrowDom that may serve as a good proof-of-concept for what we have been discussing. It's not critical, but it would be nice to confirm that the user's attestations and profile information are maintained the Redux profile "provisional" state between the time that the hash is received and the receipt is produced so that she can navigate away from the profile and back to it without temporarily losing the inputs.

sparrowDom:
@micahalcorn yup good point. 
Let me do the necessary changes relating to the transactions dropdown and then we can test it out and see if it makes sense.

sparrowDom:
@micahalcorn can you take another look please? Transaction now appears in transactions dropdown and disappears after 12 confirmations (like before). And the changes in the profile go from provisional to published after the first confirmation.

sparrowDom:
It should... though I have been testing it on Rinkeby. I wonder if the behaviour on Main-net stays the same

PR #585, opened 2018-09-26
Stan/dapp info

wanderingstan:
A little css love for dapp-info before showing it in ENG call, plus beginnings of a messaging section. 

image


micahalcorn:
I was holding back. 😁

PR #586, opened 2018-09-26
Create listing - max listing size help text

jordajm:
### Checklist:

### Description:

- Add help text saying `'Total listing size may not exceed 2MB'`
- Localize all strings in the image-upload component

screenshot 2018-09-25 21 55 10

Issue: https://github.com/OriginProtocol/origin-dapp/issues/410


jordajm:
@micahalcorn - the issue https://github.com/OriginProtocol/origin-dapp/issues/410 didn't have a super-clear task in it, so I took a whack at it. LMK if you have/had other ideas for this one.

micahalcorn:
@jordajm it looks like there may actually be more work to do here than I realized. 😇 I glossed over the current state of this implementation. It's still frustrating to have to deal with calculating the combined image sizes + some unknown size of the other form data. Can you look into https://github.com/OriginProtocol/origin-dapp/issues/489 and see if it's reasonable to quickly change [the size validation](https://github.com/OriginProtocol/origin-dapp/blob/master/src/components/listing-create.js#L232) to allow multiple 1.5MB images, for example? I also think that it might be helpful to give an in-form error rather than waiting for the user to click "Continue" before informing them that the data is too large, but that's not critical.

PR #587, opened 2018-09-26
Conversation Improvements

micahalcorn:
This sorts the conversation summaries list and prevents the listing summary from sticking when changing conversations. Can be DRY'd up later 😬


src/components/messages.js
```diff
@@ -99,9 +99,20 @@ const mapStateToProps = ({ app, messages }) => {
   const filteredMessages = messages.filter(({ content, conversationId }) => {
     return content && origin.messaging.getRecipients(conversationId).includes(web3Account)
   })
+  const conversations = groupByArray(filteredMessages, 'conversationId')
```
franckc: Any chance this sorting function could get be shared so that the code does not get duplicated ?
Like maybe it could go in a file utils/comments.js >]?

micahalcorn: Yes, that's the DRY plan. 🐪

PR #592, opened 2018-09-26
Sync with origin-box setup

cuongdo:
### Checklist:

### Description:

This syncs the various addresses with OriginProtocol/origin-box#89 for
those of us in still in the non-box stone ages.


franckc:
@micahalcorn I do think we should set an affiliate by default as a convenience for developers - without it you can't make an offer. Also will be good to have that for creating sample listing with offers automatically. If you are concerned with accounts having multiple roles, we could spread them the following way perhaps ?
Account[0] = seller
Account[1] = buyer
Account[2] = affiliate + faucet distributor
Account[3] = messaging
Account[4] = arbitrator

micahalcorn:
Ah, the sample offers is is a good argument. Yeah, I would prefer that each be a dedicated account. 🙏

cuongdo:
Closed in favor of #611

PR #597, opened 2018-09-27
Check individual image size, not size of whole listing.

jordajm:
### Checklist:

### Description:

- Add warnings for large images and too many images
- Add "remove" button for images and warnings
- Here are two big images for testing 😇 : 
https://drive.google.com/drive/folders/1KXs63LIYZRG-GT7wNoACj3Y2QKOxScAo?usp=sharing

screenshot 2018-09-26 23 45 03

---------------------------

screenshot 2018-09-26 23 45 25

----------------------------

screenshot 2018-09-26 23 45 40


franckc:
@jordajm You are awesome. Thanks for working on this.
Just a note that there is a subtlety regarding the 2MB limit: when encoded in the data URL, the image content is base64 encoded, which increases its size (the ratio if 4/3). Then at the origin-js layer, right before we submit it to the IPFS gateway, we decode it. So when you test the size of the Data URI in the DApp you could allow up to 2MB*4/3 = 2.66MB

micahalcorn:
Great work, @jordajm 💙 thx so much.

PR #601, opened 2018-09-27
Limit to only one supported network, configurable

wanderingstan:
Default to mainnet. Remove hardcoded reference to Rinkeby.
Updates to dapp-info.

EDIT: Fix for: https://github.com/OriginProtocol/origin-dapp/issues/583

### Checklist:

_Defaults to mainnet. ENV var only needed on production, and only when it's not mainnet._


wanderingstan:
@franckc will do! I don't think we need it for dev, because it won't be a production environment (`process.env.NODE_ENV` will not be `production`) so the network check is skipped.

franckc:
@wanderingstan Yes, dev is the continuous integration environment running in Google Cloud (we are not using it yet but we planned to after V1 is released). I know it is a big confusing...

franckc:
Ah also, do you mind taking a look at those conflicts ? I think due to Micah's big reformatting :)

wanderingstan:
Not a fan of the formatting ... asking if we can revert it simple one-line-per-table-row

micahalcorn:
I resolved the conflict, at least. 😬

PR #602, opened 2018-09-27
Format

micahalcorn:
This should cover it. 😑


wanderingstan:
I know we're moving fast, but for the record, I think a PR touching so many files in so many ways should have had more reviewers and more discussion... /grumpy-old-man

PR #604, opened 2018-09-27
Update exchange rates/prices every 2 minutes

jordajm:
### Checklist:

### Description:

- This moves the exchange rate requests to an interval that updates redux instead of relying on component re-renders to trigger exchange rate refreshes.

- Only hits data provider API once every 2 minutes

- It also prepares our exchange rate data structure for multiple currency pairs in expectation of moving beyond just USD/ETH at some point

- Data structure in redux ends up looking like this:
screenshot 2018-09-27 14 51 37

And the `exchangeRates` object, as well as the action and reducer that populates it, is designed to handle multiple currency pairs, which would become new keys in the object  like `USD/OGN`, etc.

Issue: https://github.com/OriginProtocol/origin-dapp/issues/425


micahalcorn:
@jordajm polling would be for realtime reactive updates. That isn't critical for v1, tho.

jordajm:
Ahh ok, right, this depends on the user interacting with the app somehow to cause components to re-render in order to update pricing - so in the case where a user is sitting on a page without doing anything, polling would allow the prices to stay fresh. Makes sense. 

Pls don't merge this yet and I'll add that to this PR.

PR #614, opened 2018-09-28
⚠️ Beta Warnings

micahalcorn:
This is a rough proof of concept for the Mainnet Beta warning, which will take two forms:

- blocking modal appearing on initial load
- reminder alert persisting after modal is dismissed

**To Do**

- [x] replace the modal's image when ready
- [x] ~prevent the seller onboarding modal from appearing until this one is dismissed~
- [x] style the modal text
- [ ] ~consider allowing dismissal of alert~ 👈 NOT WANTED
- [x] render alert on every route
- [ ] add missing `modal-backdrop` 👈 nice to have
- [ ] highlight some words as depicted [here](https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5ba521bd412fb1188c9f33c3)


micahalcorn:
@franckc this is now in a mergeable state. I ended up including a variety of copy and UX changes that are not specific to the warnings resulting from my sessions today with Matt. There will be more to come, but I thought I would try to limit the scope of this one as much as is reasonable.

I will also make another issue for the other two slight improvements left unchecked. 🔲

Issue #595, opened 2018-09-27
Invalid Netword Redirect


micahalcorn:
If a user's Web3 provider is connected to the wrong network, we currently render a blocking modal and suggest that she change it. Instead, we should redirect her to the environment that will work with her selected network.

wanderingstan:
Looking into this now...

franckc:
@wanderingstan Our Rinkeby DApp is at https://demo.staging.originprotocol.com.

PR #618, opened 2018-09-28
Issue/595 network redirect

wanderingstan:
_? Should these fixes be going to staging? I can fix this PR to do that._

### Checklist:

### Description:

Fixing #595 - redirect to our dapp on other domains corresponding to diff eth networks.


src/components/web3-provider.js
```diff
@@ -441,6 +443,17 @@ class Web3Provider extends Component {
     const isProduction = process.env.NODE_ENV === 'production'
     const networkNotSupported = supportedNetworkId != networkId
 
+    // Redirect if we know a DApp instalation that supports their network.
+    if (currentProvider && networkId && isProduction && networkNotSupported) {
+      const url = new URL(window.location)
+      if (networkId==1) {
```
micahalcorn: I've pushed for use of strict equality in the past. Unless those env vars come out as strings, I would favor using it here.

src/components/web3-provider.js
```diff
@@ -21,6 +21,8 @@ const networkNames = {
   999: 'Localhost'
 }
 const supportedNetworkId = process.env.ETH_NETWORK_ID || 1 // Default to mainnet
+const mainnetBaseUrl = process.env.MAINNET_BASEURL || 'https://dapp.originprotocol.com'
+const rinkebyBaseUrl = process.env.RINKEBY_BASEURL || 'https://demo.originprotocol.com'
```
franckc: Our rinkeby setup is at https://demo.staging.originprotocol.com

wanderingstan:
If the tests pass I'm good to merge. 

(BTW, talking with John, I'm thinking we should move all default env vars into webpack.config.js -- better to have them all in one place. But not something that has to happen right away.)

franckc:
@wanderingstan yes agreed putting all env vars in webpack.config.js would be good. sounds like a clean up for after release...

PR #629, opened 2018-09-28
Issue/593 large images

wanderingstan:
### Checklist:

### Description:

Fix https://github.com/OriginProtocol/origin-dapp/issues/593

Don't show images above a given (env-configured) size. Default to 2 MB (2 * 1024 * 1024)

Looks like:
image


micahalcorn:
Thanks @wanderingstan 😕 In my experience, this doesn't seem to actually prevent the browser from locking up if a huge chunk of data is coming across. There may actually be an unrelate over-rendering issue that is contributing to the bad experience also, but I haven't had time to investigate that.

We should also validate the file selection so that we can prevent the sender from creating the problem.

micahalcorn:
I wonder if this needs to be solved at the origin-js level or if there is even a way to prevent loading all the data from Orbit. 🤔 @crazybuster

wanderingstan:
Well that sucks. 

I believe that our IPFS node already blocks uploads >2MB, so in theory you shouldn't be able to upload huge images like that. 

❗️There is an option in `fetch()` to limit the max size of data fetched!!
https://www.npmjs.com/package/node-fetch#api

We use it here: https://github.com/OriginProtocol/origin-js/blob/master/src/services/ipfs-service.js#L118


I'm working on a PR now to add this to origin-js

Maybe also: In IPFS config: Limit size of data that our IPFS node will server up (not sure this is even possible)

micahalcorn:
@wanderingstan I’m now realizing that in production this will only affect the sender. When I’m testing locally, I’m not getting the benefit of the file size limitation in Nginx. And the sender is putting the data in her local Orbit because we don’t have the client-side size restriction. As long as we prevent the initial upload, the production server should protect us from malicious senders who circumvent the DApp. It will refrain from persisting the messages and so the intended recipient won’t get them.

wanderingstan:
@micahalcorn Just added some commits that check message image size when image is selected and when 'send' is pressed.

Works BUT: we are now testing on actual image size (when selected) and then on base64-encoded size (when user hits send).

I'll be out the next few hours. Please feel free to fix and modify!

micahalcorn:
Grat, thx @wanderingstan 👍 I've added a commit for a more pleasant, translatable error message. We could also go ahead and validate the file type if you want. Otherwise take a look at the error message and see if it is accurate in regard to file size.

wanderingstan:
Added commit that 
(1) factors in the 4/3 image expansion when it's encoded in base64
(2) changes text to say max image size is 1.5MB, which is roughly the size of image that we can store —base64 encoded— with our 2MB limit on IPFS.

micahalcorn:
Thanks @wanderingstan, good to merge?

wanderingstan:
Good by me. If Travis is happy, lets merge.

PR #645, opened 2018-10-01
Implement a simple way to feature/hide listings

sparrowDom:
### Checklist:

- ~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs)~

### Description:
A simple version of hidden/featured listings implementation. 

**Solves issue:** https://github.com/OriginProtocol/origin-dapp/issues/643#event-1875878859


src/actions/Listing.js
```diff
@@ -4,15 +4,42 @@ import keyMirror from 'utils/keyMirror'
 
 import origin from '../services/origin'
 
+const featuredListingsGist = 'https://rawgit.com/sparrowDom/23f8b219567811221cba660039c7e438/raw/5610030ef20ba79245dbdb10af18d9d10c8fb98d/gistfile1.txt'
```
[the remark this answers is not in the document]
wanderingstan: Yes. And rather than a gist, why not point to the `hide_list` [branch](https://github.com/OriginProtocol/origin-dapp/tree/hide_list) as we did before? Then anyone can propose changes with PRs, and anyone with git priveledges can merge.

src/actions/Listing.js
```diff
@@ -8,11 +8,42 @@ export const ListingConstants = keyMirror(
...
+  const readListingsFromGist = async (gitstUrl) => {
+    const response = await fetch(gitstUrl)
+    return (await response.text())
+      .split(',')
+      .map(listing => listing.trim())
+  }
+
+  return async function(dispatch) {
+    try{
+      if (process.env.FEATURED_LISTINGS_GIST === undefined || process.env.HIDDEN_LISTINGS_GIST === undefined){
```
wanderingstan: I believe these env vars are no longer needed not that we pull from git branch.

sparrowDom: jp true thanks!

src/actions/Listing.js
```diff
@@ -8,11 +8,42 @@ export const ListingConstants = keyMirror(
...
+        console.error('Add environmental variables "FEATURED_LISTINGS_GIST" & "HIDDEN_LISTINGS_GIST" to enable hidden/featured listings functionality')
+        return
+      }
+
+
+      // the ?cache= part is for cache bust (so that previous results are not kept in cache)
+      const featuredListings = await readListingsFromGist(`https://raw.githubusercontent.com/OriginProtocol/origin-dapp/hidefeature_list/hidelist_${networkId}.json`)
```
wanderingstan: I'd say to remove reverences to "Gist" as that's a (now untrue) implementation detail. 
- `readListingsFromGist ()` --> `readListingsFromUrl()`
- `gitstUrl` --> `listingsUrl`
etc

src/components/app.js
```diff
@@ -108,8 +108,8 @@ class App extends Component {
   }
 
   componentDidUpdate() {
-    if (this.props.networkId !== undefined && !featuredhiddenListingsFetched) {
-      featuredhiddenListingsFetched = true
+    if (this.props.networkId !== null && !this.featuredhiddenListingsFetched) {
```
wanderingstan: Just a js question: Is it considered better these days to explicitly test for `null` vs `undefined` vs. something like
```
if (this.props.networkId && !this.featuredhiddenListingsFetched) {
```
I know people are better about using strict comparison operators now.

wanderingstan:
@micahalcorn @franckc  This is good to go by me—merge whenever is good on your end.

src/components/app.js
sparrowDom: In this case I compared it to null because default value of `networkId` is null in the reducer: https://github.com/OriginProtocol/origin-dapp/blob/master/src/reducers/App.js#L17

Other than that I guess your code would also work fine. I just don't like the implicit conversions javascript does where `if (variable)` that variable could be: 
- null
- undefined
- 0
- false
- empty string
- NaN

and it would all convert to false. I guess this way is more informative to the one reading the code what is the default value instead of being one of 6 above. 

On the other hand the code in your comment is less cluttered. There were probably a bunch of flame wars about this topic 🚒

wanderingstan:
Note for future: If we keep this method, just realized that the network ID is actually part of the listing ID (e.g. `999-000-4,999-000-3`), so we could just have one file that contained _all_ listings to hide/feature across all networks.

wanderingstan:
So cool, 
https://github.com/OriginProtocol/origin-dapp/blob/hidefeature_list/featurelist_999.txt
yields this display: 

image

src/components/app.js
wanderingstan: Thanks! I'm returning to the js world after years away and catching up on what has been settled, and what flame wars are still burning! 🔥

sparrowDom:
@wanderingstan yeah good point much simpler having only 2 files. If we were to change files to `.json` we could have everything in 1 file.

I did create a bug with this branch 🙈 where empty listing file results in an invalid empty string listing-id. Fix is here: https://github.com/OriginProtocol/origin-dapp/pull/654

PR #651, opened 2018-10-02
Modal & Messaging

micahalcorn:
- updates the beta modal with new graphics and copy
- repairs a bug that I discovered in messaging today.

😬


micahalcorn:
I don't know WTF happened here 🙃

PR #652, opened 2018-10-02
Modal & Messaging

micahalcorn:
- updates the beta modal with new graphics and copy
- repairs a bug that I discovered in messaging today
- allows image removal in conversation form

😬


src/components/conversation.js
```diff
@@ -91,7 +93,7 @@ class Conversation extends Component {
     for (const key in filesObj) {
       if (filesObj.hasOwnProperty(key)) {
         // Base64 encoding will inflate size to roughly 4/3 of original
-        if ((filesObj[key].size / 3 * 4) > imageMaxSize) {
+        if ((filesObj[key].size / 3) * 4 > imageMaxSize) {
```
jordajm: Check out @franckc 's comment here and explanation right above that: https://github.com/OriginProtocol/origin-dapp/pull/597#issuecomment-424972125
I think you can allow the actual file size to be up to 2MB because it gets decoded in origin-js layer before it gets submitted to IPFS - no?

src/components/conversation.js
```diff
@@ -365,18 +376,32 @@ class Conversation extends Component {
...
-                <p className="text-danger" onClick={() => this.setState({ invalidFileSelected: false })}>
+                <p
+                  className="text-danger"
+                  onClick={() => this.setState({ invalidFileSelected: false })}
+                >
                   <FormattedMessage
                     id={'conversation.invalidFileSelected'}
-                    defaultMessage={'File sizes must be less than 1.5 MB. Please select a smaller image.'}
+                    defaultMessage={
+                      'File sizes must be less than 1.5 MB. Please select a smaller image.'
```
jordajm: If my above comment about file size is correct, this can be updated to 2MB

PR #656, opened 2018-10-02
Add block epoch & pass to origin.js

cuongdo:
Stopgap for speeding up scanning of events for the marketplace contract,
particularly for Rinkeby and Mainnet.

### Checklist:


cuongdo:
Also, this PR has a change to avoid calling `getListings` multiple times. When I added debug logs, I saw that `getListings` was being called ~8x for each page load. This is wasteful and slows things down.

micahalcorn:
Awesome, thx 👏 I've been meaning to investigate all of the over-fetching and over-rendering that we're surely doing at the browser's expense.

PR #657, opened 2018-10-02
Adding analytics setup with hash tracking

jordajm:
### Checklist:

### Description:

- Adds Google Analytics script  ~~(Only runs on dapp.originprotocol.com)~~ (Runs on all domains except localhost and 0.0.0.0)
- Implements hash tracking
- Adds GA_ANALYTICS_ID default to webpack.config (but also must be hard-coded in dev.html b/c that file doesn't run through webpack compiler)
- We're now set up to add events as needed in another PR
- Sets `user_id` to wallet address when available so we can track users across devices

Issue: https://github.com/OriginProtocol/origin-dapp/issues/642


micahalcorn:
@joshfraser did you want the analytics to run on all domains and forks of the DApp and then filter after the fact?

src/components/analytics.js
```diff
@@ -0,0 +1,54 @@
...
+import { withRouter } from 'react-router'
+
+class Analytics extends Component {
+
+  componentDidUpdate(props) {
+    this.trackHashChange(props)
+  }
+
+  trackHashChange({ location, history }) {
+    const gaTrackingId = process.env.GA_TRACKING_ID || 'UA-106384880-2'
```
franckc: Is this default value UA-106384880-2 needed ? is there a risk it could cause data from local testing and staging to get tracked/pollute the mainnet data set ?

jordajm:
Working now to add some code that sets the user_id for tracking across devices but wanted to go ahead and get this submitted for review... ⏳

webpack.config.js
```diff
@@ -35,7 +35,8 @@ const env = {
   MAINNET_DAPP_BASEURL: null,
   RINKEBY_DAPP_BASEURL: null,
   PROVIDER_URL: '',
-  REDUX_LOGGER: false
+  REDUX_LOGGER: false,
+  GA_TRACKING_ID: '' // must also be hard-coded in dev.html - this is used in components/analytics.js
```
franckc: FYI, here is an example of how to set those env variables to different values in staging vs prod (needs separate origin-box PR):
See for example https://github.com/OriginProtocol/origin-box/pull/102

joshfraser:
Let's run it on all domains by default & we'll do the filtering in GA.

joshfraser:
@franckc we can setup filtering by domain in GA.

jordajm:
Updated @micahalcorn

src/components/analytics.js
jordajm: @franckc unfortunately gtags.js requires the ID to be passed into the `'config'` call. However, the `gtag` global var is undefined on localhost because the script is prevented from running on `dev.html`, so it won't track local pageviews/events.

webpack.config.js
jordajm: According to @joshfraser we're using the same ID across all properties - thanks for the link tho - might be useful in future iterations

PR #664, opened 2018-10-02
Listing Detail, Reviews, and Refactor

micahalcorn:
This creates a `Reviews` component to use everywhere that we want reviews for a particular user. It also removes truncation of the listing title everywhere except the grid card.


micahalcorn:
Thanks @DanielVF 🙂 I think that "Seller Reviews" begs the question of whether or not we are displaying reviews of the user when she was acting as a buyer. I'll go ahead and merge this with the assumption that there will be other copy edits in the coming hours.

PR #676, opened 2018-10-03
Proof of concept for modals to use new transaction polling

sparrowDom:
### Checklist:

### Description:
This is a proof of concept on how to make modal dialog on Metamask 4.12.0 dissapear with only receiving confirmation event and no transaction recite. 

**Depends on PR:** https://github.com/OriginProtocol/origin-js/pull/599


sparrowDom:
I wouldn't really merge this, would just like to open a discussion, with this PR example that shows what additional DAPP changes are required to make modal disappear and show the transaction in transaction dropdown. 

The biggest downside here is that the listingId is not tied to the transaction component in the transaction dropdown.

sparrowDom:
This is no longer neeeded

PR #694, opened 2018-10-08
New image cropper component implemented on create listing workflow

jordajm:
### Checklist:

### Description:

- Adds an image cropper react component that can be reused across the dapp
- Resizes images to a max width/height of 1000px to prevent images from exceeding IPFS size limit
- If @micahalcorn and @franckc and others like this solution, I can implement it in the other places that we upload images
- The way images are displayed on the create listing "review" step will need to be tweaked slightly if this PR is merged, but wanted to pause and get feedback first.
- If we implement this on the profile editor, we can remove the `browser-image-resizer` NPM dependency
- If we decide on an aspect ratio we want to use, that can be added/enforced easily
- We could also enforce a minimum width/height for images
- I see this as a starting point for solving our image upload problems, not a "fully polished" solution, but I think it's a good step in the right direction.

-----------

screenshot 2018-10-08 00 40 56


franckc:
I like it ! Perhaps for now we decide on a specific aspect ratio until the DApp is able to support more of a "gallery view" mode that can support any (and a mix of) aspect ratios for listings ? @micahalcorn any thoughts ?

jordajm:
If we do go with a fixed aspect ratio, I think I'd propose 5:4. I experimented with a few options and that seemed to fit most use cases, but it still is fairly limiting. Here are some examples/options:

aspect-ratio-comparison

micahalcorn:
We're currently hiding similar functionality that may be offered by [`react-avatar-editor`](https://github.com/OriginProtocol/origin-dapp/blob/master/src/pages/profile/EditProfile.js#L3). Would this PR potentially replace that package and update the profile code?

I probably won't have time to dig into this today.

jordajm:
@micahalcorn yes, my new cropper component combines the functionality of `react-avatar-editor` (but a more general use case I think) and `browser-image-resizer`, so we could get rid of both of those dependencies.

micahalcorn:
In today's product meeting, it was determined that 4:3 should be the prescribed aspect ratio. We also need to respect this ratio everywhere that we are rendering listing images. Currently, the dimensions are typically determined by the width of the screen (via grid, flexbox, etc) and the relative height is not enforced. I will make a separate issue for maintaining the ratio, but do you want to wrap this PR up to some degree before the big M&M, @jordajm?

jordajm:
Closing this non-monorepo PR and opening a new one in the monorepo

PR #709, opened 2018-10-09
Featured Badge fixes

sparrowDom:
### Checklist:

- ~[ ] Wrap any new text/strings for translation~
- ~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs)~

### Description:
- show featured listings only if a listing is featured an no other badges are shown
- show featured badge on search results


sparrowDom:
The fix for showing featured badge on search results is not obvious from the code diff because of the renames. The important non changed line (but logically changed because of the renames) is src/components/listings-grid.js:102

src/components/listing-card.js
```diff
@@ -89,7 +89,9 @@ class ListingCard extends Component {
       o => offerStatusToListingAvailability(o.status) === 'sold'
     )
     const isWithdrawn = status === 'inactive'
-    const showFeaturedBadge = this.props.featured && !isSold && !isPending
+    const showPendingBadge = isPending && !isWithdrawn
+    const showSoldBadge = isSold || isWithdrawn
```
franckc: Do we actually want to show the Sold badge on withdrawn items ?
Would it rather make sense to hide those listings ? Though if it complicates the logic not worth it - we can implement that in the back-end when we transition to fetching listing IDs from back-end next sprint.

franckc: Yes, makes sense. Not worth adding complexity.

franckc:
@micahalcorn ok with this fix ?

micahalcorn:
@franckc yes, I've added some changes to handle the badges on the listing detail as well. ✔️

PR #716, opened 2018-10-10
Featured Pagination Repair

micahalcorn:
This repairs the first page of the listings component to only show 12 results with featured listings included. I have not tested with search functionality, but it should not be affected. 🤞

### DIFF
--- a/src/components/listings-grid.js
+++ b/src/components/listings-grid.js
@@ -44,15 +44,15 @@ class ListingsGrid extends Component {
 
       activePage = parseInt(this.props.match.params.activePage) || 1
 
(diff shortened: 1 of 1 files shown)


sparrowDom:
@micahalcorn did another change. The previous state of the branch left some of the listings from page 1 to page 2 out. See how start/stop slicing indexes looked like: 
```
page 1: 0, 8
page 2: 12, 24
page 3: 24, 36
```

did a small change so that slicing indexes behave like this: 
```
page 1: 0, 8
page 2: 8, 20
page 3: 20, 32
```

(there are both examples when there are 4 featured listings configured)

PR #721, opened 2018-10-11
Another pagination fix

sparrowDom:
### Checklist:

### Description:
Correct the total number of listings to include featured listings

### DIFF
--- a/src/components/listings-grid.js
+++ b/src/components/listings-grid.js
@@ -29,24 +29,17 @@ class ListingsGrid extends Component {
   }
 
   render() {
-    const { contractFound, listingIds, search, featuredListings, hidden } = this.props
+    const { contractFound, listingIds, search, featuredListingIds, hiddenListingIds } = this.props
+    let activePage, currentPageListingIds, includedFeaturedListingIds, listingIdsNotHidden
 
-    // const pinnedListingIds = [0, 1, 2, 3, 4]
-    // const arrangedListingIds = [...pinnedListingIds, ...listingIds.filter(id => !pinnedListingIds.includes(id))]
-
-    let allListingsLength, activePage, showListingsIds
-    let shownFeaturedListings = []
     if (this.props.renderMode === 'home-page') {
-      const visibleListingsIds = listingIds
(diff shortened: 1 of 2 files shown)


sparrowDom:
Another pagination fix 🤦‍♂️

micahalcorn:
@sparrowDom I removed that because it was double-counting the featured listings. 🙃 More importantly, pagination is completely broken right now otherwise. I'm going to dig in and commit to this branch.

sparrowDom:
@micahalcorn it could be that double count was the problem in a previous revision, but right now it shows the number of listings subtracted by the number of featured listings. Because those are filtered out a few lines above...

Just curious, what is the case where pagination is broken?

micahalcorn:
I think you're correct.

A lot of listings are just missing entirely. After the first page, only a couple are shown. 😳

sparrowDom:
Interestingly pagination works fine for me from `micah/pagination` branch but not from `master`

micahalcorn:
I don't have that branch anymore. What environment/network are you testing with?

sparrowDom:
Aha I see it was a bad merge. This line: https://github.com/OriginProtocol/origin-dapp/blob/master/src/components/listings-grid.js#L61
```
Math.max(0, startSlicePosition + LISTINGS_PER_PAGE - featuredListings.length)
```

should be
```
Math.max(0, startSlicePosition + LISTINGS_PER_PAGE - shownFeaturedListings.length)
```

sparrowDom:
Have to go AFK right now. I can handle this in the evening if you want

sparrowDom:
I still had the branch locally, and was testing it on Rinkeby

micahalcorn:
@sparrowDom thanks for catching the bad merge. I've got an opinionated fix that I'll commit here for collective review.

micahalcorn:
@sparrowDom I've committed my changes and tested this with Mainnet data, the published Origin.js, and the production discovery server.

sparrowDom:
@micahalcorn nice that you have handled the case with sorting the featured listings at the top of each search page.

PR #728, opened 2018-10-12
Git Cleanup

micahalcorn:
This is an attempt to consolidate the git-related files.


.gitignore
```diff
@@ -1,8 +1,176 @@
...
+##########
+.env
+.idea
+.sass-cache
+build/
+build_webpack/
+coverage
+coverage.json
+npm-debug.log*
+node_modules/
```
franckc: There shouldn't be any node_modules getting created in the root dir of the repo ?

.gitignore
```diff
@@ -1,8 +1,176 @@
-*/node_modules/
-/development/.ipfs
+# Common #
+##########
+.env
+.idea
+.sass-cache
+build/
```
franckc: Will we have any webpack files in the root ? wouldn't they be under origin-dapp/, origin-js/ etc... ?

.gitignore
```diff
@@ -1,8 +1,176 @@
...
+/origin-bridge/.lock-wscript
+
+# Dependency directories
+/origin-bridge/jspm_packages/
+
+# Typescript v1 declaration files
+/origin-bridge/typings/
+
+# Optional npm cache directory
+/origin-bridge/.npm
```
franckc: Why would npm create a cache under bridge which is python code ?

.gitignore
```diff
@@ -1,8 +1,176 @@
...
+/origin-bridge/*.tgz
+
+# Yarn Integrity file
+/origin-bridge/.yarn-integrity
+
+# Python
+/origin-bridge/*.pyc
+/origin-bridge/ve
+
+# celery
```
franckc: Bridge no longer use celerey. We can remove.

.gitignore
```diff
@@ -1,8 +1,176 @@
...
+/origin-bridge/celerybeat-schedule.db
+
+# pytest
+/origin-bridge/.pytest_cache
+/origin-bridge/app/test_project.db
+
+# Flask Session files
+/origin-bridge/*flask_session/
+
+# Development contract directory
```
franckc: Bridge no longer deals with contracts. We can remove.

.gitignore
```diff
@@ -1,8 +1,176 @@
...
+/origin-docs/.cache
+/origin-docs/.vagrant
+/origin-docs/.yardoc
+/origin-docs/_yardoc
+/origin-docs/doc/
+
+# origin-js #
+#############
+
+/origin-js/contracts/secret/
```
franckc: no idea what this is ?

micahalcorn:
Yo, I basically just copied all of these to the top level and consolidated the ones that were obviously duplicate. I don't know what most of them are. And I was under the impression that anything without a leading slash would match anything nested or not (such as node_modules, for example). Feel free to commit here as you see fit.

PR #731, opened 2018-10-12
Image cropper and resizer with drag-and-drop reordering

jordajm:
### Checklist:

### Description:

- Adds an image cropper react component that can be reused across the dapp
- Resizes images to a max width/height of 1000px to prevent images from exceeding IPFS size limit
- If @micahalcorn and @franckc and others like this solution, I can implement it in the other places that we upload images
- The way images are displayed on the create listing "review" step will need to be tweaked slightly if this PR is merged, but wanted to pause and get feedback first.
- If we implement this on the profile editor, we can remove the `browser-image-resizer` NPM dependency
- If we decide on an aspect ratio we want to use, that can be added/enforced easily
- We could also enforce a minimum width/height for images
- I see this as a starting point for solving our image upload problems, not a "fully polished" solution, but I think it's a good step in the right direction.

-----------

screenshot 2018-10-08 00 40 56


jordajm:
@micahalcorn this now forces 4:3 aspect ratio and is ready for a test spin / review / merge. 

We'll need follow-up PRs to change the way images are displayed and maybe to implement this cropper in the profile editor so we can get rid of the two NPM dependencies that it uses. 

Also could be useful to implement in the messaging image uploader to help with the 2MB limit (this downsizes the images as well)

micahalcorn:
Nicely done, @jordajm 👏 A couple of comments:

1. It seems that if I re-click the file selector and eventually choose the same file, the modal will never reappear. That's a bit of an edge case, but it might be a simple fix.
2. The editor breaks animated `.gif`s. @DanielVF won't mind, but it pains me to embrace this reality.

Do you prefer to merge this before respecting the aspect ratio on image rendering and/or adding the same functionality to the profile/messaging/etc? For what it's worth, I think we want to update the listing detail to [this design](https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bbc0003326ac118dbbc7ef3) at the same time.

micahalcorn:
Great work 🏆 A couple of requests:

1. Let's hide the row of thumbnail~s~ when there is only one photo.
2. The aspect ratio is not maintained on the listing card (shown on browse and search results). It's also a bit less-than-ideal on the listings/purchases/sales cards, but we can let them slide for now.

I see that we're now preventing multiple-selection of files during listing creation. Is there an established UX for using this image cropper with multiple images? We don't need to handle that right now, but it might be a nice-to-have.

What other related changes would you like to take on after merging this PR?

- profile avatar
- messaging image
- ?

jordajm:
@micahalcorn - I pushed a commit to hide thumbnail row if there's only one image - good call on that. 

I played with the listing cards some more and didn't find a way to make old listing images (of various aspect ratios) and new (cropped & standardized) listing images all look good with a single setup, but I think the current setup makes them all look as good as possible at all screen sizes. LMK if you have other ideas for how to improve it.

Re: multi-image uploads with a cropper, I'm not sure what UX makes sense, but here's Aure's take on it:
https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5a361b9ca67910e89ccf1b5f

That would require a whole step in listing-create just for images, so maybe we wait on that for now? 

And yes, I agree with your follow-up to-do list and I don't know of anything besides implementing this cropper on profile avatar and messaging and removing the dependencies that those are relying on currently.

micahalcorn:
I'll think on it and make some issues, thanks. ✅

wanderingstan:
Not sure where this convo thread is continued, but I would vote that we follow airbnb's example of **not** showing the cropping tool upon upload. Just automatically scale the image and stick it in, and leave the cropping to an optional later step.

lucadgtrip:
some feedback after uploading a few listings:
while it's nice to have a cropper the situation at the moment is making an upload much slower.
For a listing with 10 images **it adds several clicks** ( probably 30-40 extra) that it makes it much longer and harder.
It also **increases mistakes** as I have to manually select image 1, then 2, then 3 etc..and forget some images in the process.
Also, as a user I don't have time to manually crop each image and simply accept the default crop.
Suggestion:
- as @wanderingstan says: Just automatically scale the image and stick it in, and leave the cropping to an optional later step.
- allow multiple uploads

at the moment this feature feels like a "nice to have" making the site more cumbersome.
As a result less listings will be uploaded.
It's a great feature but it has to be placed as an option in my opinion.

micahalcorn:
Great feedback @lucadgtrip, thanks 🙏

PR #732, opened 2018-10-12
Fix box install

franckc:
We no longer need to clone the repos.

I'll update the README in a separate PR to mention that to get a stable version of origin-box, it is best to run install.sh from the stable branch of the origin repo.


franckc:
Domen is fixing in #733 - closing this PR.

PR #759, opened 2018-10-16
Sparrow dom/attestation in ipfs

sparrowDom:
### Checklist:

### Description:
Store Airbnb and Twitter external ids to IPFS

### DIFF
--- a/origin-bridge/logic/attestation_service.py
+++ b/origin-bridge/logic/attestation_service.py
@@ -33,6 +33,7 @@ signing_key = settings.ATTESTATION_SIGNING_KEY
 twitter_request_token_url = 'https://api.twitter.com/oauth/request_token'
 twitter_authenticate_url = 'https://api.twitter.com/oauth/authenticate'
 twitter_access_token_url = 'https://api.twitter.com/oauth/access_token'
+twitter_show_user_url = 'https://api.twitter.com/1.1/users/show.json'
 
 CLAIM_TYPES = {
     'phone': 10,
@@ -357,8 +358,7 @@ class VerificationService:
             response.raise_for_status()
         except requests.exceptions.HTTPError as exc:
             logger.exception(exc)
-            #raise TwitterVerificationError('Invalid response from Twitter.')
-            raise TwitterVerificationError(str(exc))
+            raise TwitterVerificationError('Invalid response from Twitter.')
 
         as_bytes = urllib.parse.parse_qs(response.content)
         token_bytes = as_bytes[b'oauth_token'][0]
@@ -423,7 +423,7 @@ class VerificationService:
             'signature': signature,
             'claim_type': CLAIM_TYPES['twitter'],
             'data': data,
-            'external_id': airbnbUserId
+            'external_id': screen_name
         })
 
     def generate_airbnb_verification_code(eth_address, airbnbUserId):
@@ -481,7 +481,8 @@ class VerificationService:
         return VerificationServiceResponse({
             'signature': signature,
             'claim_type': CLAIM_TYPES['airbnb'],
-            'data': data
+            'data': data,
+            'external_id': airbnbUserId
         })
 
 
--- a/origin-bridge/logic/attestation_service.py
+++ b/origin-bridge/logic/attestation_service.py
@@ -357,7 +357,8 @@ class VerificationService:
             response.raise_for_status()
         except requests.exceptions.HTTPError as exc:
             logger.exception(exc)
-            raise TwitterVerificationError('Invalid response from Twitter.')
+            #raise TwitterVerificationError('Invalid response from Twitter.')
+            raise TwitterVerificationError(str(exc))
 
         as_bytes = urllib.parse.parse_qs(response.content)
         token_bytes = as_bytes[b'oauth_token'][0]
@@ -421,7 +422,8 @@ class VerificationService:
         return VerificationServiceResponse({
             'signature': signature,
             'claim_type': CLAIM_TYPES['twitter'],
-            'data': data
+            'data': data,
+            'external_id': airbnbUserId
         })
 
     def generate_airbnb_verification_code(eth_address, airbnbUserId):
--- a/origin-js/src/resources/attestations.js
+++ b/origin-js/src/resources/attestations.js
@@ -22,7 +22,8 @@ class Attestations {
       return new AttestationObject({
         topic: resp['claim-type'],
         data: Web3.utils.soliditySha3(resp['data']),
-        signature: resp['signature']
+        signature: resp['signature'],
+        externalId: resp['external_id']
       })
     }
   }
--- a/TODO.txt
+++ b/TODO.txt
@@ -1,0 +1,207 @@
+TODO:
+- once you get volumes to sync sometimes the JS is still not recompiling
+
+Current task:
+- Airbnb is not really publicly written on the blockChain might need to correct the graphics, and explanation text
+
+Issues/PRs open:
+
+Fractional Usage Stuff to do: 
+- currently we have simple Craigs list steps. Fractional usage has different steps, because you are not buying stuff but renting stuff. Or using a service. See how may workflows we need to cover with this. No arbitration logic untill now. Start some kind of product document as a result of this. Ask Tyler or Stan. The result is the UI with the steps that needed to be taken outlined.  
+
+
+curl -X GET "localhost:9200/listings/_search?pretty" -H 'Content-Type: application/json' -d'
+{
+  "query": {
+    "bool": {
+      "must": [
+        { "match": {"all_text": "garage"} }
+      ],
+      "filter": [
+        {
+          "range": {
+            "price.amount": {
+              "lte": 10
+            }
+          }
+        },
+        {
+          "range": {
+            "price.amount": {
+              "gte": 2
+            }
+          }
+        },
+        {
+          "bool": {
+            "should": [
+              { "match": {"category": "Housing"} },
+              { "match": {"category": "Housing111"} }
+            ]
+          }
+        },
+        {
+          "bool": {
+            "should": [
+              { "match": {"subCategory": "Housing"} },
+              { "match": {"subCategory": "Vacation Rentals"} }
+            ]
+          }
+        }
+      ]
+    }
+  },
+  "_source": ["title", "description", "price"]
+}
+'
+
+curl -X POST "localhost:9200/listings/listing/999-001-17/_explain?pretty" -H 'Content-Type: application/json' -d'
+{
+  "query": {
+    "function_score": {
+      "query": {
+        "bool": {
+          "must": [
+            { "match": {"all_text": "wish"} }
+          ]
+        }
+      },
+      "field_value_factor": { 
+        "field": "commission.amount",
+        "factor": 0.005
+      },
+      "boost_mode": "sum"
+    }
+  }
+}
+'
+
+curl -X POST "localhost:9200/listings/_search?pretty" -H 'Content-Type: application/json' -d'
+{
+  "query": {
+    "function_score": {
+        "boost_mode": "sum",
+        "field_value_factor": {
+            "factor": 0.005,
+            "field": "commission.amount"
+        },
+        "query": {
+            "bool": {
+                "filter": [],
+                "must": [
+                  { "match": {
+                      "all_text": {
+                        "query": "garages",
+                        "fuzziness": "AUTO"
+                      }
+                    }
+                  },
+                  {
+                    "match": {
+                      "status": "active"
+                    }
+                  }
+                ],
+                "must_not": [
+                {
+                  "ids" : {
+                    "values" : ["999-000-1"]
+                  }
+                }
+                ],
+                "should": [
+                {
+                  "ids" : {
+                    "values" : ["999-000-0"],
+                    "boost": 10
+                  }
+                }]
+            }
+        }
+    }
+  }
+}
+'
+
+// get all listings with missing comission
+curl -X POST "localhost:9200/listings/_search?pretty" -H 'Content-Type: application/json' -d'
+{
+  "query": {
+    "bool": {
+      "must_not": {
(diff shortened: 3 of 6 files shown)


sparrowDom:
@micahalcorn I guess it is best I leave PR to you since you will be doing the changes in the DApp. As this PR stands currently the external identifiable information is stored in the `data` field of the claim encoded in hex. There is a `decodeHex` function added for the DApp to decode data ( see `resource_attestations.test.js` example how to retrieve attestation data)

Good thing to keep in mind is that existing Twitter and Airbnb attestations will probably throw an error if you try to hex decode/JSON parse their data fields.

micahalcorn:
@sparrowDom I think those implementation details should be origin-js concerns. Do you mind taking a stab at handling that logic in the user adapter (or wherever appropriate) so that the DApp can just read the results like it does for the profile?

sparrowDom:
Good idea, will do

sparrowDom:
@micahalcorn @franckc let me make the necessary changes so that external ids are stored to the IPFS. Will ping you once I am done with this.

tyleryasaka:
Definitely *should not* put any PII directly on-chain. Should at the very least be an IPFS hash.

sparrowDom:
@franckc @micahalcorn the external ids are now stored to IPFS. `Origin-js` hands a claim with extra `ipfsHash` property when that claim is either twitter / Airbnb. 

Btw I don't know why 1 of Travis builds is failing. It fails in a test that has already been corrected.

origin-bridge/util/ipfs.py
```diff
@@ -51,6 +51,9 @@ def directly_pinned_hashes(self, pin_types=None):
         pinned_keys = self.connector.pin_ls()['Keys']
         return [k[0] for k in pinned_keys.items() if k[1]['Type'] in pin_types]
 
+    def pin_json(self, json):
```
franckc: This method does not pin. It adds json data to IPFS. So let's just call it "add_json" to avoid any confusion.

origin-bridge/logic/attestation_service.py
```diff
@@ -398,16 +399,15 @@ def verify_twitter(oauth_verifier, eth_address):
...
-        data = 'twitter verified'
-        # TODO: determine claim type integer code for phone verification
-        signature = attestations.generate_signature(
-            signing_key, eth_address, CLAIM_TYPES['twitter'], data
-        )
-
         query_string = urllib.parse.parse_qs(response.content)
         screen_name = query_string[b'screen_name'][0].decode('utf-8')
 
+        ipfs_hash = ipfs_helper.pin_json({'screen_name': screen_name})
```
franckc: To be consistent with all the other data we are storing in IPFS that uses JSON schema and also to allow future validation and versioning, let's store a schemaID. I'd suggest something like this:
```
{
  "schemaId": "http://schema.originprotocol.com/twitter-attestation_v1.0.0",
  "screen_name": screen_name
}
```

origin-bridge/util/attestations.py
```diff
@@ -7,12 +7,13 @@
 # http://web3py.readthedocs.io/en/stable/web3.eth.account.html?highlight=private#not-acceptable-for-production
 w3.eth.enable_unaudited_features()
 
+DONT_HASH_DATA_TOPICS = [4, 5]
```
franckc: Why do we need to special case these ? It complicates the code...
What would be wrong with hashing the IPFS hash we get for Twitter/Airbnb ?

origin-js/src/contractInterface/users/v00_adapter.js
```diff
@@ -183,15 +185,31 @@ class V00_UsersAdapter {
     if (profileClaims.length) {
       const bytes32 = profileClaims[profileClaims.length - 1].data
       const ipfsHash = this.contractService.getIpfsHashFromBytes32(bytes32)
-      profile = await this.ipfsDataStore.load(PROFILE_DATA_TYPE, ipfsHash)
+      try {
+        profile = await this.ipfsDataStore.load(PROFILE_DATA_TYPE, ipfsHash)
+      } catch (error) {
+        console.error(`Can not read profile data from ipfs: ${error.message}`)
```
franckc: Let's also add the ipfsHash to this error message to make it easier to troubleshoot.

origin-js/src/contractInterface/users/v00_adapter.js
```diff
@@ -116,7 +117,7 @@ class V00_UsersAdapter {
             return data.substr(2)
           })
           .join('')
-      const dataOffsets = attestations.map(() => 32) // all data hashes will be 32 bytes
+      const dataOffsets = attestations.map(({ data }) => (data.length - 2) / 2) // deduct 2 because '0x' is removed
```
franckc: nit: it is not completely obvious so maybe add a comment to explain the calculation of data offset. Something along the lines that binary data for each attestation is passed as string using hexadecimal encoding. For ex: "0x12AB03". To get the binary data length, we subtract the "0x" prefix and divide the reminder string length by 2.

franckc:
Note: I fixed the Travis CI configuration so now bridge tests will run as part of the build.

origin-bridge/logic/attestation_service.py
sparrowDom: good point 👍

origin-bridge/util/attestations.py
sparrowDom: If we hash the data field then it becomes unreadable to the DApp, and it won't be able to fetch the IPFS hash from that field. There is some additional complexity here and I will add explanation as a comment:
```
- IPFS hash is a base58 encoded string
- We store IPFS hashes in solidity claims in byte32 hex encoding to minimise gas cost. 
- byte32 is not string serialisable so it can not be transmitted in that form from discovery to the DApp
- discovery needs to transform ipfs hash to base32 hex encoding (that is how it is going to be stored in the contract) before signing the claim, and then send IPFS hash to the DApp in base58 encoding.
- this way claim has a correct signature if IPFS hash has byte32 hex encoding
- the DApp takes signature and other claim info and transforms the base58 encoded IPFS hash to base32 hex before submitting the claim to web3. 
```

origin-bridge/util/ipfs.py
sparrowDom: From what I understand any content that is added to the `local` IPFS node is automatically pinned and not garbage collected. Content from other nodes can be garbage collected on the `local` node unless it is pinned explicitly.

origin-js/src/contractInterface/users/v00_adapter.js
sparrowDom: 👍

origin-js/src/contractInterface/users/v00_adapter.js
sparrowDom: Good point, added it.

origin-bridge/logic/attestation_service.py
```diff
@@ -460,10 +464,25 @@ def verify_airbnb(eth_address, airbnbUserId):
...
+            'schemaId': 'http://schema.originprotocol.com/airbnb-attestation_v1.0.0',
+            'airbnb_user_id': airbnbUserId
+        })
+
+        """ - IPFS hash is a base58 encoded string
+            - We store IPFS hashes in solidity claims in byte32 hex encoding to minimise
+              gas cost.
+            - byte32 is not string serialisable so it can not be transmitted in that form
+              from discovery to the DApp
+            - discovery needs to transform ipfs hash to base32 hex encoding (that is how
```
franckc: nits:
 - it is not discovery, it is the bridge server.
 - there is no such thing as base32 encoding. it is just encoded in hexadecimal

origin-bridge/util/attestations.py
franckc: Yes, thanks for the details, it makes sense.

origin-bridge/util/ipfs.py
franckc: Yes, it is correct but I still think it is confusing to call this method "pin_json" since a) it is a thin wrapper around the "add_json" method and b) pinning typically refers to pinning a hash :)

origin-js/src/contractInterface/users/v00_adapter.js
```diff
@@ -116,7 +117,12 @@ class V00_UsersAdapter {
             return data.substr(2)
           })
           .join('')
-      const dataOffsets = attestations.map(() => 32) // all data hashes will be 32 bytes
+
+      /* All the data is currenlty base32 encoded. To calculate the lenght we just deduct 2 (becase '0x') gets
```
franckc: nit: same as previous comment. The data binary and since it is hex encoded, it is twice as long as in its binary encoded form.

origin-js/src/contractInterface/users/v00_adapter.js
```diff
@@ -183,15 +190,31 @@ class V00_UsersAdapter {
...
+
+    const attestations = validAttestations
+      .map(att => {
+        try {
+          if (ClaimDataIsIpfsHash.includes(att.topic))
+            att.ipfsHash = this.contractService.getIpfsHashFromBytes32(att.data)
+          return att
+        }
+        catch (error) {
+          console.error(`Can not convert to ipfs hash: ${error.message}`)
```
franckc: Would be nice if we would keep backward compatibility and still support legacy Airbnb/Twitter attestations that were not storing data in IPFS ?

origin-js/src/resources/attestations.js
```diff
@@ -11,6 +11,20 @@ const responseToUrl = (resp = {}) => {
   return resp['url']
 }
 
+/* - IPFS hash is a base58 encoded string
```
franckc: Same comments as above.

origin-bridge/util/ipfs.py
tomlinton: Yep if we stick to the [IPFS API](https://github.com/ipfs/js-ipfs-api) it should reduce confusion. An add operation is different to a pin operation. You are totally right that an add operation on our node will result in it being pinned though.

origin-bridge/logic/attestation_service.py
```diff
@@ -460,10 +464,25 @@ def verify_airbnb(eth_address, airbnbUserId):
...
 
-        # TODO: determine the schema for claim data
-        data = 'airbnbUserId:' + airbnbUserId
+        ipfs_hash = ipfs_helper.pin_json({
+            'schemaId': 'http://schema.originprotocol.com/airbnb-attestation_v1.0.0',
+            'airbnb_user_id': airbnbUserId
+        })
+
+        """ - IPFS hash is a base58 encoded string
+            - We store IPFS hashes in solidity claims in byte32 hex encoding to minimise
```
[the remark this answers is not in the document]
sparrowDom: Yup thanks for that. Changing the line to: 
```
- We store IPFS hashes in solidity claims in bytes32 format to minimise
   gas cost.
```

origin-bridge/logic/attestation_service.py
sparrowDom: 👍 
Changing the line to: 
```
- bridge needs to transform ipfs hash to bytes32 format (that is how
     it is going to be stored in the contract) before signing the claim, and then
     send IPFS hash to the DApp in base58 string encoding.
```

origin-bridge/util/ipfs.py
sparrowDom: Good point. Renaming to `app_json` to increase clarity.

origin-js/src/contractInterface/users/v00_adapter.js
sparrowDom: 👍 changing the comment to: 
```
All the data is in bytes32 binary format in the contract. To calculate the lenght we just
deduct 2 (becase '0x') gets removed. And then divide the remaining lenght by 2 becase
the data is twice the size when it is represented as string (with hex encoding).
```

origin-js/src/resources/attestations.js
sparrowDom: 👍

origin-js/src/contractInterface/users/v00_adapter.js
sparrowDom: good point. Have changed it that it is always a valid attestation. 

There is one downside though, when testing with old attestations, this function `this.contractService.getIpfsHashFromBytes32` will most of the time succeed and come up with something that looks like an IPFS hash but it is not. We have previously populated the data field on the bridge server and hashed it in the DApp before storing it. Twitter attestation used to have non changing data (string "twitter verified" ) but Airbnb had an id in the data field (https://github.com/OriginProtocol/origin/blob/master/origin-bridge/logic/attestation_service.py#L464). Meaning after hashing it in the DApp we can not recognise legacy data attestation with the current data attestations (for Twitter we can). 

With the current implementation the DApp needs to expect that not all ipfs fileds in attestations are valid. Some are legacy that do not resove to a file in IPFS (even though they will appear to have an IPFS hash location) @micahalcorn 

Can anyone think of a better solution for this? Should we make a special case to recognise legacy Twitter attestations?

franckc: It is hacky but could we look at the block number the attestation was issued at (I think the event should include the blockNumber) and have a cut-off hard coded constant that indicates if it is old style attestation vs new style ? e.g. any attestation with blockNumber > cutoff is new style

franckc:
Note: looks like there is a relevant origin-js test failure for test "should be able to deploy new identity with 6 presigned claims" - see https://travis-ci.org/OriginProtocol/origin/jobs/445102489

origin-js/src/contractInterface/users/v00_adapter.js
sparrowDom: That is a good idea. Have added a new property in `origin-dapp.env` file. The only caveat is when deploying for the first time on `Rinkeby` and `Mainnet` that variable needs to be set. The current block height can be fetched in browser in console: 
```
await window.web3.eth.getBlockNumber()
```

PR #762, opened 2018-10-16
[EventListener] Wait 1 block confirmation before processing events

franckc:

### Checklist:

### Description:
Wait 1 block confirmation before processing events. This should help with the out-of-sync node Infura issue we've seen (though currently it is mitigated by retries).


franckc:
Yep, it is a tradeoff between data consistency and indexing lag.

This adds a 15sec delay to indexer (since it takes avg 15sec for a block to get mined), but on the other hand also adds a little bit of extra protection against indexing data that eventually may not be part of the final blockchain. I think typically it is recommended to wait for at least 6 blocks confirmation...

Given that after issuing a transaction to add a listing we show the users a progress bar (and I think there we show 100% complete only after 12 blocks have been confirmed), I think it is ok to add some indexing delay.

tomlinton:
@franckc sounds good, thanks for the clarification!

PR #778, opened 2018-10-18
Make TravisCI exit on first test failure

tomlinton:
- Small change to TravisCI to make it exit on first test failure


tomlinton:
No longer needed due to #786.

PR #795, opened 2018-10-22
Push Notifications (Chrome & Firefox)

micahalcorn:
### Checklist:

- [ ] ~Format code to be consistent with the project~
- [ ] ~Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/origin-docs)~

### Description:

This is a proof-of-concept for implementing browser-based push notifications. Here is an overview of what is included and what may or may not need to be handled in the near/distant future:

**Notifications Server**
------
A new subdirectory (origin-notifications) is created. It contains an Express server with a PostgreSQL connection mapped using [Sequelize](http://docs.sequelizejs.com/). There is only one model, the `PushSubscription`, which includes the following attributes:

- `id`: default, automatically-incremented integer
- `createdAt`: default, automatically-generated timestamp
- `updatedAt`: default, automatically-generated timestamp
- `endpoint`: URL string generated when a subscription is created in the browser, determined by the push service provider (Google FCM or Mozilla), with a unique token appended to the path
- `keys`: a pair of values (`auth` and `p256dh`) generated when `toJson` is called on a subscription
- `expirationTime`: currently left blank
- `account`: the subscriber's then-current ETH address

Since the subscription object is unique to the client device, it may be relevant to more than one ETH account. For this reason, neither the `endpoint` nor the `account` values are separately unique in the `PushSubscription` table. However, no two records should contain both the same `endpoint` and the same `account`.

The only events currently resulting in notifications are those involving an offer. New messages and listings could be candidates for future notifications.

Three environment variables are necessary. An email address should be provided in case the push service provider needs to contact the sender of the push events. And a [VAPID key pair](https://datatracker.ietf.org/doc/rfc8292/) can be generated by the [web-push](https://github.com/web-push-libs/web-push) library.
```
VAPID_EMAIL_ADDRESS=
VAPID_PRIVATE_KEY=
VAPID_PUBLIC_KEY=
```

_Routes_
- GET `/`: returns all `PushSubscription` records
- POST `/`: creates a new `PushSubscription` if one does not exist with both a matching `endpoint` and a matching `account`
- POST `/events`: conditionally converts a blockchain event into a push notification, which is sent to any subscription endpoints generated by an ETH account matching the buyer or seller

_To Consider_
- appropriate messages to send
- hiding the current GET results at `/`
- storing and queueing events
- CORS rules
- safeguarding against duplicates
- localization of messages
- connecting to a production database, perhaps the same one that is used by origin-discovery and/or the upcoming origin-mobile
- indexing
- including API keys if required to send notifications at scale
- integration with Discord webhook or other webhook subscribers
- replacing integer IDs with UUIDs or something else
- formatting code to match origin-js

_To Test_
1. `createdb notifications`
1. `psql notifications -c 'CREATE EXTENSION hstore;'`
1. `npm install`
1. `npm install -g sequelize-cli`
1. `sequelize db:migrate`
1. `node app.js`
1. Run origin-discovery listener with `node listener/listener.js --continue-file=continue --webhook=http://localhost:3456/events`

**Service Worker**
------
The service worker is installed in the browser when the DApp is loaded. It allows for the creation of `PushSubscription`s and handles push events from the server and user interactions with notifications. The code runs in a separate thread outside of the DOM context and is usually executed when the DApp is not currently open in the browser. It is scoped to the origin, and thus, registrations and notifications will be specific to the dev/staging/prod environment as long as we continue to use separate subdomains.

_To Consider_
- analytics
- consolidating multiple active notifications
- refraining from rendering notifications when relevant URL is already the focused tab
- providing various actions for the user to take besides clicking to open or closing
- the consequences of updates to the JS file
- bundling the JS file with React vs deploying it in the `public` directory
- caching (unrelated to notifications)

_To Test_
1. Run origin-dapp

**DApp Changes**
------
The most critical UX precaution is to ask the user to enable notifications at the most compelling time _before_ actually initiating the browser's native permission request prompt. Once a user denies permission, she is unlikely to ever grant it since doing so would require changing the browser's settings. This first implementation prompts the user to enable notifications immediately after creating a listing or making an offer. If the modal is dismissed, a stronger prompt is rendered explaining the importance of push notifications in a decentralized application. Permission is only requested at the API level after the user affirms her interest in enabling notifications. A `PushSubscription` is created after this permission is granted. Additional `PushSubscription`s are created each time the user creates a listing or makes an offer using a different ETH account using a browser in which permission was already granted.

Two environment variables are required: the public VAPID key and the notification server's URL.
```
NOTIFICATIONS_KEY=
NOTIFICATIONS_URL=
```

_To Consider_
- not rendering the call-to-action modals on top of existing modals
- prompting for notifications in later offer lifecycle stages if not previously enabled
- analytics
- scoping subscriptions to ETH networks
- displaying non-push notifications
- blocking/obscuring the UI while the native permission dialog is displayed
- supporting notifications configuration in a settings page

_To Test_
1. Create a listing
1. Enable notifications
1. Make an offer from a different account


origin-notifications/app.js
```diff
@@ -0,0 +1,96 @@
...
+app.use((req, res, next) => {
+  res.header("Access-Control-Allow-Origin", "*")
+  res.header("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept")
+
+  next()
+})
+
+app.use(bodyParser.json())
+
+// currently showing all subscriptions in an unauthenticated index view - maybe not a good idea in production?
```
franckc: Ok for debugging on local environment but we should disable this in production.

origin-notifications/app.js
```diff
@@ -0,0 +1,96 @@
...
+app.get('/', async (req, res) => {
+  const subs = await PushSubscription.findAll()
+
+  res.send(`<h1>${subs.length} Push Subscriptions</h1><ul>${subs.map(s => `<li><pre style="white-space: pre-wrap;word-break: break-all">${JSON.stringify(s)}</pre></li>`)}</ul>`)
+})
+
+app.post('/', async(req, res) => {
```
franckc: We add authentication to this endpoint so that only the account owner can insert data.
Perhaps the message could be signed using the account pk and then the back-end verifies the signature ? Though that would mean a meta-mask popup which is always an annoying user experience... Another possibility would be to create authenticated sessions at the DApp layer so that the user does not have to sign messages every single time she takes an action that requires authentication with the back-end. That could be a good framework to have in place for future ? Let's discuss...

origin-notifications/app.js
```diff
@@ -0,0 +1,96 @@
...
+  if (existing) {
+    return res.sendStatus(200)
+  }
+
+  await PushSubscription.create(req.body)
+
+  res.sendStatus(201)
+})
+
+app.post('/events', async (req, res) => {
```
franckc: Need to add authentication otherwise user could get spammed by an external service calling this webhook.
We could have a secret that the listener passes as a header value in the request and that the notifier service checks. Another possibility would be to rely on internal queues rather than webhooks: listener would enqueue events and notifier would dequeue and process. Let's discuss...

origin-notifications/app.js
```diff
@@ -0,0 +1,96 @@
...
+  })
+
+  // filter out redundant endpoints before iterating
+  subs.filter((s, i, self) => {
+    return self.map(ms => ms.endpoint).indexOf(s.endpoint) === i
+  }).forEach(async s => {
+    try {
+      // should safeguard against duplicates
+      await webpush.sendNotification(s, JSON.stringify({
+        title: log.eventName,
```
franckc: We'll probably want to localize this string. Add a TODO

origin-notifications/app.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+  })
+
+  // filter out redundant endpoints before iterating
+  subs.filter((s, i, self) => {
+    return self.map(ms => ms.endpoint).indexOf(s.endpoint) === i
+  }).forEach(async s => {
+    try {
+      // should safeguard against duplicates
+      await webpush.sendNotification(s, JSON.stringify({
```
franckc: We'll probably want to have some retry in case for some reason that server is temporarily down.
Ok to just add a TODO for now.

Note: there is a retry function implemented in origin-js/src/utils/retries.js
We need this in quite a few places so perhaps we should just pull it out of origin-js and put it in a util package or equivalent ?

origin-notifications/migrations/20181019004743-create-push-subscription.js
```diff
@@ -0,0 +1,39 @@
...
+      createdAt: {
+        allowNull: false,
+        type: Sequelize.DATE
+      },
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE
+      }
+    })
+  },
```
franckc: I think there should be an index on (account, endpoint) since you have a query doing a lookup based on that. Otherwise it would cause the query to do a full table scan (which is terrible for performance if we start having lot of rows in this table).

Adding a line like this would do:
.then(() => queryInterface.addIndex('PushSubscriptions', ['account', 'endpoint']))

origin-notifications/src/config/config.json
```diff
@@ -0,0 +1,9 @@
+{
```
franckc: Add a TODO. For productionizing this we'll need to use env variables (DATA_USERNAME, DATABASE_PASSWORD, etc...)

origin-notifications/migrations/20181019004743-create-push-subscription.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('PushSubscriptions', {
+      id: {
+        allowNull: false,
```
franckc: Unecessary since this is a primary key and therefore it can't be null.

micahalcorn:
Based on my conversation with @franckc this morning, here are some additional considerations:

- authentication for write endpoints
- server-side preferences
- retries

Before merging this PR, I will add a compound index for the endpoints and accounts along with rate limiting for subscription creation. The Boss will add add a shared secret to this new server and the discovery server.

origin-notifications/migrations/20181019004743-create-push-subscription.js
franckc: nice ! :)

origin-notifications/app.js
```diff
@@ -50,8 +50,13 @@ app.use(bodyParser.json())
 // currently showing all subscriptions in an unauthenticated index view - maybe not a good idea in production?
 app.get('/', async (req, res) => {
   const subs = await PushSubscription.findAll()
```
franckc: I'd recommend moving the DB query within the (app.get('env') === 'development') statement.

origin-notifications/app.js
micahalcorn: Now rate limited, as we agreed that anyone should be able to subscribe to any events ✅

origin-notifications/app.js
tomlinton: This doesn't need to be exposed to the outside world right? The simple solution here is I don't expose this endpoint outside the Kubernetes cluster. The listener can just call it from inside the cluster.

micahalcorn: Correct, @tomlinton 👌

franckc: Excellent idea @tomlinton !

DanielVF:
Vapid keygen, for reference:

```javascript
const webpush = require('web-push');
const vapidKeys = webpush.generateVAPIDKeys()

vapidKeys.publicKey
#gives 'BDO0P...eoH'

vapidKeys.privateKey
#gives 'sjfiejfwief_wefwfwe'
```

PR #802, opened 2018-10-23
[EventListener] Use sequelize to store Listing and Offer data in DB

franckc:

### Checklist:

- [] Map any new environment variables with a default value in the Webpack config

### Description:
Changes
 - Save Listing and Offer data, as returned by the origin-js marketplace.getListing/Offer methods into the database
 - Use Sequelize as ORM (since we decided to standardize on that library)
 - Update origin box and deployment config

There are 2 use cases for the data:
1. Analytics - we want to create dashboards and analyze the data.
2. Replacement to fetching data from web3 which is slow and unreliable. We want to move towards a model where the DApp makes calls to the GraphQL server which will then fetch the data from the DB and return it to the DApp.

Notes:
 - We need to be clear on what is the ground truth for the back-end data. Is it Elastic or Postgres ? I'd like to propose that Elastic is only used for searching data to power the search page. Search would return listing IDs that then get looked up in Postgres. Any other data reads than search would go against Postgres (ex: Browse page, My Sales page, My Offers page, etc...).
 - The main reasons for this proposal: a) sequelize and postgres make it very straightforward to write queries against the data  - it is simpler and more concise than Elastic search. b) we need the data in a DB anyway for Analytics. Metabase (the dashboarding solution we use) does support ElasticSearch, but overall most of our analytics user are non engineers (Josh, Jon, Coleman, etc...) and will be more comfortable using a traditional SQL syntax. c) even for engineers, using the postgres cli makes it lot easier to look at the data compared to crafting ES queries.


origin-discovery/listener/listener.js
```diff
@@ -399,15 +424,44 @@ async function handleLog(log, rule, contractVersion, context) {
...
+        sellerAddress: listing.seller.toLowerCase(),
+        data: listing,
+      }
+      if (rule.eventName === 'ListingCreated') {
+        listingData.createdAt = log.date
+      } else {
+        listingData.updatedAt = log.date
+      }
+      await withRetrys(async () => {
+        await db.Listing.insertOrUpdate(listingData)
```
tomlinton: If I'm understanding correctly this represents a change from storing all the events for a listing to storing a single row per listing that gets updated with each new event. Are there any situations where we need the event history for a listing? Maybe if we had the complete event history we could recreate a listing at the time an offer was made? Might be something we need to think about later. I normally like an event sourcing style pattern for this sort of thing, but this may be cleaner and have better performance.

Not asking you to change anything, just thinking out loud. :grin:

franckc: In order to create a Listing object, we need more than just events. We need the data from the smart contract as well as the IPFS data. From a latency standpoint, it would be pretty slow to fetch all that data for let's say 100 listings we want to show on Browse/Search page. That's why I went for storing the latest state of the Listing/Offer.

But I do agree that having a quick way to lookup events could be very useful in future. For example the DApp currently queries web3 for fetching data to use in the Notificationt tab.
Today I actually spend some time populating a Postgres "Event" table via the event-listener that will give us the event history. I'll send a PR once this PR is merged in since I need the sequelize scaffolding first :)

sparrowDom:
@franckc I don't know about My Sales page & My Offers page, but it might make sense to read listings from elasticsearch for Browse page. One advantage of this is that the search will require implementation of hidden and featured listings in the search query. (I have that implemented in a branch that is a bit stale ATM). And it was pretty straightforward implementing that in elasticsearch. If we read Browse listings from postgres we need to reimplement that. 

There is also pagination and ordering that will be shared between search page and browse page.

franckc:
@sparrowDom Given that we anyway want to persist what listings are hidden/featured in the DB (to replace the current lists stored in github), it should not be too difficult to write a query for Browse page that uses that data. Then the question remains for search page in terms of handling hidden listings in a simple way so that it does not make pagination too complicated. Perhaps we can have a field in elastic search that indicates whether listing is hidden or not ?
Anyway, let's discuss and come up with a plan before implementing... :)

PR #807, opened 2018-10-24
[EventListener] Index contract events

franckc:

### Checklist:

### Description:
Store events emitted by marketplace contracts into a new Postgres "event" table.
Use cases for this data:
 - history of activity on Listings and Offers: useful for analytics but also when investigating engineering issues/bugs
 - power some DApp features via the back-end rather than web3 calls. For example to fetch notifications that are shown in the DApp, see marketplace.getNotifications method.


sparrowDom:
Nice, didn't know of `padStart` function. Pretty useful thing. LGTM, just have 1 thought:

would it make sense to also add `logIndex` into the Postgres table this way we could select events in this manner: `SELECT * FROM event ORDER BY blockNumber, logIndex` to replay events in chronological order.

franckc:
@sparrowDom The id field stores a concatenation of <blockNumber>-<logIndex> (with numbers padded to keep ordering). To get the logs in chronological order we can then just do:
`SELECT * FROM event ORDER BY id`

sparrowDom:
A right, thanks!

franckc:
Yeah, you guys are right. It is simpler to just use a composite pk (blockNumber, logIndex). Thanks for the suggestion. Let me update the schema...

PR #816, opened 2018-10-24
fix the command that starts origin-js process

sparrowDom:
### Checklist:

### Description:

Fix a command that starts `origin-js` process.


sparrowDom:
I have rebuilt all the containers and this solves the problem for me with `origin-js` quitting with error code 0.

PR #843, opened 2018-10-29
Removing '{' and '}'  from visible text

wanderingstan:
It read `{ETH}` Looks like maybe a leftover from trying to support multiple currencies?

image


micahalcorn:
@wanderingstan this was probably a case of passing in a link to what ETH is, but we don't need that.

PR #845, opened 2018-10-30
Automated deployment configuration

tomlinton:

### Checklist:

### Description:

- Adds Google Cloud Build config for build -> deploy -> commit tags
- Adds code for generating Discord webhooks in response to Cloud Build events (started, success, fail, timeout)
- Cleanup production vs dev dependencies for the DApp (anything that is required to build or run the DApp is a production dependency, everything else is dev)
- Optimised the containers for production builds and faster build times
- Created new secrets files that use the gcloud encrypt/decrypt tool rather than sops
- Added encrypted SSH key for GitHub so that the CI setup can push deployed tags
- Added missing Origin testnet to supported networks in DApp
- Couple of other minor cleanups

After this gets merged I just need to configure the triggers in GCC and we should be all go.

Future TODOs:

- Use `lerna bootstrap` to further speed up builds where common dependencies exist
- Use Docker's `--cache-from` flag to speed up container builds


deployment/cloudbuild-deploy.yaml
```diff
@@ -0,0 +1,140 @@
...
+    entrypoint: '/bin/bash'
+    args:
+      - '-c'
+      - |
+        docker push \
+          gcr.io/${PROJECT_ID}/$(cat .namespace)/${_CONTAINER}:${SHORT_SHA}"
+
+  # Add latest tag for to the container
+  - name: 'gcr.io/cloud-builders/gcloud'
+    entrypoint: 'sh'
```
franckc: nit: should this be /bin/bash so that all entrypoints are consistent re what shell they use ?
same questions for other entrypoints below.

deployment/dockerfiles/dev/event-listener
```diff
@@ -2,32 +2,25 @@
...
+COPY ./origin-discovery /app
 WORKDIR /app
 
-# Invalidate Docker cache if necessary
-COPY --from=origin-js-build /version.json /
-COPY --from=origin-js-build /app/origin-discovery/ /app/
-
-RUN npm install --quiet --no-progress
-
+ENV NODE_ENV production
```
franckc: nit: duplicate since already defined on line #8, so could be removed.
same comment for other dockerfiles that have same pattern.

deployment/utils/gcb-discord-notifications/index.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  // Send message to Discord.
+  const webhook = await createDiscordWebhook(build);
+  hook.send(webhook.message, webhook.options).then(callback);
+};
+
+// eventToBuild transforms pubsub event message to a build object.
+const eventToBuild = (data) => {
+  return data;
```
franckc: is data b64 encoded or not ? delete one of the two return statement :)

deployment/cloudbuild-deploy.yaml
```diff
@@ -0,0 +1,140 @@
...
+            echo "staging" > .namespace
+            ;;
+          stable)
+            echo "prod" > .namespace
+            ;;
+        esac
+
+        # Build container
+        docker build \
+          -f deployment/dockerfiles/$(cat .namespace)/${_CONTAINER} \
```
franckc: For my understanding, I assume the triggers will be in charge of determining what container(s) should be built depending on what files got changed in a PR ?

origin-js/package.json
```diff
@@ -42,13 +42,18 @@
     "crypto-random-string": "^1.0.0",
     "ethereumjs-util": "^5.2.0",
     "form-data": "^2.3.2",
+    "fs-extra": "^7.0.0",
```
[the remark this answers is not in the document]
tomlinton: Yea that is a tricky one. What I'm trying to do here is avoid installing all the dev dependencies because it greatly speeds up the builds. But I need to install the dependencies that are required for building production. I might be able to refactor to remove fs-extra, and then just add npm install for truffle and webpack.

I'm not sure that those extra packages will alter the build size at all anyway. I think webpack won't include anything extra in the bundle that is not required by the source. So I don't think there is any disadvantage. I will test out the build sizes though.

deployment/cloudbuild-deploy.yaml
tomlinton: Good point, I'll probably change them all to sh because I'm not sure all those containers have bash.

deployment/utils/gcb-discord-notifications/index.js
tomlinton: Good catch! This was me debugging. :grin:

deployment/cloudbuild-deploy.yaml
tomlinton: Yea that is right, so I'll define a trigger for something like changes on `origin-dapp/src/*` and then give it the _CONTAINER variable `origin-dapp`.

origin-js/package.json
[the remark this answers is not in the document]
franckc: Got it. Thanks for the details.

Yes, you are right, I don't think the build size will be impacted by adding those dependencies. But when developers in the future will install the origin-js package, it will pull those unnecessary dependencies which is not great. See https://www.npmjs.com/package/origin

Hmm.... that's tricky. One idea could be to have a base container that contains node and all the tools/npm packages we need to build. And use that base container rather than "FROM node:9". A drawback of this approach is that, assuming we still want folks to have those tools installed locally when they develop on their host, then they would also need to be in the package.json of origin-js and other packages... and it may become difficult to keep in sync the definition of that base container with the package.json from those various packages ?

deployment/dockerfiles/dev/event-listener
```diff
@@ -2,32 +2,25 @@
 
 FROM node:9 as origin-js-build
```
franckc: For Travis CI tests, we use node:8 - would probably makes sense to be consistent ?

origin-js/package.json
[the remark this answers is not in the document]
tomlinton: Looking into this further I realised that these extra dependencies will get unnecessarily pulled when origin-js is installed, but since origin-js is an in browser bundle compiled by webpack none of the dependencies in the production dependency list need to be installed. If a developer runs `NODE_ENV=production npm install origin-js` ~1500 packages will be installed that aren't needed at all. So I think there is a larger problem here.

If you checkout the [package.json](https://github.com/vuejs/vue/blob/dev/package.json) for Vue you can see that everything is listed under devDependencies for this reason.

I'm not sure how to solve this but I suggest we deal with this as a separate issue.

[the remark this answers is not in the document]
franckc: Great point regarding no need to pull those dependencies since we use webpack to bundle...
Agreed we can handle as a separate issue from this CI work. Do you want to file an issue to keep track of this ? I'll approve this CL in the meantime...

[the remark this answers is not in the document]
tomlinton: Thanks, raised an issue for this!

deployment/dockerfiles/dev/event-listener
tomlinton: Yep good plan, I will update these.

tomlinton:
FYI I'm going to centralise all the environment variables that are in the Dockerfiles in EnvKey. I'll add it to this PR. It's going to delay this a bit but I think it is worth doing here.

tomlinton:
@franckc this is ready to go again if you want to give it another quick pass. I made a bunch of changes, but the key points are:

- Moved everything I could to EnvKey
- Removed the separate Dockerfiles for each environment. Instead a single Dockerfile now pulls the configuration from EnvKey at build time for the DApp. For all other apps the EnvKey configuration is loaded at runtime.
- Moved some container startup scripts out of Kubernetes config and into the actual containers to avoid that issue we ran into the other day.

It's too big for you to review all of this, but if we just merge it we can iron out any issues on dev and staging. I left the prod setup the same so we can still deploy there in an emergency.

franckc:
Awesome. I'll review first thing tomorrow morning. Then let's get this in and try it out on dev and staging.

origin-dapp/src/services/origin.js
```diff
@@ -10,12 +10,9 @@ import OrbitDB from 'orbit-db'
  * But Micah couldn't get it to connect ¯\_(ツ)_/¯
  */
 const defaultProviderUrl = process.env.PROVIDER_URL
-const defaultBridgeUrl = 'https://bridge.originprotocol.com'
 const bridgeProtocol = process.env.BRIDGE_SERVER_PROTOCOL
 const bridgeDomain = process.env.BRIDGE_SERVER_DOMAIN
-const customBridgeUrl = `${bridgeProtocol}://${bridgeDomain}`
-const hasCustomBridge = bridgeProtocol && bridgeDomain
-const bridgeUrl = hasCustomBridge ? customBridgeUrl : defaultBridgeUrl
+const bridgeUrl = `${bridgeProtocol}://${bridgeDomain}`
```
franckc: As opposed to using 2 env vars, one for PROTOCOL and the other for DOMAIN, I'd be in favor of a single var call URL that stores both. This is not only for here, I'm thinking as a general thing for our codebase. We don't need to fix it in this PR though.

origin-dapp/webpack.config.js
```diff
@@ -14,32 +14,32 @@ const isProduction = process.env.NODE_ENV === 'production'
...
-  MAINNET_DAPP_BASEURL: null,
+  IPFS_SWARM: 'None',
+  MESSAGING_ACCOUNT: null,
+  MESSAGING_NAMESPACE: null,
+  MAINNET_DAPP_BASEURL: 'https://dapp.originprotocol.com',
   NOTIFICATIONS_KEY: null,
-  NOTIFICATIONS_URL: null,
-  RINKEBY_DAPP_BASEURL: null,
-  PROVIDER_URL: '',
+  NOTIFICATIONS_URL: 'notifications.originprotocol.com',
```
franckc: Should this have https:// ? This is how I configured it in ENVKEY...

deployment/utils/gcb-discord-notifications/index.js
```diff
@@ -0,0 +1,95 @@
...
+module.exports.subscribe = async (event, callback) => {
+  const build = eventToBuild(event.data);
+
+  const status = ['WORKING', 'SUCCESS', 'FAILURE', 'INTERNAL_ERROR', 'TIMEOUT'];
+  if (status.indexOf(build.status) === -1) {
+    return;
+  }
+
+  // Send message to Discord.
+  const webhook = await createDiscordWebhook(build);
```
franckc: paranoia:  test for webhook.message not being undefined before calling hook.send, since if for whatever reason status.build is not something expected, I think createDiscordWebhook would return webhook.message undefined.

origin-dapp/src/services/origin.js
tomlinton: Yea agreed, I think this is this way because it is easier to display it in the connectivity dropdown without having to parse a URL and extract the hostname. A similar thing applies to the IPFS settings where there are four environment variables to set that should be replaced with just a URL.

PR #848, opened 2018-10-30
fix bug that crashed ipfs proxy when 404 status should be reported

sparrowDom:
### Description:

Prevents a crash of `ipfs-proxy` container when 404 response should be returned


sparrowDom:
thanks @cuongdo

PR #862, opened 2018-11-01
Add NAMESPACE env var to event-listener container

franckc:

### Checklist:

### Description:

NAMESPACE needs to be defined since it is used by start-event-listener.sh

Now, after that is fixed, there is still something broken I think with the event-listener webhook to the notification server. From a shell within the event-listener container, I could not connect to the notification server:
```# curl http://notifications.staging.svc.cluster.local:3456 
curl: (6) Could not resolve host: notifications.staging.svc.cluster.local
```


franckc:
oh. I see. thanks for fixing the URL !

Issue #814, opened 2018-10-24
Mobile Seller Onboarding Modal


Natasha08:
Note: This issue was created based on testing in Coinbase Wallet (previously Toshi).

When a user is on mobile without wallet functionality they cannot create listings or verify their identities. Even when a user does have a mobile wallet, the onboarding mobile blocks the messaging UI.

- [x] For mobile browsers hide the seller onboarding modal.

micahalcorn:
Resolved by #865 ✅

PR #865, opened 2018-11-01
hide the onboarding modals when on mobile

Natasha08:
### Checklist:

### Description:
[Issue #814](https://github.com/OriginProtocol/origin/issues/814)

- Hide the onboarding modal on mobile devices (including iPad-size devices)

### Manual Testing
- You may need to change `src/components/onboarding` line `245` from `{ networkType === 'Mainnet Beta' && <BetaModal /> }` to `<BetaModal />`


micahalcorn:
@Natasha08 we're currently setting an `onMobile` flag in Redux, which is determined [here](https://github.com/drcmda/react-spring). Can we consolidate this and your approach? I'm not sure which detection method is best.

wanderingstan:
@micahalcorn that link for `onMobile` goes to an animation library. Can you re-send? I didn't realize we had that var being set either!

micahalcorn:
@wanderingstan looks like a clipboard mixup, sorry: https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/components/app.js#L133-L141

Natasha08:
@wanderingstan @micahalcorn Looking at the var in Redux, we're setting which device the user is on based on the user agent string in the `detectMobile` method in `src/components/app`.

I am concerned about only using whether 'Android' or the iOS devices' names are in the user agent string as a determination of whether or not the user is on mobile. I'm thinking about how to combine them, and I see two separate use cases:
- Some components need to know whether the user is on iOS or Android;
- Sometimes we need to know if they are on a mobile device of any kind, or a screen small enough to qualify as mobile

Natasha08:
@micahalcorn @wanderingstan I have add both approaches to the global Redux state with some modifications. Mainly I moved detecting by width from the `isMobile` method; I'm not sure if we want to detect by width at all

Natasha08:
I'm good to merge this

origin-dapp/src/utils/mobile.js
```diff
@@ -0,0 +1,24 @@
+const userAgent = navigator.userAgent || navigator.vendor || window.opera
+
+export default function isMobile() {
```
wanderingstan: I'm generally a proponent of using fewer vars and avoiding any chance at data corruption. (aka data normalization) Seems like `isMobile` and `mobileDevice` could be consolidated since a `null` value on mobileDevice means that we're not on mobile.

Having both opens up the possibility of `isMobile` being true and `mobileDevice` being null (or vice versa) which would lead to strange behavior. 

@micahalcorn You have opinions on this?

micahalcorn: @wanderingstan I agree, but I suspect that @Natasha08 was keeping the old detection method so that we maintained the ability to distinguish between mobile vendors. Should we keep both methods and rename `mobileDevice` to avoid confusion, or should we just return the OS vendor instead of `true` in a single detection method?

wanderingstan: Yeah, we should drop `isMobile()` and add it's logic as fallback in `mobileDevice()`  such as 
```
export function mobileDevice() {
  if (/android/i.test(userAgent)) {
    return 'Android'
  } else if (/iPad|iPhone|iPod/.test(userAgent)) {
    return 'iOS'
  } else if (userAgent.match(/Mobi/)) {
    return 'Unknown Mobile'
  } else {
    return null
  }
}

@Natasha08 what do you think?

```

PR #869, opened 2018-11-01
[Notifications] Do no crash server when receiving non offer related events

franckc:

### Checklist:

### Description:
A few fixes to harden the notification server and its integration in the event listener:
 - Fix bug in notifier where non offer events (for ex. ListingCreated) would cause a crash
 - Do not crash the listener if a webhook calls fails: after maximum number of reties reached, skip the webhook call.

### DIFF
--- a/origin-discovery/listener/listener.js
+++ b/origin-discovery/listener/listener.js
@@ -488,14 +488,24 @@ async function handleLog(log, rule, contractVersion, context) {
 
   if (context.config.webhook) {
     console.log('\n-- WEBHOOK to ' + context.config.webhook + ' --\n')
-    await withRetrys(async () => {
-      await postToWebhook(context.config.webhook, json)
-    })
+    try {
+      await withRetrys(async () => {
+        await postToWebhook(context.config.webhook, json)
+      }, exitOnError=false)
+    } catch(e) {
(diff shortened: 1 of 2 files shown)


origin-discovery/listener/listener.js
```diff
@@ -488,14 +488,24 @@ async function handleLog(log, rule, contractVersion, context) {
...
     console.log('\n-- WEBHOOK to ' + context.config.webhook + ' --\n')
-    await withRetrys(async () => {
-      await postToWebhook(context.config.webhook, json)
-    })
+    try {
+      await withRetrys(async () => {
+        await postToWebhook(context.config.webhook, json)
+      }, exitOnError=false)
+    } catch(e) {
+      console.log(`Skipping webhook for ${logDetails}`)
```
tomlinton: What is the consequence of this? I'm thinking the worst thing that can happen is that people don't receive notifications, but nothing worse than that?

franckc: Yes, exactly. I think it is better to not send those notifications vs entirely stopping the event-listener which will impact indexing and freshness of the data displayed in the DApp (since soon we'll rely on back-end data for powering DApp).
Ideally we would log a stat here that would trigger alerting so that we can quickly take a look.

PR #873, opened 2018-11-02
Add EnvKey support to discovery server

tomlinton:

### Checklist:

### Description:

- Add EnvKey support
- Made port configurable via env

### DIFF
--- a/origin-discovery/apollo/index.js
+++ b/origin-discovery/apollo/index.js
@@ -1,5 +1,8 @@
+require('dotenv').config()
+require('envkey')
+
 const express = require('express')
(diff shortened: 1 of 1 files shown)


origin-discovery/apollo/index.js
```diff
@@ -1,5 +1,8 @@
+require('dotenv').config()
+require('envkey')
```
franckc: should this be in a try/catch statement so that when running in local dev environment we setting ENVKEY is optional ? like same as what you did on event-listener: https://github.com/OriginProtocol/origin/blob/master/origin-discovery/listener/listener.js#L2

PR #879, opened 2018-11-03
Fixes for migration scripts in cloudbuild deployments

tomlinton:

### Checklist:

### Description:

- Fixes issue where EnvKey variables were not available to migration scripts
- Tweaked announcement lengths for Discord deployment announcements

### DIFF
--- a/deployment/dockerfiles/event-listener
+++ b/deployment/dockerfiles/event-listener
@@ -13,6 +13,11 @@ RUN npm run build
 
 FROM node:8
 
+# Install envkey-source
+RUN curl -s -L -o envkey-source.tar.gz https://github.com/envkey/envkey-source/releases/download/v1.2.2/envkey-source_1.2.2_linux_amd64.tar.gz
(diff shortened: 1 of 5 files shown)


deployment/dockerfiles/event-listener
```diff
@@ -13,6 +13,11 @@ RUN npm run build
 
 FROM node:8
 
+# Install envkey-source
+RUN curl -s -L -o envkey-source.tar.gz https://github.com/envkey/envkey-source/releases/download/v1.2.2/envkey-source_1.2.2_linux_amd64.tar.gz
```
franckc: Just curious: why not using the install.sh script as documented [here](https://github.com/envkey/envkey-source#docker) ?

tomlinton: It's less of a security risk doing it manually like this. You can imagine a scenario where someone replaces that install.sh script with something nasty and it gets pulled into the CI server and run. Someone could still replace the contents of that release tar with something malicious, but it is probably less likely that'd happen.

PR #881, opened 2018-11-05
[Discovery] Add linter, unit tests and Travis CI build integration

franckc:

### Checklist:

### Description:
 - Moved code under src/ to be consistent with the structure of other origin subdirectories
 - Configured linter in package.json, ran it against discovery codebase and fixed the warnings/errors it found. Note: I used the standard eslint parser rather than the babel-eslint parser since origin-discovery is purely back-end and does not need to be compiled with babel to be browser compatible.
 - Added unit tests scaffolding and implemented a few unit tests for the listener (more unittests to follow in separate PRs). Also configured unit tests to run linter.
 - Added an origin-discovery build in Travis CI and configured it to start a Postgres 9.6 (with JSONB support) so that it can be used for by unittests (this is very handy !).


tomlinton:
The automated deployment merge made a bit of a mess of this. I'll take care of resolving those conflicts tomorrow!

origin-discovery/scripts/test.js
```diff
@@ -0,0 +1,20 @@
+const chalk = require('chalk')
+const testJavascript = require('./helpers/test-javascript')
+const testJSFormat = require('./helpers/test-js-format')
```
tomlinton: Could we simplify this by just adding `mocha --timeout 10000 --exist && eslint src/*/*.js test/**/*test.js scripts/**/*.js` to `package.json` as the test command? I know this is how it is done for `origin-js` but it seems like a lot of extra boilerplate.

sparrowDom: In `origin-js` we seem to have both approaches, one in [package.json](https://github.com/OriginProtocol/origin/blob/master/origin-js/package.json#L21) and another as [standalone JS file](https://github.com/OriginProtocol/origin/blob/master/origin-js/scripts/helpers/test-contracts.js), duplicating the functionality.

I guess we need the ability to group different test types into a general test command, and for that reason there is the boilerplate code that wraps test commands. I personally am ok with boilerplate code, just for the reason, of using the same approach as in `origin-js`.

sparrowDom: Hmm I guess we could also do away with a lot of boilerplate like @tomlinton suggests and group scripts in `package.json` like so: 
```
"test:js": "node helpers/test-javascript.js", // script change to runnable
"test:js-format": "node helpers/test-js-format.js", // script change to runnable
"test": "npm run test:js && npm run test:js-format"
```

I am ok with either approach, though it would be nice to be consistent across all the repos.

franckc: I was thinking in the future we may need those helper scripts to start local services like blockchain and IPFs as part of running the tests. But let's keep things simple for now and we can cross that bridge when we need it. Based on feedback, remove boilerplate but used @sparrowDom's idea to have test:js, test:jsformat and test

franckc:
Let me work on those merge conflicts. I will ask for help if I get confused while doing those @tomlinton....

PR #882, opened 2018-11-05
Basic support for editable listings

jordajm:
### Checklist:

### Description:

- Adds three new methods to the Origin JS marketplace API: `updateListing()`, `getPurchases()` and `getSales()`
- The getPurchases() and getSales() methods return the listing data as it was when a particular offer was made, even if it has been edited since then
- Removes some very hairy logic from the dapp layer and replaces it with a simple call to `origin.marketplace.getPurchases()` or `origin.marketplace.getSales()`
- Adds support for using the "create listing" workflow as a listing editor - i.e. `/update/<listing-id>` takes the user to Step 2 (details form) of the create listing workflow with all listing data pre-populated.
- NOTE: Users will see the latest listing data in the listing grid and listing detail page, but "my sales" and "my purchases" will show the listing data as it was at the time of the purchase 🎉 💥 

**TODO** (probably in a follow-up PR)
- ~~Add unit tests for new marketplace API methods~~
- Add "edit" buttons to all appropriate places in the dapp

**Resolves issue:** https://github.com/OriginProtocol/origin/issues/928

--------------------------------------

### Listing grid (latest listing data)
screenshot 2018-11-05 12 21 39

--------------------------------------

### "My Purchases" (listing data as it was at time of purchase - notice that it doesn't have the "EDITED!" text in the title)
screenshot 2018-11-05 12 21 21


origin-js/src/contractInterface/marketplace/v00_adapter.js
```diff
@@ -250,7 +269,9 @@ class V00_MarkeplaceAdapter {
       if (event.event === 'ListingCreated') {
         ipfsHash = event.returnValues.ipfsHash
       } else if (event.event === 'ListingUpdated') {
-        ipfsHash = event.returnValues.ipfsHash
+        if (!offerBlockNumber || offerBlockNumber < event.blockNumber) {
+          ipfsHash = event.returnValues.ipfsHash
+        }
```
jordajm: @wanderingstan - this is the code that enables us to show the listing data as it was when the listing was created for "my purchases" and "my sales"

jordajm: Actually, this may not be necessary - something is working differently than I expected... still looking into it... I'm trying to grok this code - I don't think I know what it's doing actually:
https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/components/my-sales.js#L36-L61

jordajm: lol - okay - yes, this code is needed and works now - temporarily thought I was losing my marbles 😅

origin-js/src/contractInterface/marketplace/v00_adapter.js
```diff
@@ -241,6 +251,15 @@ class V00_MarkeplaceAdapter {
       fromBlock: this.blockEpoch
     })
 
+    // if an account is passed in, we're looking for the listing state at the time
+    // that account made an offer, so find the block number of the offer
+    // and ignore any "ListingUpdated" events after that block number
```
wanderingstan: I may not fully get what's going on here, but it seems that it may cause trouble to use `account` for getting old versions of a listing. For example, couldn't an account have multiple offers open on the same listing? E.g. I bought one beanie baby, but then the next day the seller dropped the price so I decided to buy another one. 

Seems more generic and natural to either (a) pass in an offerId to get the listing as it exists for that offer or (b) pass in a block number and let the dapp do its own logic to figure out which block number it wants.

origin-dapp/src/components/listing-detail.js
```diff
@@ -187,32 +187,9 @@ class ListingsDetail extends Component {
...
-      const withOffers = await Promise.all(offerPromises)
-      const offersByListing = withOffers.map(obj => {
-        return obj.offers.map(offer => Object.assign({}, obj, { offer }))
-      })
-      const offersFlattened = [].concat(...offersByListing)
-
-      this.setState({ purchases: offersFlattened })
+      const purchases = await origin.marketplace.getPurchases(web3Account)
+      const transformedPurchases = transformPurchasesOrSales(purchases)
+      this.setState({ purchases: transformedPurchases })
```
jordajm: This is an example of the simplification of the logic in the DApp layer @micahalcorn @wanderingstan

origin-dapp/src/components/my-purchases.js
```diff
@@ -33,32 +33,9 @@ class MyPurchases extends Component {
...
-    const withOffers = await Promise.all(offerPromises)
-    const offersByListing = withOffers.map(obj => {
-      return obj.offers.map(offer => Object.assign({}, obj, { offer }))
-    })
-    const offersFlattened = [].concat(...offersByListing)
-
-    this.setState({ loading: false, purchases: offersFlattened })
+    const purchases = await origin.marketplace.getPurchases(web3Account)
+    const transformedPurchases = transformPurchasesOrSales(purchases)
+    this.setState({ loading: false, purchases: transformedPurchases })
```
jordajm: Same simplification here

origin-dapp/src/components/my-sales.js
```diff
@@ -34,30 +34,11 @@ class MySales extends Component {
...
-    const offersByListing = withOffers.map(obj => {
-      return obj.offers.map(offer => Object.assign({}, obj, { offer }))
-    })
-    const offersFlattened = [].concat(...offersByListing)
+    const { web3Account } = this.props
+    const sales = await origin.marketplace.getSales(web3Account)
+    const transformedSales = transformPurchasesOrSales(sales)
 
-    this.setState({ loading: false, purchases: offersFlattened })
+    this.setState({ loading: false, purchases: transformedSales })
```
jordajm: And here

jordajm:
@micahalcorn @wanderingstan - this is ready for review. There's a lot here... sorry 😅 - I had a hard time finding a good stopping place without getting all of the editable listings feature sets working together. LMK and we can run through it on a call if needed.

franckc:
Nice work on this @jordajm !

Based on our discussion this afternoon:
 - Let's use (blockNumber, logIndex) rather then just blockNumber for the event sourcing logic.
 - On the back-end side, I'll make sure the "performance mode" for marketplace.getListings supports the purchasesFor and listingsFor options; and that getListing can support (blockNumber, logIndex).

Now, thinking a bit more about the "complete schema", I'm still not 100% sure we need a JSON schema to validate the model right now. IMHO currently the biggest difficulty when using a model is that it is pretty difficult to know what fields a model includes. In particular the fields coming from the marketplace logic (aka chain data). A complete schema would provide that information but it seems a bit heavy, especially given that we are going to iterate quite a bit on listing types and that the Listing model is far from being stable. Based on that, I'd lean towards starting with a lighter weight solution which is to be explicit in the model about what fields we set. e.g. as opposed to what we do currently which is `Object.assign(this, ipfsListing, chainListing)`, doing something like:
```
this.title = ipfsListing.title
this.description = ipfsListing.description
...
this.status = chainListing.status
this.events = chainListing.events
...
```
Also, one last though is that the complete schema and editable listing changes could be worked on as separate smaller PRs rather than one big PR...  I personally always find it easier to reason about and review smaller PRs...

I do make a lot of typo and mistakes when typing so I'm really looking forward the editable listing feature :)

jordajm:
@franckc also removing all of the stubbed-in "complete" schema validation stuff per your comment above and based on our conversations. 

I agree that we could possibly make it explicit in the Listing and Offer models what data will be included in the output of the model, I think I'd prefer to just include a big JS Doc style comment that clearly defines the output data structure. 

Let's discuss that when the time comes (probably later this week if I can get to it)

origin-js/src/resources/marketplace.js
```diff
@@ -69,11 +138,20 @@ export default class Marketplace {
...
-    return Promise.all(
-      listingIds.map(async listingId => {
-        return await this.getListing(listingId)
-      })
-    )
+    if (opts.withBlockInfo) {
+      return Promise.all(
+        listingIds.map(async listingData => {
+          const { listingId, blockInfo } = listingData
+          return await this.getListing(listingId, blockInfo)
```
franckc: I'm working on implementing this logic on the discovery side. Quick clarifications:
 - On line 145, why do we need to pass blockInfo when calling this.getListing ? I thought withBlockInfo options just indicates that blockInfo for the latest version of each listing should be returned along with listingId ?
 - at a higher level, what is actually the use case for the withBlockInfo options on getListings ? My understanding was that for editable listings, blockInfo always comes from data on the offer, and then we make individual calls to getListing and pass the offer blockInfo to fetch the listing at that point of time. Do we actually need the withBlockInfo option on getListings ? I could see that for optimization we could perhaps define a new function getListingsById([(listingId1, blockNum1, logIndex1), ..., (listingIdN, blockNumN, logIndexN)) to batch retrieve a list of listings as part of the getPurchases implementation. But that's separate.

Happy to jump on a call to chat if faster compared to commenting on this PR :)

jordajm: @franckc 
>On line 145, why do we need to pass blockInfo when calling this.getListing ?

The `blockInfo` object is used [here](https://github.com/OriginProtocol/origin/blob/62173514375f77b6537399f1af96ed9e5f4a0bc8/origin-js/src/contractInterface/marketplace/v00_adapter.js#L266) so to return the listing data as it was when an offer was made. This is different than the `withBlockInfo` boolean option that can be passed into `getListings()`. If you think that's too confusing or if you have an idea for how to make it better, let me know!

------------------------------------------------

>at a higher level, what is actually the use case for the withBlockInfo options on getListings ?

Yeah, you've got it right... I think this could definitely use a refactor at some point to make it less confusing, but when you call `getListings()` with `opts.purchasesFor`, it's actually fetching the `OfferCreated` events ([here](https://github.com/OriginProtocol/origin/blob/62173514375f77b6537399f1af96ed9e5f4a0bc8/origin-js/src/contractInterface/marketplace/v00_adapter.js#L385-L388)) and that's where we get the block info. 

In reality, `withBlockInfo` only works with the `PurchasesFor` option, which is even more reason for a future refactor IMO. However, this is as clean as I can figure out how to make it without really changing the scope of this PR.

franckc: Oh dear.... Thanks for the details ! Yes, it is pretty confusing ! :)
A few ideas/comments:
 - if withBlockInfo only works in conjunction with PurchasesFor, do we need both ? could we just rely on purchasesFor ?
 - let's add comments throughout the code to help the reader follow thru this logic. otherwise I have the feeling we'll get confused every time we look at this code
 - strong unit test coverage will help us being more confident we can refactor without breaking things in the future. so the more unit test the better in my opinion :)

jordajm: > if withBlockInfo only works in conjunction with PurchasesFor, do we need both ? 

They could probably be combined, but IMO this should be left for the refactor because it changes the way the `opts` work in a fundamental way by preventing the user from getting listing ID's only (not full listing data) in certain cases, and I'd want to make sure we aren't using that and that there's not a good use case for it before we remove it... I know it's messy, but I'd rather do it as part of a refactor that cleans all of this stuff up at once and tests thoroughly that we aren't breaking anything... sound okay to you?

jordajm:
Okay @franckc - I added some unit tests and comments and I think this baby is ready to merge 😁 
@wanderingstan and I reviewed it on a call yesterday and he gave his 👍as well
LMK if you have any final feedback, otherwise, feel free to merge when you're ready.

franckc:
@jordajm Sounds good ! Let's hold off just a little bit before merging this PR to master. Here is the sequencing I'm thinking of, let me know if that sounds good to yo:
1. [Today+ Thursday] Rollout to staging and prod new "performance mode".
2. [Friday] Merge this PR to master.
3. [Friday] Merge Franck's PR to add support for editable listing when performance mode enabled.
4. [Wed 11/20, before Thanksgiving] Deploy to staging. Test thoroughly.
5. [Mon 11/26] Deploy to prod.

CC @sparrowDom @wanderingstan

franckc:
@wanderingstan and @micahalcorn:  you ok if we merge this PR in master ?

@jordajm Looks like there is a couple of branch conflicts. Do you mind taking a quick look ?

wanderingstan:
Good by me, I knew you had made a few requests. 

(@franckc - I agree we need a side effort to document/explain some of the "routing" through origin-js which is getting hard to follow.)

PR #884, opened 2018-11-05
Nick/extract origin contracts

nick:
### Checklist:

- [x] Update any relevant READMEs and [docs]

### Description:

This PR extracts the contracts from origin-js to origin-contracts. It also switches the monorepo to using [Lerna](https://github.com/lerna/lerna) for managing our packages. This will enable us to manage multiple packages that depend on each other more easily. Lerna also mean we no longer need `npm link` for managing internal dependencies.

Steps to test this pull request:

```
git clone git@github.com:OriginProtocol/origin.git origin-tst
cd origin-tst
git checkout nick/extract-origin-contracts
npm install
cd origin-js && npm start
```

In a new console, once origin-js is running:

```
cd origin-tst/origin-dapp
npm start
```

To run origin-js and origin-contracts tests (we can add others too):

```
cd origin-tst
npm test
```


.travis.yml
```diff
@@ -1,29 +1,24 @@
 matrix:
   include:
-  - name: "DApp unit tests"
+  - name: "origin-contracts / origin-js / origin-dapp unit tests"
```
franckc: By combining all these test targets, we lose the benefits of running tests in parallel.
I understand the install will be done once for all these targets and that's a gain.
But the tests themselves may take some time to run and by losinig parallelization we may overall be worse off in terms of overall time it takes to run the Travis CI build ?

Travis CI gives a break down of how much time is spent running each step (e.g. npm install vs npm test) so we should be able to get data to decide...

.travis.yml
```diff
@@ -1,29 +1,24 @@
...
     language: node_js
-    node_js: 8
+    node_js: 10
```
franckc: +1 for upgrading to node 10.
Now, if we use it for testing, we should also switch our prod containers to using it. See for example [here](https://github.com/OriginProtocol/origin/blob/master/deployment/dockerfiles/event-listener#L3) and in other containers in that directory.

.travis.yml
```diff
@@ -1,48 +1,60 @@
+sudo: false
+git:
+  depth: 3
 matrix:
+  fast_finish: true
   include:
-  - name: "DApp unit tests"
```
franckc: I don't see those being run in the new version of this file. But we probably want to keep those DApp unit tests since they run the linter which is useful to enforce consistent coding style for the DApp ?

PR #887, opened 2018-11-06
Fetch listing data from discovery server

franckc:

### Checklist:

### Description:
Add an option for the DApp to fetch Listing data from the discovery server rather than the blockchain. This is called "performance mode" since it optimizes for latency.
 - When performance mode is enabled, origin-js methods maketplace.getListing(s) fetch listing data via the DiscoveryService rather than web3/IPFS.
 - The ENABLE_PERFORMANCE_MODE env var is used to gate this feature. 
    - local: We don't want to break developers development workflows. So to start with, we won't enable the performance mode until we have more confidence that origin-box (or an alternative) can reliably run the event-listener and discovery server.
    - production: we'll enable performance mode on dev and staging. Then if it all looks good, we'll enable in prod.

How to test:
 - In local environment, update origin-dapp/.env file to include:
```
DISCOVERY_SERVER_URL=http://localhost:4000/graphql
ENABLE_PERFORMANCE_MODE=true
```

Things left to do (as separate PRs):
 - Implement listingsFor and purchasesFor options for the marketplace.getListings method. Those options are used by the DApp.
 - Implement hidden/featured listings in the back-end
 - Fix DApp pagination: currently it requests *all* listings, then figures the number of pages. This is not scalable (even when fetching data from the back-end, if we have a few thousand listings it would be very slow to fetch them all).


sparrowDom:
LGTM!

One possible approach for testing discovery service would be to mock/intercept the `_query` method other functions are calling after creating the graphQL query and check if the generated query string is expected.

franckc:
Yes, good idea. I'm looking into using fetch-mock to mock the response from the discovery server in those tests.

Issue #568, opened 2018-09-24
Browse page: fetch listingIds from back-end rather than web3


franckc:
This improve dramatically performance of the browse page on Mainnet.

**TODO:**
- [x] read listings from discovery server
- [x] create devops script that uses index aliases, reindexing

Devops script Pull Request: https://github.com/OriginProtocol/origin/pull/844
Implement hidden/featured listings in Apollo PR: https://github.com/OriginProtocol/origin/pull/890

sparrowDom:
@franckc thanks for creating this. 

@tomlinton since we will be relying on search more from now on (the listings on home-page are going to be queried using elasticsearch) are there any devops tools that you would critically require for launch? Perhaps something like: 
- script that deletes the listing index
- script that applies (like database migrations) the mappings/ index schema
- something else?

tomlinton:
@sparrowDom yes if we want to change the indexes post launch we'll need something to handle that. A way to do migrations would be amazing, but I'm not too sure what is available with elasticsearch. Initially a script that would deletes an index and recreates it would probably be fine. I started putting together a script that creates the indexes based on what you added to the elasticsearch dockerfile [here](https://github.com/OriginProtocol/origin-box/blob/master/development/scripts/create-es-index.sh). I was planning on just adding a delete of the index at the start. We can then update this script and run it to change the indexes if needed.

Do you think that will suffice? I doubt we'll be changing the indexes too much anyway.

sparrowDom:
Hey @tomlinton yeah that is a good start, the only downside is that this causes downtime every time we need to alter indexes. As part of the solution to this issue I will prepare a script that creates index aliases. So the steps for index mutation would be: 
- create an alias for listings index
- when index mutation is required create a fresh index with the updated mappings schema
- reindex all the content from the original index that is currently live to the freshly created index
- switch aliases from the old index to the new one
- delete old index (this step should be manual, so we can fallback any time)

Does that make sense to you?

tomlinton:
Yep that is a much better approach if you think it is achievable. Go for it. :+1:

sparrowDom:
@tomlinton in [this branch](https://github.com/OriginProtocol/origin/tree/sparrowDom/esIndexMigration) I have prepared a script that should make index migrations easier and without downtime. The script is located here: `development/scripts/elasticsearch/migrationTool.sh` and this is how the functionality looks like: 
[...]

PR #890, opened 2018-11-07
Implement Hidden/Featured listings in Apollo

sparrowDom:
### Checklist:

- [ ] ~Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/origin-docs)~

**This is still a work in progress**

### Description:
Implement hidden/featured listings in apollo and remove the Dapp implementation. 

- move configuration to . env file - done
- implement hidden and featured listings in elasticsearch/apollo - done
- update boost query to preserve featured listings order - done
- remove featured/hidden listings implementation from the DApp - done

Tests for elasticsearch query generation are going to be done in another PR.

Related to issue: https://github.com/OriginProtocol/origin/issues/568


origin-discovery/src/apollo/app.js
```diff
@@ -27,12 +28,53 @@ const bundle = promBundle({
 })
 app.use(bundle)
 
+// how frequently featured/hidden listings list updates
```
franckc: I'd recommend moving all this new logic into a separate file. maybe called info.js ?

origin-discovery/src/apollo/resolvers.js
```diff
@@ -22,94 +22,99 @@ function relatedUserResolver (walletAddress, info) {
...
-        .search(
-          args.searchQuery,
-          args.filters,
-          args.page.numberOfItems,
-          args.page.offset,
-          true // idsOnly
-        )
-      // Get listing objects from DB based on Ids.
-      const listings = await getListings(listingIds)
+const getResolvers = function (listingInfo) {
```
franckc: Passing listingInfo works but it might be simpler to just access the listingInfo global variable ? (I think it would simplify the code a bit).
That's anyway what the updateHiddenFeaturedListings method already does...

origin-js/src/resources/discovery.js
```diff
@@ -77,6 +78,17 @@ class Discovery {
 
     return this.query(query)
   }
+
+  async getListing(listingId) {
```
franckc: What's the use for this method ? Currently the DApp gets listing data by calling marketplace.getListing - it might be simpler to keep it that way since then it works both in decentralized and centralize mode ?

origin-discovery/src/apollo/schema.js
```diff
@@ -65,6 +70,7 @@ const typeDefs = gql`
     subCategory: String!
     price: Price!
     offers: OfferConnection
+    displayType: DisplayType!
```
franckc: Naming is hard. How about keeping things shorter and just using "display" and "Display" for the type ? It's an enum anyway so when browsing the schema via graphql it is pretty explicit what this is...
`display: Display!`

development/envfiles/origin-discovery.env
```diff
@@ -0,0 +1,8 @@
+DATABASE_HOST=postgres
```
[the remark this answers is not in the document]
sparrowDom: I see, thanks... removing others

origin-discovery/src/apollo/app.js
sparrowDom: 👍

origin-js/src/resources/discovery.js
sparrowDom: Thanks yes, this is no longer needed. Deleting it

origin-dapp/src/components/listings-grid.js
```diff
@@ -29,50 +29,22 @@ class ListingsGrid extends Component {
   }
 
   render() {
-    const { contractFound, listingIds, search, featuredListingIds, hiddenListingIds } = this.props
-    let activePage, currentPageListingIds, includedFeaturedListingIds, listingIdsNotHidden, resultsCount
+    const { contractFound, listingIds, search, featuredListingIds = [] } = this.props
```
franckc: Can we remove featuredListingIds ? Looks like it is only used for passing the "featured" property to the ListingCard, but the ListingCard should be able to get that from the listing data itself so the "featured" property could be removed ?

origin-dapp/src/components/search/search-result.js
```diff
@@ -204,15 +204,19 @@ class SearchResult extends Component {
...
+      const searchResp = await origin.discovery.search({
+        searchQuery: this.props.query || '',
+        numberOfItems: LISTINGS_PER_PAGE,
+        offset: (this.state.page - 1) * LISTINGS_PER_PAGE,
+        filters: Object.values(filters).flatMap(arrayOfFilters => arrayOfFilters)
+      })
 
       this.setState({
-        listingIds: searchResp.data.listings.nodes.map(listing => listing.id),
+        listings: searchResp.data.listings.nodes.map(listing => {
```
franckc: This looks a bit hacky :)
Do we need to store the whole listing data in the state ? Or could could we just store both listingIds and featuredIds ?
e.g. something like this:
```
this.setState({
  listingIds: searchResp.data.listings.nodes.map(listing =>listing.id),
  featuredIds: searchResp.data.listings.nodes
    .filter(listing => listing.display === 'featured')
    .map(listing => listing.id)
  ...
})
```

origin-discovery/src/apollo/db.js
```diff
@@ -7,7 +7,7 @@ const db = require('../models')
  * @param listingIds
  * @return {Promise<Array>}
  */
-async function getListings (listingIds) {
+async function getListings (listingIds, hiddenIds, featuredIds) {
```
franckc: nit: document the new params hiddenIds and featuredIds
Also maybe add a TODO saying we should read those directly from DB within this method as opposed to getting them as params.

origin-discovery/src/apollo/db.js
```diff
@@ -30,6 +30,15 @@ async function getListings (listingIds) {
   // Note: preserve ranking by keeping returned listings in same order as listingIds.
   const listings = []
   listingIds.forEach(id => {
+    let display = "normal"
+    /* hidden listings are not returned right now, but at some point in the future
```
franckc: super nit: "returned" is a bit confusing. clarify that no hidden listing ids are passed to this method (since they get filtered at the layer above).

origin-discovery/src/apollo/listingMetadata.js
```diff
@@ -0,0 +1,38 @@
+// how frequently featured/hidden listings list updates
```
franckc: nit: I think our convention is that we don't use cameCase for names of files. So please rename to listing-metadata.js (or simply metadata.js, up to you). BTW I like the name metadata much better than info. good call on that ! :)

nit2: Add a high level comment at top of the file explaining what this does.
It loads and caches featured and hidden listing ids metadata.

origin-discovery/src/apollo/listingMetadata.js
```diff
@@ -0,0 +1,38 @@
...
+const LISTINGS_STALE_TIME = 60 * 1000 //60 seconds
```
franckc: nit: rename to METADATA_STALE_TIME or METADATA_STALE_THRESHOLD

origin-js/src/services/discovery-service.js
```diff
@@ -126,10 +128,16 @@ class DiscoveryService {
     if (opts.idsOnly) {
       return resp.data.listings.nodes.map(listing => listing.data.id)
     } else {
-      return resp.data.listings.nodes.map(listing => listing.data)
+      return resp.data.listings.nodes.map(listing => this.aggregateListingData(listing))
     }
   }
 
+  aggregateListingData(listingNode) {
```
franckc: nit: this is more "flattening" rather than aggregating. maybe "flattenListingData" would be more representative of what this method does ? and since it is private and does not need to be exposed, prefix with "_". So perhaps call it _flattenListingData.

supernit: move it at top of the file since in general it improves readability to have the code for a function first before it gets used.

origin-dapp/src/components/listings-grid.js
sparrowDom: Yes good point, leaving it to the listing card to get featured information.

origin-dapp/src/components/search/search-result.js
sparrowDom: Yeah, since featured badge is in the recent change handled by the `listing-card` we don't need the `featuredIds` anymore. So going back to only storing `listingIds`. (what you propose in example minus the `featureIds` property)

origin-discovery/src/apollo/db.js
sparrowDom: Good point 👍

origin-discovery/src/apollo/db.js
sparrowDom: Good catch, I have changed the comment.

origin-discovery/src/apollo/listingMetadata.js
sparrowDom: 👍

origin-discovery/src/apollo/listingMetadata.js
sparrowDom: thanks 👍

origin-js/src/services/discovery-service.js
sparrowDom: 👍 agreed

sparrowDom:
@franckc thanks for a good review. All the comments have been resolved. 

I will prepare the Apollo queries in another branch so this one doesn't get too big. I think this is getting ready to be merged.

origin-discovery/src/apollo/listing-metadata.js
```diff
@@ -0,0 +1,41 @@
+/* Catches hidden & featured listings ids and updates them according
```
franckc: typo: Caches

sparrowDom:
@franckc ok merging in master. 

Oh sorry about clumsy wording. I wanted to say, that I will do the tests that check that Apollo generates the expected query for `elasticsearch` in another branch.

origin-discovery/continueTest
```diff
@@ -0,0 +1 @@
+{"lastLogBlock":123,"version":1}
```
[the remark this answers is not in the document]
sparrowDom: nice catch. Removing it from git and adding it to `.gitignore`

origin-discovery/src/apollo/listing-metadata.js
sparrowDom: 😇

.gitignore
```diff
@@ -76,6 +76,10 @@ node_modules/
 # Flask Session files
 /origin-bridge/*flask_session/
 
+# origin-discovery #
+####################
+origin-discovery/continueTest
```
franckc: I think we should have the uni test delete this file (I can do that in a separate PR. ok to leave this in for now...)

sparrowDom: Its a quick and minor change.... a solution in this PR: https://github.com/OriginProtocol/origin/pull/902

PR #900, opened 2018-11-09
Implement marketplace.getListings options in back-end

franckc:

### Checklist:

### Description:
Implement the options listingsFor and purchasesFor of marketplace.getListing for the performance mode.

This is still WIP, I haven't tested the origin-js piece and it is missing unit tests. But wanted to share early for any feedback...


origin-discovery/src/apollo/resolvers.js
```diff
@@ -57,28 +58,40 @@ const resolvers = {
...
+      }
+      const rows = await db.Offer.findAll({ where: clause })
+      const offers = rows.map(offer => offer.data)
       return { nodes: offers }
     },
 
     async offer (root, args, context, info) {
-      return search.Offer.get(args.id)
+      const row = await db.Offer.findByPk({ id: args.id })
+      return (row.length === 0) ? row.data : null
```
sparrowDom: Should this line be: 
```
return row === null ? null : row.data
```

origin-discovery/src/apollo/resolvers.js
```diff
@@ -57,28 +58,40 @@ const resolvers = {
...
     },
 
     user (root, args, context, info) {
-      return search.User.get(args.walletAddress)
+      // FIXME(franck): some users did not get indexed in prod due to a bug in attestations.
+      // For now only return the address until data gets re-indexed.
+      return { walletAddress: args.walletAddress }
```
sparrowDom: Could we return null if user is not found and a user object when it is?

origin-discovery/src/apollo/resolvers.js
```diff
@@ -93,21 +106,32 @@ const resolvers = {
...
       const requestedSubFields = info.fieldNodes[0].selectionSet.selections
-      const isIdOnly = requestedSubFields.filter(x => x.name.value !== 'id').length === 0
-      if (isIdOnly) {
+      const idsOnly = requestedSubFields.filter(x => x.name.value !== 'id').length === 0
+      if (idsOnly) {
         return { id: offer.listingId }
       } else {
-        return search.Listing.get(offer.listingId)
+        const row = await db.Listing.findByPk(offer.listingId)
+        return row.data
```
sparrowDom: We could handle the case when a wrong/not existing `listingId` is queried and return null instead of throwing an exception.

origin-discovery/src/apollo/resolvers.js
```diff
@@ -57,28 +58,40 @@ const resolvers = {
...
+      if (args.listingId) {
+        clause.listingId = args.listingId
+      }
+      if (args.buyerAddress) {
+        clause.buyerAddress = args.buyerAddress.toLowerCase()
+      }
+      if (args.sellerAddress) {
+        clause.sellerAddress = args.sellerAddress.toLowerCase()
+      }
+      const rows = await db.Offer.findAll({ where: clause })
```
sparrowDom: This might open the possibility to get all offers in the DB by not specifying `listingId`, `buyerAddress` or `sellerAddress`.  Should we add an explicit check that at least one of the filters is specified (I know we unfortunately can't do that in Apollo schema). Perhaps something like: 
```
if (Object.keys(clause) === 0)
    throw 'At least one of the following filters needs to be specified: listingId, buyerAddress or sellerAddress'
const rows = await db.Offer.findAll({ where: clause })
```

origin-discovery/src/apollo/resolvers.js
franckc: yes ! good catch :)

origin-discovery/src/apollo/resolvers.js
franckc: The problem is that search.User.get(args.walletAddress) throws an exception if it does not find the address.... I could have a try/catch statement but since currently the DApp is not query user data from graphQL I was thinking ok to only return the wallet address. And I plan on fixing the issue in the listener causing some of those user to not get indexed (I'm actually thinking putting the data in a new "user" table in the DB rather than elastic since I don't think we'll nee full text search on users ?

origin-discovery/src/apollo/resolvers.js
franckc: We could though if the listing is not found, it does indicate there is a problem... it should not be possible that an offer exists but there is no associated listing in the DB.
I guess we should discuss about our overall GraphQL error handling. If the back-end errors out for a valid reason (ex: data should be there but is not), do we want to mask that by returning no data or do we want to surface the exception ? (I'm leaning towards the latter, to help us investigating issues).

origin-discovery/src/apollo/resolvers.js
sparrowDom: Postgres has [some full text search capabilities](https://www.postgresql.org/docs/9.5/textsearch.html) in case we need them. We could copy `firstName` and `lastName` to a `name` full text enabled field. And querying should work there. Even if user has multiple first/last names and we query only by part (one of the multiple names) of the first/last name.

origin-discovery/src/apollo/resolvers.js
sparrowDom: Ah yes, good point it is tied to an offer. Yes I agree, an exception is a good way to respond here. Perhaps a custom exception message? Something like `Listing of the offer with id: ${offerId} could not be found!` instead of "Can not read property `data` of undefined"

origin-discovery/src/apollo/resolvers.js
franckc: Agreed, makes sense. Let's not worry too much about it now - looks like we have plenty of good option if needed in future... :)

origin-discovery/src/apollo/resolvers.js
franckc: Yes, good idea. Added.

franckc:
@sparrowDom Thanks for the review, I addressed your comments.
I added some unit tests and also tested manually.
This is now ready for final review.

Note that I did a bit of cleanup/simplification on the hidden/featured logic (using import rather than passing around the hiddenIds and featuredIds everywhere). You may want to wait until I merge this in before writing additional unit tests.

jordajm:
@franckc - as I dug in further on making editable listings stuff work with discovery server, I realized that we'll probably need to implement another boolean `opts` property for `getListings(opts)` called `withBlockInfo` or something like that so we can return the `blockNumber` and `logIndex` of the offer and then pass that to the `getListing()` call like `getListing(listingId, blockNumber, logIndex)`. I have a [POC for this here](https://github.com/OriginProtocol/origin/pull/882/commits/156459f58bd3713cf40beba0be939f85b3e742c7) - it's not quite working yet but I'll debug it tomorrow - just wanted to give you a heads-up in case you wanted to include it in this PR or give me feedback / different ideas for how to do this.

origin-discovery/src/apollo/db.js
```diff
@@ -1,68 +1,91 @@
...
+ * Helper function. Returns a listing object compatible with the GraphQL Listing schema.
+ * @param {Object} row - Row read from DB listing table.
+ * @return {Object}
+ * @private
  */
-async function getListings (listingIds, hiddenIds = [], featuredIds = []) {
+function _makeListing (row) {
+  // Compute display.
+  let display = 'normal'
+  if (listingMetadata.hiddenIds.includes(row.id)) {
```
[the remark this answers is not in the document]
franckc: Good idea. Done.

franckc:
@jordajm Thanks for the heads up, appreciate it :) Yes, I think adding a withBlockInfo option to getListings and supporting blockInfo as an optional input to getListing makes sense. We'll implement those changes in the backend. Perhaps we could define a **blockInfo** object with 2 fields blockNum and logIndex and use that in the getListings output and as optional input to getListing ?

origin-discovery/src/apollo/listing-metadata.js
```diff
@@ -29,13 +27,15 @@ class ListingMetadata {
...
+        this.featuredIds = await this.readListingsFromUrl(this.featuredListingsUrl)
       } catch (e) {
         console.error('Could not update hidden/featured listings ', e)
       }
     }
   }
 }
 
-module.exports = ListingMetadata
+const metadata = new ListingMetadata()
```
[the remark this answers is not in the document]
sparrowDom: Not static class ... singleton 🤓

PR #904, opened 2018-11-11
Node Version

micahalcorn:
Is there a more specific version that should be recommended?


tomlinton:
No specific version at this stage, they should all be fine.

PR #911, opened 2018-11-12
Notification Counterparty Label

micahalcorn:
This repairs the reference to "buyer" or "seller" in derived notifications when they are displayed in the DApp.


sparrowDom:
LGTM

PR #921, opened 2018-11-13
force timeout

kyriediculous:
Seems like origin-js is using a gateway?

If so the issue is related to this open issue: ipfs/go-ipfs#4157
"gateway hangs instead of returning 404"

So until that issue is closed I propose hacking around it by creating a manual timeout and racing promises.

```

  async loadFile(ipfsHash) {
    let timeout = new Promise((resolve, reject) => {
      const ms = 7000
      setTimeout(() => {
        reject(new Error("Timed out after " + ms + " ms"))
      }, ms)
    })
    try {
      return await Promise.race([timeout, fetch(this.gatewayUrlForHash(ipfsHash))])
    } catch (error) {
      throw new Error('Failure to get IPFS file: ' + error.message)
    }
  }

```


kyriediculous:
Sorry it took me a while to get back to this, passes now.

PR #923, opened 2018-11-13
Add more docker volume mappings

sparrowDom:
### Description:
- add contracts build dir to `event-listener` container. Enabling event listener to interact with local web3
- add some more volumes in `origin-js` developers might want to interact with from the host machine

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,10 +53,6 @@ services:
       # written to the build files
       - ./origin-contracts/build:/app/origin-contracts/build
       - ./origin-js/scripts:/app/origin-js/scripts
-      - ./origin-js/src:/app/origin-js/src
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -52,6 +52,11 @@ services:
       # Addresses for deployment of contracts to local blockchain will get
       # written to the build files
       - ./origin-contracts/build:/app/origin-contracts/build
+      - ./origin-js/scripts:/app/origin-js/scripts
+      - ./origin-js/src:/app/origin-js/src
```
tomlinton: Only the `scripts/` directory matters here as the `origin-js` code doesn't compile at all in this container, it just runs some scripts

sparrowDom: Aha I see tnx

PR #926, opened 2018-11-14
Update nginx ingress config

tomlinton:

### Checklist:

- ~~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/origin-docs)~~

### Description:

- Move values out for nginx-ingress Helm chart into separate files
- Add basic rate limits to all of the ingresses for DDOS protection, these values might need to be tweaked over time. I just guessed at reasonable ones.

### DIFF
--- a/deployment/kubernetes/charts/origin/templates/dapp.ingress.yaml
+++ b/deployment/kubernetes/charts/origin/templates/dapp.ingress.yaml
@@ -12,6 +12,7 @@ metadata:
     kubernetes.io/tls-acme: "true"
     certmanager.k8s.io/cluster-issuer: {{ .Values.clusterIssuer }}
     nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
+    nginx.ingress.kubernetes.io/limit-rps: "2"
 spec:
   tls:
     - secretName: {{ template "dapp.host" . }}
--- a/deployment/kubernetes/charts/origin/templates/discovery.ingress.yaml
+++ b/deployment/kubernetes/charts/origin/templates/discovery.ingress.yaml
@@ -13,6 +13,7 @@ metadata:
     certmanager.k8s.io/cluster-issuer: {{ .Values.clusterIssuer }}
     nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
     nginx.ingress.kubernetes.io/rewrite-target: /graphql
+    nginx.ingress.kubernetes.io/limit-rps: "5"
 spec:
   tls:
     - secretName: {{ template "discovery.host" . }}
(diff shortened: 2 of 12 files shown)


deployment/kubernetes/charts/origin/templates/dapp.ingress.yaml
```diff
@@ -12,6 +12,7 @@ metadata:
     kubernetes.io/tls-acme: "true"
     certmanager.k8s.io/cluster-issuer: {{ .Values.clusterIssuer }}
     nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
+    nginx.ingress.kubernetes.io/limit-rps: "2"
```
franckc: Does that include serving static content ? If yes, that seems low. When I reload the Dapp in my browser, I see in the order of 30 requests to get static content (css, images).

deployment/kubernetes/charts/origin/templates/discovery.ingress.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
     certmanager.k8s.io/cluster-issuer: {{ .Values.clusterIssuer }}
     nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
     nginx.ingress.kubernetes.io/rewrite-target: /graphql
+    nginx.ingress.kubernetes.io/limit-rps: "5"
```
franckc: That's low. On the browse/search page the DApp will issue 1 request per listing card and we have 12 cards. Based on that, I'd set this to something in the order of 24.

deployment/kubernetes/charts/origin/templates/eth.ingress.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
     certmanager.k8s.io/cluster-issuer: {{ .Values.clusterIssuer }}
     nginx.ingress.kubernetes.io/rewrite-target: /
     nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
+    nginx.ingress.kubernetes.io/limit-rps: "2"
```
franckc: That seems low. The DApp when not using discovery server makes more than 1 web3 call per listing card on browse/search page. I'd set this to something in the order of 24.

deployment/kubernetes/charts/origin/templates/ipfs.ingress.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
     nginx.ingress.kubernetes.io/enable-cors: "true"
     nginx.ingress.kubernetes.io/cors-allow-methods: "GET, PUT, POST"
     nginx.ingress.kubernetes.io/cors-allow-origin: "*"
+    nginx.ingress.kubernetes.io/limit-rps: "20"
```
franckc: Seems a bit low. If we have 2 concurrent users on browse page, that's 2*12 listings = 24 listings to fetch. Each listing has its JSON blob hosted on ipfs and then at least 1 image. So that's about 2 * 12 * 2 = 48.

tomlinton: These are per IP address, but you are right, maybe I'll make it 30.

deployment/kubernetes/charts/origin/templates/discovery.ingress.yaml
tomlinton: Thanks, I didn't consider the retrieving listings bit.

PR #939, opened 2018-11-15
[Elastic] Fixes to index tool for production

franckc:

### Checklist:

### Description:
Yesterday I used the tool in production and faced a couple of minor issues
1. I needed to run the tool from a production container but given that the source code for the tool is under development, it was not packaged and deployed to container. To fix this, I'm moving this tool to the origin-discovery directory so that it gets deployed to the discovery and listener container which is where we are most likely to run the tool.
2. Env var ELASTICSEARCH_HOST was not used.
3. Import style for util methods was causing error.


origin-discovery/devops/es-cli.js
```diff
@@ -1,16 +1,35 @@
...
 const http = require('http')
 const https = require('https')
 const readline = require('readline')
-const arrayFunctions = require('./../../../origin-js/src/utils/arrayFunctions')
+const { groupBy, mapValues } = require('./../../origin-js/src/utils/arrayFunctions.js')
 
 // Configuration
-const host = 'localhost'
-const port = '9200'
+var host = 'localhost'
```
tomlinton: I think the linter is going to complain about `var` here. We have a universal linter setup now, you can run `npm run lint` from root or `lerna run lint --scope origin-discovery` for just linting discovery.

I'm sure you'll lern this. 😆

franckc: oh sorry. just saw this.
ah looks like linter did not catch me because that directory is not included ! :)
I'll fix the linter config and that file. Thanks for catching this - Mr Linter ;)

Issue #895, opened 2018-11-08
Listing Image Upload Improvements


micahalcorn:
The current state of the listing image upload feature leaves the cropping to the user and prevents multiple selection of files. We should re-enable the selection of multiple files to encourage multiple images and save the user from the cognitive load and extra clicks associated with uploading them individually. As a result, it won't make sense to require cropping each image, so they should be automatically scaled and cropped. We should also provide some hint text indicating the recommended dimensions so that the user is not surprised when the images are cropped to a 4:3 ratio.

micahalcorn:
Resolved by #945 ✅

PR #945, opened 2018-11-16
 Create-listing image upload UX improvements

jordajm:
### Description:
- Allow multiple images to be uploaded at once
- Automagically crop images to 4:3 (crop area is centered vertically and horizontally in image)
- Add a new image overlay button beside the "delete image" button that allows the user to re-crop the original image
- Issue: https://github.com/OriginProtocol/origin/issues/895

#### TODO
- Go find the help text stuff that used to exist for telling the user how to select multiple images and put that back in here
- Merge this code into the editable listings branch and make this work with edit mode

### Checklist:

------------------------------------------------------

screenshot 2018-11-15 23 36 32


micahalcorn:
Thanks @jordajm 🔥 I agree that we should add back the instructions for multiple file selection since we have the other text about aspect ratio, etc. I noticed a couple of things:

1. The initial crop doesn't quite make use of the entire height or width of the image. In other words, I can manually crop and get "more image" with the same aspect ratio. Is this intentional? Or let me know if you can't reproduce with your test image(s).
2. If I select multiple images and then go to adjust the crop on the first one, it gets removed from the gallery. So if I "Continue" the crop, it gets appended to the end. And if I "Cancel", it remains removed. Is it possible for us to maintain the state of the selected photo's inclusion and placement within the sorted gallery?

jordajm:
@micahalcorn thanks for the good feedback. 

I added back the multiple image upload help text.

I made the improvement you suggested for maintaining the order of the images and not deleting the image if you click the "cancel" button on the cropper modal.

I looked into the issue you mentioned about the auto-cropper math not maximizing the crop area, but I couldn't see any issues with it after testing with a few different images. For landscape images (including the bike one you sent me), it sets the crop area to the full height of the image, then sets the width to a 4:3 ratio based on the height, and centers it in the image. Portrait images are the inverse and the crop area is centered vertically. I'd be happy to look into it further if you still think this is an issue.

Issue #897, opened 2018-11-08
Message Image Upload Improvements


micahalcorn:
We should use `react-image-crop` with the image upload functionality in the conversation UI to automatically resize images. Let's not expose the crop feature in this case and let's not enforce a particular aspect ratio. Users should be able to upload images that we do not edit other than reducing the overall size to fall under our maximum threshold.

Related: https://github.com/OriginProtocol/origin/issues/593 & https://github.com/OriginProtocol/origin/issues/817

micahalcorn:
Resolved by #957 ✅

PR #957, opened 2018-11-19
Untested messaging image resizing

jordajm:
### Description:

@micahalcorn I haven't tested this because I'd need to spin up box and I don't have time tonight, but I think the resizing you describe [here](https://github.com/OriginProtocol/origin/issues/897) could be done something like this... feel free to test it out locally or LMK and I can try to spin up box tomorrow.


micahalcorn:
Thanks @jordajm I'll do it in the morning 👍

PR #962, opened 2018-11-19
[Discovery] Support for editable listings

franckc:

### Description:
This PR updates the discovery stack to take into account listing versioning, which is necessary for editable listings. Here are the changes included:
 - listing graphQL query now takes an optional blockInfo argument
 - offer and listing data returned by disco server now includes blockInfo (this is not being used by origin-js but is useful to inspect the data when issuing manual query).
 - when fetching a listing in the context of an offer, use the offer's blockInfo to fetch a specific version of the listing.
 - updated event-listener to load specific version of the listing so as to preserve history when re-indexing data.

### Checklist:


jordajm:
👍 LGTM @franckc - I'm not super familiar with this code and I haven't run this locally but it does look like it covers the bases

origin-discovery/src/apollo/db.js
```diff
@@ -161,6 +162,10 @@ async function getListing (listingId, blockInfo = null) {
 function _makeOffer (row) {
   return {
     id: row.id,
+    blockInfo: {
+      blockNumber: row.data.events[0].blockNumber,
```
sparrowDom: This should aways work, just to improve error readability in case it faults I would add: 
```
if (row.data.events.length === 0 || row.data.events[0].event !== 'OfferCreated')
  throw new Error('Can not find OfferCreated event')

return {
    id: row.id,
    blockInfo: {
    ...
```

franckc: Defensive coding is always a good idea. Thanks for the suggestion. Added.

sparrowDom:
@franckc we don't need to do any changes to support multi unit listings right? 

We should be able to track when offers have been made and if a user has changed the `unitsTotal` property, to rebuild the whole history of units available / purchases made if needed?

franckc:
@sparrowDom re any change required for multi-units, I'm not sure. I need to review cuong's PR #977 to understand better the details of multi-units implementation... :)

sparrowDom:
Thanks @franckc. Was just asking because I didn't know what status to write in `Sprint Check-In` paragraph of the [Engineering Weekly doc ](https://docs.google.com/document/d/1aRcAk_rEjRgd1BppzxZJK9RXfDkbuwKKH8nPQk7FfaU/edit#) next to the `Support editable/multi-units listings`

PR #963, opened 2018-11-19
Add src and tests folders to origin-js container

sparrowDom:
### Description:

Add `src` and `test` volume maps for `origin-js` container so `origin-js` test can be ran inside the container

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,6 +53,8 @@ services:
       # written to the build files
       - ./origin-contracts/build:/app/origin-contracts/build
       - ./origin-js/scripts:/app/origin-js/scripts
+      - ./origin-js/tests:/app/origin-js/tests
+      - ./origin-js/src:/app/origin-js/src
 
   origin-discovery:
     container_name: origin-discovery
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,7 +53,7 @@ services:
       # written to the build files
       - ./origin-contracts/build:/app/origin-contracts/build
       - ./origin-js/scripts:/app/origin-js/scripts
-      - ./origin-js/tests:/app/origin-js/tests
+      - ./origin-js/test:/app/origin-js/test
       - ./origin-js/src:/app/origin-js/src
 
   origin-discovery:


sparrowDom:
@tomlinton this solved the problem for me. Feel free to reject PR if there is a better solution 🤷‍♂️

tomlinton:
Is there some reason for running the tests inside the container? I just run them outside of it. I'm hoping to move away from running an `origin-js` container at all and instead use dedicated IPFS and Ganache containers so we can remove these extra dependencies from `origin-js`.

sparrowDom:
I don't get very far running tests outside of `origin-js` container: 
```
MacBook-Pro-4:origin-js domen-privat$ pwd
/Users/domen-privat/projects/OriginProtocol/origin/origin-js
MacBook-Pro-4:origin-js domen-privat$ npm test

> origin@0.8.6 test /Users/domen-privat/projects/OriginProtocol/origin/origin-js
> node scripts/test.js

internal/modules/cjs/loader.js:596
    throw err;
    ^

Error: Cannot find module './build/Release/scrypt'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:594:15)
    at Function.Module._load (internal/modules/cjs/loader.js:520:25)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/Users/domen-privat/projects/OriginProtocol/origin/origin-js/node_modules/ganache-core/node_modules/scrypt/index.js:3:20)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
npm ERR! Test failed.  See above for more details.
```

that is outside the container in`origin-box` project.  

In a non box setup with linter I also had some problems yesterday. After resolving the problem with IPFS (adding `0.0.0.0    localhost` in `/etc/hosts`) some node_modules were not installed after `linter boostrap` (I think it was `chai`). And I didn't continue resolving problems after that. 

In what environment do you run the tests?

tomlinton:
I run them outside (or just push and let TravisCI run them), [these steps](https://github.com/OriginProtocol/origin/blob/master/.travis.yml#L81-L86) from root work because TravisCI runs them every time (sometimes there is the odd timeout but it is generally pretty reliable).

If you do `lerna bootstrap` from root and then `npm install` from inside the `origin-js` directory it'll make a mess of the links lerna created. Perhaps that is what happened? `chai` is a dev dependency at the root level so it should be installed by `lerna bootstrap`.

I used to have a separate test only Dockerfile which I can bring back if that still doesn't work.

sparrowDom:
I had a separate clone of origin github repo only using lerna, so that `lerna bootstrap` would be ran separately from the repo where origin-box is running. I was running it using commands: 
```
lerna clean
lerna boostrap
lerna run test --scope origin --stream
```
I wasn't aware that instead of `lerna clean && lerna boostrap` this should be ran: 
```
- npm install --ignore-scripts
- npm run bootstrap -- --scope origin-contracts
- npm run bootstrap -- --scope origin
```

Also this is not ideal if you already have the box running, since it requires a secondary repository checkout and manual file syncing between `origin-box` repo and lerna repo. I think it would be useful for all the box users to have the ability to run the tests in their environment. A docker container dedicated to tests that you propose sounds great to me. Can I help with that effort?

tomlinton:
That second lot of commands don't need to be run in addition. The whole process should just be:

```
npm install
lerna run test --scope origin --stream
```

That works fine for me from a fresh clone. If that doesn't work from a fresh repo there is an error/issue somewhere that needs to be resolved (regardless of whether we run the tests in the box or not, those commands should always work).

The second lot of commands are what the CI server does and just show that the process seems to work. They are a faster version of the above that only install the dependencies necessary to make the tests for origin-js work rather than installing all dependencies.

sparrowDom:
Nice @tomlinton I can confirm that running 
```
npm install
lerna run test --scope origin --stream
```
from the root folder of `origin-js` container works for me. The 2 things I need in order to get that running is: 
 - changes from this PR so that `src` and `test` folders sync with host machine
 - changing the `serve.js` script so that it doesn't start services and doesn't exit effectively shutting down `origin-js` container: 
```
const start = async () => {
  if (!noGanache) {
    console.log(chalk`\n{bold.hex('#1a82ff') ⬢  Starting Local Blockchain }\n`)
    //await startGanache()
  }
  console.log(chalk`\n{bold.hex('#26d198') ⬢  Deploying Smart Contracts }\n`)
  //await deployContracts()
  console.log(chalk`\n{bold.hex('#6e3bea') ⬢  Starting Local IPFS }\n`)
  //await startIpfs()
  new Promise(resolve => setTimeout(resolve, 1000000000));
}
```

tomlinton:
That should work from the root of the `origin` monorepo though, no container needed.

sparrowDom:
On my OSX the install goes through normally, but when running I get this error: 
screenshot 2018-11-21 09 33 20

Using `nvm` I have the same `node` and `npm` versions as inside the container.  Attempting to installing `scrypt` globally didn't help with the above error: `npm uninstall -g scrypt`. 

If this is only on my machine, then it is not a problem, I will manage with minor modifications inside the `origin-js` container. If others will have the same problems then we might need a more general solution. Let me ask in engineering to see how widespread this problem is

sparrowDom:
@tomlinton yeah I have tried a couple of things from the link you have posted before. Let me dedicate some more time to this, to try to get it running.

tomlinton:
I'll close this for now @sparrowDom, let me know if it is still a problem!

sparrowDom:
@tomlinton thanks for cleaning this up

Issue #877, opened 2018-11-03
Notifications Analytics


micahalcorn:
Here is a non-exhaustive list of the notification-related events that we should probably track:

- [x] [`push`](https://developer.mozilla.org/en-US/docs/Web/API/PushEvent): the [service worker](https://developer.mozilla.org/en-US/docs/Web/API/PushEvent) receives data from the server
- [ ] [`pushsubscriptionchange`](https://developer.mozilla.org/en-US/docs/Web/Events/pushsubscriptionchange): the browser updates the push subscription (not currently handled 😬)
- [x] [`notificationclick`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/onclick): the user triggers the default action
- [x] [existing window/tab is focused](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/public/sw.js#L39)
- [x] [new window/tab is opened](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/public/sw.js#L41)
- [x] [`notificationclose`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/onclose): the user dismisses the notification
- [x] [`onerror`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/onerror): something is not happy
- [x] [`onshow`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/onshow): a notification is displayed
- [x] [`serviceWorker` is not supported](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/utils/notifications.js#L52)
- [x] [`pushManager` is not supported](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/utils/notifications.js#L56)
- [x] [service worker is registered](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/utils/notifications.js#L62)
- [x] [subscription is created](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/utils/notifications.js#L20-L24)
- [x] [permission is granted](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/utils/notifications.js#L92)
- [x] [permission is not granted](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/utils/notifications.js#L89)
- [x] [prompt is dismissed/declined](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/components/onboarding.js#L201)
[...]

PR #969, opened 2018-11-20
Notification Analytics

DanielVF:
Adds the notifications analytics from issue #877

@micahalcorn, you will want to test that GA is actually receiving these events.

### Description:

- Adds an analytics service module to the DAPP
- Adds an analytics helper to the service worker (can't use the one from the dapp)
- Adds analytics events as described in #877
- Console logs all analytics events for help in testing. (Can be removed before this PR is merged)
- Fixes a  bug in the notification server

### Checklist:


origin-notifications/src/app.js
```diff
@@ -108,11 +108,11 @@ app.get('/', async (req, res) => {
 app.post('/', async(req, res) => {
   const { account, endpoint } = req.body
   // cannot upsert because neither is unique
-  const existing = await PushSubscription.find({
+  const existing = await PushSubscription.findAll({
```
micahalcorn: @DanielVF why `findAll` here instead of `find`? There shouldn't be more than one, should there? And if there is, it doesn't affect this request. I assumed that `find` would be marginally faster, but just curious.

origin-dapp/src/components/onboarding.js
```diff
@@ -95,6 +96,7 @@ class Onboarding extends Component {
 
     // To Do: handle incoming messages when no Origin Messaging Private Key is available
     origin.messaging.events.on('emsg', obj => {
+      analytics.event('Notifications', 'ErrorNoDecryption')
```
micahalcorn: @DanielVF this is going to produce a lot of noise. 🙉 I'm not sure that we care about this, but I guess we can filter it out?

origin-notifications/src/app.js
DanielVF: There isn't a `find` in the Sequelize API we are using, so I switched to `findAll`. (http://docs.sequelizejs.com/class/lib/sequelize.js~Sequelize.html). (Turns out there is a `findOne` though.)

Although there should only every be one, in the event of a race condition, multiple identical ones could get inserted and multiple returned. I guess this is a tiny, tiny bit more clear that more than one could be returned. I'm pretty neutral though - you are welcome to change it to `findOne`.

origin-dapp/src/components/onboarding.js
DanielVF: Yeah, you can comment it out. I was just putting analytics on things I saw happen.

origin-notifications/src/app.js
micahalcorn: I guess no method is a decent reason not to use it. 😕 I must have been reading the wrong docs.

origin-dapp/public/swAnalytics.js
```diff
@@ -0,0 +1,89 @@
+// Based off
+// https://developers.google.com/web/ilt/pwa/integrating-analytics
+
+class Analytics {
+    constructor(){
```
micahalcorn: @DanielVF is this file's indentation style intentional or should I update it to match the others?

origin-dapp/public/swAnalytics.js
```diff
@@ -0,0 +1,89 @@
...
+            // Event Category
+            ec: eventCategory,
+            // Event Action
+            ea: eventAction,
+            // Event Label
+            el: eventLabel
+            };
+        
+            // Format hit data into URI
+            var payloadString = Object.keys(payloadData)
```
micahalcorn: @DanielVF any reason not to use `const` here? I did in the `sw.js`, which I doubt gets transpiled.

origin-dapp/public/swAnalytics.js
DanielVF: You should update it - sorry.

origin-dapp/public/swAnalytics.js
DanielVF: We can use it.

PR #978, opened 2018-11-21
Misc fixes for create housing listing, photo-picker, etc

jordajm:
### Description:

Just a few fixes for random stuff that has been broken by merges this week... 
- Housing schema was missing a message ID and prevent user from creating a listing
- Photo picker needed to have the conditional wrapper around cropper modal removed so the `componentDidUpdate()` would fire on render
- The `react-big-calendar` lib updated the way they do localization so that needed a small tweak


jordajm:
Thanks @Natasha08 - I'll do another PR with fixes for these - good finds 👍 🤘

Issue #372, opened 2018-08-09
Potentially non-localizable string


wanderingstan:
For localization, we cannot be assembling sentences from component parts, as other languages' grammar/segmenting/ordering may not be like English, and they may have extra requirements around gender/case changes.

This code is used to create sentences like "In order to create a listing, you must install MetaMask":
 
```
    <div>In order to {props.web3Intent}, you must install MetaMask.</div>
```
([link](https://github.com/OriginProtocol/origin-dapp/blob/master/src/components/web3-provider.js#L41))

This needs to be structured into complete sentences that can be translated whole.

micahalcorn:
No longer relevant in DApp2 🌐

PR #982, opened 2018-11-22
Require Messaging

micahalcorn:
This isn't particularly pretty, but should serve as a stopgap for preventing listing/offer creation at the DApp level without messaging enabled. I reused the `web3Intent` variable, which is arguably different than "intent to do something that requires messaging", for now. All of this will need to be refactored very soon in conjunction with a better user activation scheme. But the business team is reporting that users are continuing to transact on Mainnet without first enabling messaging, so this should be deployed sooner rather than later.

I also moved the web3 check from the "Add a Listing" link to the creation component so that direct links don't circumvent the check.

The changes here continue to kick the can on [this translation issue](https://github.com/OriginProtocol/origin/issues/372) because it would be best be refactored along with the broader activation changes. Handling it now would result in a lot of new strings for translators to work on only to have them all be changed later this quarter.


origin-dapp/src/components/listing-create.js
```diff
@@ -137,6 +145,9 @@ class ListingCreate extends Component {
         console.error(`Error fetching contract or IPFS info for listing: ${this.props.listingId}`)
         console.error(error)
       }
+    } else {
+      this.props.history.push('/')
```
sparrowDom: The problem might be in this line of code. When I have messaging enabled and try to create a new listing I just get redirected to home page without any message.

micahalcorn: Thanks 😬 I'll look into it later today.

micahalcorn: Good catch @sparrowDom 🤦‍♂️

PR #987, opened 2018-11-23
Hotfix: Ganache Mining Interval

micahalcorn:
_This is intentionally pointed at `stable`._

#985 updates the `blockTime` option in the Ganache server to use a four-second mining interval rather than "instant". Here I am cherry-picking that commit and merging it to `stable` so that the blockchain transactions dropdown will work using a local environment for product demos.


franckc:
@micahalcorn Any objection if I merge this into master and staging ?

franckc:
Nevermind. I see you have PR #985 to take care of merging this into master.

PR #997, opened 2018-11-26
Fix TravisCI build status

tyleryasaka:


wanderingstan:
Do we need to change here too?  
https://github.com/OriginProtocol/origin/blob/master/origin-js/README.md

tyleryasaka:
@wanderingstan ah yes, fixed ✅

PR #1003, opened 2018-11-27
Use schemas as source of truth for fractional listing types

jordajm:
### Description:
- in DApp - Adds `"listingType"` enum to schemas
- in DApp - Adds `"slotIncrementMinutes"` to fractional schemas
- in origin-js - updates listing schema to match dapp schemas
- in origin-js - adds support for `slots` in the offer model

- NOTE:  I'm not sure this is ready to merge quite yet, but wanted to get it out there for review. I'll run through it again tomorrow and run/update the tests, etc.

### Checklist:


micahalcorn:
@wanderingstan @franckc should I roll my category restructuring into this same PR? I don't know that it makes sense to merge `listingType` decisions here that won't necessarily be accurate or relevant.

jordajm:
@micahalcorn has that work been pushed to a branch? Can't find it 👀

micahalcorn:
@jordajm no, sorry 😐 I’m waiting for confirmation from M&J. Right now, it’s just a spreadsheet.

origin-dapp/public/schemas/announcements.json
```diff
@@ -3,6 +3,14 @@
   "type": "object",
   "required": ["name","category","description","price"],
   "properties": {
+    "listingType": {
```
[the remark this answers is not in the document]
micahalcorn: @franckc Does this mean that all listings of schema "services", for example, will have to be fractional? It's looking like we're going to be able to have a one-to-one mapping between schema and `listingType` based on our consensus (pre-approval from J&M). But that is not likely to be true in the future.

origin-contracts/package-lock.json
```diff
@@ -10,16 +10,16 @@
...
-				"commander": "^2.8.1",
-				"convert-source-map": "^1.1.0",
-				"fs-readdir-recursive": "^1.1.0",
-				"glob": "^7.0.0",
-				"lodash": "^4.17.10",
-				"mkdirp": "^0.5.1",
-				"output-file-sync": "^2.0.0",
-				"slash": "^2.0.0",
-				"source-map": "^0.5.0"
+				"chokidar": "2.0.4",
```
micahalcorn: Any reason to commit the package-lock in this PR?

origin-dapp/public/schemas/housing.json
```diff
@@ -3,6 +3,13 @@
   "type": "object",
   "required": ["name","category","description"],
   "properties": {
+    "listingType": {
+      "const": "fractional"
```
micahalcorn: Will this "take effect" or have any consequences if deployed to production before we "wire on" the calendar, etc?

I'm fairly certain that I should include my changes to the categories before this is merged. Several of these will be inaccurate or removed entirely.

origin-contracts/package-lock.json
jordajm: Don't think so - I'll pull it out

origin-dapp/public/schemas/announcements.json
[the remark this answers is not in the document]
jordajm: Yeah that's a good point @micahalcorn - my original thinking was to make it an `enum` with a default so a dapp developer could override it if they want, but after @franckc 's comment, I figured it would be just as easy (and maybe cleaner) for them to change a `const` than override an enum's default... maybe it's about the same really.

origin-dapp/public/schemas/housing.json
jordajm: Good point @micahalcorn - I'll make a change so that this will still respect the `ENABLE_FRACTIONAL` environment variable, but we can have this in the schema without breaking anything. 

If your changes are ready to implement, I'm all for including them here. IMO, not a big deal to merge this and do them separately though if the final approval from J&M is going to take a while. If we were talking about origin-js schema changes that required a version bump, ofc I'd feel differently, but these are pretty low effort to change.

micahalcorn: I'll point a PR to here so that you can review before I hijack your PR.

origin-dapp/public/schemas/announcements.json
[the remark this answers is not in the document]
wanderingstan: Late to the party, but I agree with having a single listingType per schema. 

It seems clear we'll eventually need to go to having one jsonschema per listing type in the future, e.g. one for bikes with size/color/make, one for homesharing with bedrooms/baths. As the jsonschema files get more specific, we certainly only want one 'listingType' per schema.

franckc:
@wanderingstan @jordajm When we move to u sing @micahalcorn's [new taxonomy](https://docs.google.com/spreadsheets/d/1K7ZGnEqLJxPM9MI27sjr4m_U9yyA5DFdylmzOxJq04M/edit#gid=1657275235), how should we handle legacy listings that have already been created but are in a type that is being deprecated (ex: Housing) ? If we completely delete that old Housing schema, this is I think going to cause errors in the DApp in couple of areas:
 1. Listing details page: one issue for example is we'll fail loading the translated string. See [here](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/utils/listing.js#L132).
 2. Listing editing: since the schema for the housing type won't be there, I think the DApp will fail displaying that page.
Haven't looked at details but we will probably have issues if any legacy listing was created in a category we are planning to deprecate.

We could perhaps keep the old schema in a "deprecated" directory ?
Any thoughts on better options ? Happy to discuss in a call...

Note: I can easily query the DB to get a list of type/cat for existing listings...

wanderingstan:
Good point @franckc . Is the time we need to talk about versioning??

origin-dapp/src/utils/listing.js
```diff
@@ -72,15 +92,16 @@ export function dappFormDataToOriginListing(formData) {
  * @return {object} DApp compatible listing object.
  */
 export function originToDAppListing(originListing) {
+console.log('================================== originListing: ', originListing)
```
franckc: remove

origin-dapp/src/utils/listing.js
```diff
@@ -62,6 +61,27 @@ export function dappFormDataToOriginListing(formData) {
   return listingData
 }
 
+function getCategory(category) {
```
franckc: nit: for both methods, add a comment explaining why we are doing this (for backward compatibility purposes to support legacy listings created in now deprecated cats/subcats).

origin-dapp/src/utils/listing.js
```diff
@@ -62,6 +61,27 @@ export function dappFormDataToOriginListing(formData) {
...
+    case 'schema.tickets':
+      return 'forSale'
+    default:
+      return category
+  }
+}
+
+function getSubCategory(subCat) {
+  switch(subCat) {
+    case 'schema.housing.vacationRentals':
```
franckc: I think there are a few more subcats that have legacy listings and are being deprecated.
See this xsheet with data I queried from prod (look at 3rd tab): https://docs.google.com/spreadsheets/d/1K7ZGnEqLJxPM9MI27sjr4m_U9yyA5DFdylmzOxJq04M/edit#gid=1600462284

jordajm:
Closing in favor of #1026
All the important stuffs from this PR will be implemented there.

PR #1012, opened 2018-11-28
remove expiry field from listing

sparrowDom:
### Description:
Removes `expiry` field from listing model. 

### Checklist:

- ~[ ] Wrap any new text/strings for translation~
- ~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/origin-docs)~


sparrowDom:
@franckc @jordajm @wanderingstan any objections to removing this field?

cuongdo:
What effect will this have on existing prod listings that will still have the `expiry` field? Will they continue to work, or will they fail to load because of schema validation?

sparrowDom:
@cuongdo good thinking. Seems that the listing validation only checks for presence of fields and field types and doesn't fail when there is an unexpected field. 

Just to be sure have just tested this with creating default listings with `origin-js` fixtures on master  branch(with `expiry` field) and then checked if they load/display on this branch and they do.

jordajm:
What's the motivation for removing the field? I could see it potentially being useful - for example, if a dapp is selling tickets to events, they might choose to set the expiry to the date of the event and then set up their search engine to ignore listings with expiry dates in the past.

cuongdo:
@jordajm brings up a good point. Also, I think that we want listings expiring at some point to ensure that each version of the marketplace contract has a finite lifespan. We don't have any mechanism for porting listings from an older version of the marketplace contract to a newer one, and it'll become cost-prohibitive to support contracts that are too old.

sparrowDom:
@jordajm @cuongdo those are both good points.

The motivation for removing it was solely that it isn't used by anything / populated at listing create time at the moment. If we are ok with having partly implemented feature, that we are pretty sure that it is going to be used in the future, we can cancel this PR. 

The downside is that there is slight engineering overhead supporting expiry in the codebase.

franckc:
Good points on both sides.
Personally I'd vote for cleaning this up and implementing it across our stack when we are ready.
If we leave the code in, then let's add comments so that people are not falsely under the impression that this field is currently supported (it is for future).
Let's bring it up and make decision during the eng weekly.

sparrowDom:
@franckc when you have a minute this is ready to be stamped and merged

franckc:
@sparrowDom Looks like there are a few conflicts with master. Could you hep resolving them then I'll merge. Thank you :)

sparrowDom:
FIxed, just waiting for the tests to pass.

sparrowDom:
@franckc all green here

Issue #678, opened 2018-10-03
Listing Types Not Localized


micahalcorn:
This probably isn't blocking, but should likely be prioritized over other localization quirks. Did I break this?

<img width="1440" alt="screen shot 2018-10-03 at 4 51 15 pm" src="https://user-images.githubusercontent.com/273937/46445694-1388fa80-c72d-11e8-8224-18ec5bb58666.png">

micahalcorn:
Secondarily, [this injection of the translated type](https://github.com/OriginProtocol/origin-dapp/blob/master/src/components/listing-create.js#L352) should be replaced with multiple formatted messages; one for each listing type.

jordajm:
@micahalcorn I think it's possible that @sparrowDom fixed this issue [here](https://github.com/OriginProtocol/origin/pull/396), but the listing type names still show up in English because Crowdin auto-translation hasn't been run since that PR was merged. I see errors in the browser console that would indicate this.

micahalcorn:
Cool 🤞 We can close and reopen if necessary.

Issue #1006, opened 2018-11-27
Fractional Usage: Length of Stay Prices


lucadgtrip:
Monthly Prices in the home sharing vertical: I think it's important to consider developing them as early as possible in the process. In this was we could tap an undeserved and growing market of long stays (driven by the raise of remote working, digital nomads and generally more flexible lifestyles around the world).
If you need some data to support the claim, I'll be happy to share.
See also https://www.airbnb.com/sublets which is basically a mini site dedicated to this market (not a niche, but a whole different market in my opinion).

Probably the best way it to replicate Airbnb's approach.
It gives two ways to set monthly prices (28 days or longer)
A) discount on nightly rates: a percentage applied to all months.
B) price per specific month: price for January, February and so on. It overrides A in case both are set.

Price per specific month (B) gives much more control, as with percent (A) the moment I change the nightly price the monthly price changes too.
With B I am able to dedicate some months to tourists and others to longer stays.

In case developing both at this stage is hard we'd need to discuss which one is better.
In my opinion (B) is more effective

![airbnb-price-per-specific-month](https://user-images.githubusercontent.com/34341206/49093365-38e54100-f264-11e8-95a3-696fa61d0dc8.jpg)
![airbnb-length-of-stay-prices](https://user-images.githubusercontent.com/34341206/49093366-38e54100-f264-11e8-8511-826d1c01aef5.jpg)

WEEKLY PRICES
Similarly Airbnb allows to set a discount for a week or for each specific week but I am not sure the specific week is useful as it never corresponds to specific seasons.
Anyway there may be some use case (maybe even major ones) for which it's useful.
I am just not aware of them.
![airbnb-custom-weekly-prices](https://user-images.githubusercontent.com/34341206/49093585-b27d2f00-f264-11e8-9155-ebec1205bbc1.jpg)

@jordajm

micahalcorn:
Great, thanks @lucadgtrip for the thorough request and screenshots. 👌

jordajm:
Yeah, thank you @lucadgtrip 
From a technical point of view, I don't think this would be overly difficult since we've already done similar work on the calendar components and since origin-js and our contracts already support a similar use case for nightly bookings. 

We'd probably create a new UI component (or maybe two - one for buyer's view and one for seller's view) like the monthly price widget in the screenshot and make the components generate a `slots` array for a listing/offer just like the calendar component does. 

@lucadgtrip - what does the monthly booking component on Airbnb look like from the buyer's point of view?

[...]

PR #1013, opened 2018-11-28
Restructured Schemas & Categories

micahalcorn:
⚠️ _**This has breaking changes and should not be merged into master **_ ⚠️

These changes branch from `john/listing_type_in_schema` and are intended to be merged before #1003. The schemas are consolidated so that all categories fit into one of the following types:

- Announcements
- For Rent
- For Sale
- Services

See [this Sheet](https://docs.google.com/spreadsheets/d/1K7ZGnEqLJxPM9MI27sjr4m_U9yyA5DFdylmzOxJq04M/edit#gid=1657275235) for a detailed audit of the "current" state of things vs what is proposed and implemented here. Comments in the second tab ("Current") help navigate the diff and track some of the thinking. I believe that we have a general consensus that we should move forward with this new structure. We know that it's not perfect, but we just want to move forward with something that is more simply comprehended by the user. This new structure also allows us to make some assumptions about schema ➡️ listing type so that we can render the correct form in step 2 of listing creation without worrying too much about the category selection. This will need to be revisited when we start supporting custom schemas per-category, e.g., Housing > Bedroom Count.

@jordajm you're welcome to merge this if you want to, but it should not land in master yet.

@sparrowDom I didn't see any consequences in the searchFilters, but let me know if we need to be concerned with something there or in -discovery.

@franckc we need to figure out how to support existing listings that won't work, such as Taylor Swift's `schema.tickets.music`.

@wanderingstan now may or may not be a good time to look at https://github.com/OriginProtocol/origin/issues/678.

**Remaining Concerns**
- [ ] Broken existing listings lost in translation
- [ ] [Fixtures in origin-js](https://github.com/OriginProtocol/origin/tree/master/origin-js/test/fixtures) 👈 I'll leave this alone to support reproducing the translation issue
- [ ] [Purchase card](https://github.com/OriginProtocol/origin/pull/1013/files#diff-85b33b52b48e4d37892257542742dcceR50)
- [ ] [Listing detail](https://github.com/OriginProtocol/origin/pull/1013/files#diff-6f2d61cbdc05d512a8788e8946bde623R796)


origin-dapp/src/components/listing-detail.js
```diff
@@ -793,6 +793,7 @@ class ListingsDetail extends Component {
                 />
               )}
             </div>
+            {/* to-do: condition viewType based on category within for-rent schemaType rather than schemaType itself */}
```
jordajm: Does this mean that some for-rent listings will be fractional and some will be unit?

micahalcorn: No @jordajm but I suspected that various categories within for-rent would call for an hourly calendar while housing and maybe a few others would call for daily. 📆

micahalcorn: @jordajm can either the seller or buyer choose hourly vs daily? Or do we need to set it per-category for now? Some of them could realistically go either way.

jordajm: Good question. I think we could allow the seller to set hourly vs daily, but the buyer will have to book slots based on however the seller sets it up. I'll have to think through the technical side of letting the seller select what kind of calendar they want (and by extension what length of time constitutes a slot for their listing), but I think it makes a lot of sense. CC @wanderingstan

micahalcorn: Related: what's the max possible duration for each? Can an hourly selection be 25 hours or more? Can a daily selection be 365 days or more? Many of these categories might involve a couple of days but can otherwise be considered long, hourly rentals.

jordajm: I think it's really about the UX... 

If an hourly listing is longer than 24 hrs, the UX is going to get tricky just trying to let the user select all those hours. Same with a daily rental that goes longer than a month. 

I think we should probably lean towards hourly for `<24hrs`, daily for `>24hrs && <1month`, and something like [Luca's monthly booking UX](https://github.com/OriginProtocol/origin/issues/1006) for `>1month`

micahalcorn: Works for me 👍

micahalcorn:
cc @franckc ☝️

Issue #933, opened 2018-11-15
Mobile-Friendly Profile/Identity


micahalcorn:
The profile/identity experience is currently not great on small screens. The following designs have been mocked up to improve this:

https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bd66f264ac6b11e002aba27
https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bd6714490d58d09c85c51ca
https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bd66f266ab2640ac58f18e0
https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bd66f26103fa809941b4696

This issue calls for the following (non-exhaustive):

- [x] arrangement of elements
- [x] truncation of description
- [x] non-modal edit form
- [x] dismissible info box

Changes should be conditioned based on the screen size and should not alter the current large-screen experience.

**PR for this issue:** https://github.com/OriginProtocol/origin/pull/1022

micahalcorn:
This will require nesting at least one view (the edit form) in the navigation. Something similar is being done in [this branch](https://github.com/OriginProtocol/origin/tree/natasha/mobile-messaging-ui) to solve [this issue](https://github.com/OriginProtocol/origin/issues/809), so there should be some coordination on that implementation.

micahalcorn:
Resolved by #1022 ✅

PR #1022, opened 2018-11-29
Sparrow dom/mobile identity

sparrowDom:
### Description:
Changes to make profile part of the DApp mobile friendly:
- rearrange elements according to Zepplin specs
- make guidance message expandable and dismissible on mobile 
- non mobile profile edit form where navigation bar and welcome message disappear
- description text counter on mobile edit form 
- profile description expandable 
- different button styles on attestation modals

Zepplin mocks: 
- https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bd66f264ac6b11e002aba27
- https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bd6714490d58d09c85c51ca
- https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bd66f266ab2640ac58f18e0
- https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bd66f26103fa809941b4696

### Checklist:

- ~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/origin-docs)~

Solves issue: https://github.com/OriginProtocol/origin/issues/933


sparrowDom:
@micahalcorn can you review this when you have time please?

origin-dapp/src/css/app.css
```diff
@@ -5392,6 +5407,47 @@ a:not([href]):not([tabindex]).span-link {
   }
 }
 
+/* sm - */
```
micahalcorn: @sparrowDom does this mean "sm and smaller" or "smaller than sm"? I think it's the latter, in which case I would recommend `xs` as the comment.

origin-dapp/src/css/app.css
```diff
@@ -5490,8 +5546,13 @@ a:not([href]):not([tabindex]).span-link {
   }
 }
 
-/* anything smaller than md */
+/* md - */
```
micahalcorn: @sparrowDom is this a common way to say "smaller and not including md"? I'm not familiar with this. At a glance, it seems to include `md`.

origin-dapp/src/css/app.css
```diff
@@ -5519,6 +5580,76 @@ a:not([href]):not([tabindex]).span-link {
...
+
+  .phone.modal .dropdown .dropdown-item > :last-child {
+    padding-left: 5px;
+  }
+
+  .profile-wrapper .title h2 {
+    margin-bottom: 13px;
+    margin-top: 13px;
+    text-align: center;
+    color: var(--white);
```
micahalcorn: @sparrowDom can we just use `white` instead of adding a variable?

origin-dapp/src/pages/profile/Profile.js
```diff
@@ -163,6 +165,22 @@ class Profile extends Component {
         published: this.props.publishedProgress
       })
     }
+
```
micahalcorn: @sparrowDom can you throw in a comment here to explain what triggers these conditionals to be truthy? Is it just the window resizing?

micahalcorn:
@sparrowDom the CSS may not be working as intended. Do I need to set a variable or anything on my end?

screen shot 2018-12-03 at 12 09 13 pm

micahalcorn:
@sparrowDom also, you may have changed some default strings and then not committed the resulting JSON files. They get automatically generated when you start the DApp [again].

micahalcorn:
@sparrowDom also 😇, there seems to be an odd UX where you upload an image on a small screen and it bounces you out of the form briefly while you edit the crop. I don't know how simple it might be to fix that.

origin-dapp/src/css/app.css
sparrowDom: 👍

origin-dapp/src/css/app.css
sparrowDom: Yup good point. Renaming to `sm -` to avoid confusion

origin-dapp/src/css/app.css
sparrowDom: 👍 good point

origin-dapp/src/pages/profile/Profile.js
sparrowDom: sure, yeah it is window resizing and opening/closing of the edit profile form

sparrowDom:
@micahalcorn regarding the broken CSS I don't know how to reproduce it on my end. Seems like a chunk of it is missing. [This is](https://www.youtube.com/watch?v=_X3d3u2Kuhg) how it works for me. Can you check please if for example you see the same CSS rules applied to `Edit profile` `h2`: 
screenshot 2018-12-04 09 07 34

sparrowDom:
@micahalcorn thanks for the heads up regarding translations. I was running from the box and seems that `origin-dapp` was missing a `translations` volume mapping. So they were generated inside the container but not exposed to the host machine. Fixed in this commit: https://github.com/OriginProtocol/origin/pull/1022/commits/cb8bb7105c9c18c1d6802f38f13e03d3fdd68997

sparrowDom:
@micahalcorn the image cropper UX issue also resolved. Turned out to be a 1 line fix 🕺 

This is ready for another pass.

sparrowDom:
@micahalcorn resolved

micahalcorn:
@sparrowDom one more thing: it seems that we are showing the "Read more v" link regardless of how long the description is. Can we come up with a relatively simply way to detect that the text is long enough to warrant expansion?

Issue #810, opened 2018-10-24
Consistent Event Notifications, Properly Documented


micahalcorn:
[This](https://github.com/OriginProtocol/origin/blame/master/origin-docs/source/includes/resources/marketplace.md#L158-L170) isn't quite accurate. I believe that [on-demand, derived notifications only include `offerCreated`, `offerFinalized`, and `offerAccepted`](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/contractInterface/marketplace/v00_adapter.js#L494-L525). We should expand them to include [all offer-related events](https://github.com/OriginProtocol/origin/blob/master/origin-js/contracts/contracts/marketplace/v00/Marketplace.sol#L32-L39), including rejection/withdrawal, arbitration-related, etc. Let's also ensure that [push notifications](https://github.com/OriginProtocol/origin/blob/master/origin-notifications/app.js#L17-L47) are consistent as well and then document all of it.

micahalcorn:
I started to work on this, but [`getNotifications` relies on `getListing`](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/contractInterface/marketplace/v00_adapter.js#L495), which [excludes withdrawn/rejected offers](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/contractInterface/marketplace/v00_adapter.js#L255). I'm not sure what consequences there would be to changing that behavior, so I'll leave this for now.

micahalcorn:
Resolved by #1025 ✅

Issue #563, opened 2018-09-24
🐃 Invalid IPFS offer file causes never ending error popups


DanielVF:
If you make an offer to someone with invalid IPFS data, then the seller sees a grey popup with an error message about it, over and over and over again.

DanielVF:
This also happens for offer amounts that are too low:

![image](https://user-images.githubusercontent.com/837/45963256-c2705c80-bff0-11e8-8a52-17c0ee14fa2b.png)

franckc:
@DanielVF Can you confirm this only happens if someone creates an offer by interacting directly with IPFS/contracts rather than using origin-js, correct ?

DanielVF:
You can do it with origin.js - just set the offer price lower than the listing price.

PR #1025, opened 2018-11-30
Cleanup DApp notifications

franckc:

### Description:
This covers DApp fixes for #810 - fixes to the notification server will be handled as a separate PR.
- Updated notification logic to include all offer statuses and to filter out notification based on party that initiated the action.
- Added unit test to cover new notification types
- Tweaked a bit code in marketplace tests to make it more explicit what account was being used.
- Cleaned up Notification model
- Updated origin-js doc

### Checklist:


origin-js/src/resources/marketplace.js
```diff
@@ -643,15 +659,24 @@ export default class Marketplace {
             })
           )
         } catch(e) {
+          // TBD: what exactly are we guarding against here ?
```
micahalcorn: @franckc see [this](https://github.com/OriginProtocol/origin-js/issues/558) and probably [this](https://github.com/OriginProtocol/origin/issues/563) can be closed.

franckc: Thanks for pointing this out. Makes sense - we want to protect against invalid offers.
Will add comment. 

Will also add checks in other places of the notification code to harden it further: when fetching listing in this method, and also at the adapter level in the getNotification method (for ex an invalid IPFS Hash could cause that layer to error out).

Issue #1024, opened 2018-11-29
Current User Not Properly Recognized


micahalcorn:
Jon is seeing his own "about the buyer" card and is not getting a call-to-action for finalizing the transaction below. The purchase-detail component is not recognizing that the current user (as defined by web3) is the same as the buyer. My theory is that we are now storing a lowercase `buyerAddress` in the offer but are case-ignorantly comparing this to the account address.

<img width="1555" alt="screen_shot_2018-11-29_at_10 43 42_am" src="https://user-images.githubusercontent.com/273937/49256371-a4095180-f3e3-11e8-88cb-286217a886ee.png">

micahalcorn:
Resolved by #1028 ✅

PR #1028, opened 2018-12-02
DApp Cleanup, Reorganization & Repairs

micahalcorn:
### 😇 Included:
- activation-related Redux state moved from `app` to `activation`
- new `formattedAddress` util added for comparison and display of Ethereum accounts
- prevention of listing creation step 1 skipping to step 2 before "Continue" is clicked
- consolidation of `app.web3.account` into previously-redundant `wallet.address`
- bug fix for when no "support" messaging account was enabled
- fix for https://github.com/OriginProtocol/origin/issues/1024
- prevention of modals when "about token" is visited from seller onboarding
- repair of seller onboarding modal vertical position
- enforcement of checksum format in messaging resource

### 😬 Tested:
- [x] listing creation without OGN
- [x] listing creation with OGN
- [x] blockchain transactions indicator
- [x] offer creation
- [x] offer withdrawal
- [x] offer rejection
- [x] offer acceptance
- [x] offer finalization
- [x] offer data addition
- [x] profile publication
- [x] phone attestation
- [x] email attestation
- [x] facebook attestation
- [x] twitter attestation

Looking for one ✅ but alerting others to changes; let me know if I can help resolve conflicts caused by this.

### DIFF
--- a/origin-dapp/src/utils/user.js
+++ b/origin-dapp/src/utils/user.js
@@ -1,0 +1,14 @@
+import origin from '../services/origin'
+
+const { web3 } = origin.contractService
+
+/**
+ * Takes an Ethereum address and formats it for reliable comparison or display
+ * e.g. 0x627306090abab3a6e1400e9345bc60c78a8bef57 becomes 0x627306090abaB3A6e1400e9345bC60c78a8BEf57
+ * @param {string} an address in any case
+ * @return {string} a case-specific address (currently checksum)
+ */
+
+export function formattedAddress(string) {
+  return web3.utils.toChecksumAddress(string)
+}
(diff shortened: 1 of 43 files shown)


tomlinton:
This is lovely 👍

origin-dapp/src/utils/user.js
```diff
@@ -0,0 +1,14 @@
+import origin from '../services/origin'
+
+const { web3 } = origin.contractService
+
+/**
+ * Takes an Ethereum address and formats it for reliable comparison or display
+ * e.g. 0x627306090abab3a6e1400e9345bc60c78a8bef57 becomes 0x627306090abaB3A6e1400e9345bC60c78a8BEf57
+ * @param {string} an address in any case
+ * @return {string} a case-specific address (currently checksum)
+ */
```
franckc: nit: document that this function throws:
`@throws {Error} if the input address is not a valid Ethereum address`

micahalcorn: @franckc I ran into that when letting `null` get passed in a few times. Do you think it would be better to catch the exception or leave as-is and document the warning?

franckc: I think documenting makes sense to me rather than eating the exception.
It may help us find bugs where address is null but should not...

sparrowDom:
Nice 👍

PR #1029, opened 2018-12-03
Add new field dappSchemaId to origin-js listing schema

franckc:

### Description:

Per [discussion](https://docs.google.com/document/d/19a1TcT47SRpNy1lDwG8EGHLvDe2h28Q2KCDBnb35OrA/edit) last week, adding a new field "dappSchemaId" to the origin-js listing schema.
This is for the DApp to store the ID of the JSON schema it used for creating the listing.

### Checklist:


origin-docs/source/includes/_protocol_schemas.mdown
```diff
@@ -8,6 +8,7 @@ This describes a listing for purchase on the Origin marketplace.
 ```javascript
 {
     schemaId: "http://schema.originprotocol.com/listing_v1.0.0",
+    dappSchemaId: "http://schema.originprotocol.com/forSale.carsTrucks_v1.0.0"
```
jordajm: Two thoughts -
1) Does schema.originprotocol.com exist? I can't find any URLs at that subdomain that work. The dapp makes schema requests to `https://dapp.originprotocol.com/schemas/<schemaName>.json`.
2) Can I suggest this URL as the example instead? I think we'll be using this format in the dapp: `forSale-carsTrucks_1-0-0.json` (per smart thinking from this article: https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/ and the WIP PR here: https://github.com/OriginProtocol/origin/pull/1026)

origin-js/src/ipfsInterface/adapters/listing/v1-listing-adapter.js
```diff
@@ -81,8 +82,8 @@ export default class ListingAdapterV1 extends AdapterBase {
         ? new Money(ipfsData.commission)
         : null
       listing.commissionPerUnit = ipfsData.commissionPerUnit
-      ? new Money(ipfsData.commissionPerUnit)
-      : null
+        ? new Money(ipfsData.commissionPerUnit)
+        : null
```
jordajm: I'm a little confused about the use of `new Money()` as it is used in this file (not just from this new use of it). 

To me it looks like you'd at least have to pass it an object with a key called `amount` in order for it to return anything useful, and ideally, you'd also pass it another key called `currency`, which differs between `OGN` and `ETH` in this file (price is normally ETH and commission is normally OGN), but that's not defined in this file...

How does `new Money()` return a useful object without being explicitly passed `amount` or `currency`?

origin-js/test/fixtures/listing-valid.json
```diff
@@ -1,5 +1,6 @@
 {
   "schemaId": "http://schema.originprotocol.com/listing_v1.0.0",
+  "dappSchemaId": "http://schema.originprotocol.com/forSale.mushrooms_v1.0.0",
```
jordajm: Same comment as above re: URL

origin-js/test/ipfs_store.test.js
```diff
@@ -65,6 +66,16 @@ describe('Listing IpfsDataStore load', () => {
...
+  it(`Should load a listing with no dappSchemaId`, async () => {
+    const listingNoDappSChemaId = Object.assign({}, validListing)
+    delete listingNoDappSChemaId.dappSchemaId
+    mockIpfsService.loadObjFromFile = sinon.stub().resolves(listingNoDappSChemaId)
+    mockIpfsService.rewriteUrl = sinon.stub().returns('http://test-gateway')
+
+    const listing = await store.load(LISTING_DATA_TYPE, 'TestHash')
+    expect(listing.dappSchemaId).to.equal(undefined)
+  })
+
```
jordajm: Would be great to see `dappSchemaId` added to a test listing data set and make sure it gets returned by a call to `getListing()` in the marketplace API

origin-docs/source/includes/_protocol_schemas.mdown
franckc: Thanks for thinking through all these details, super useful.
1. No, schema.originprotocol.com hasn't been created yet. But I was planning on setting it up at some point in the future since it fits well with our goals of keeping all data public and easily accessible. I was thinking it would serve schemas that are checked in our origin-js repo.
Regarding dapp schema URL, I think it's ok for now to keep using https://dapp.originprotocol.com/schemas since it is working well. In future maybe we can mirror those schemas under http://schema.originprotocol.com/dapp/

2. For the schema name, using schemaVer sounds good. Couple details:
 - I'm not really convinced using "-" rather than "." in version is super helpful. The schemaVer blog mentions that it helps differentiating a semver version vs a schemaVer version, but since we are by definition dealing with schemas, not sure it is needed ?
 - the JSON schema `#schema` property does not include the extension JSON. It is: [http://json-schema.org/draft-06/schema#](http://json-schema.org/draft-06/schema#) That's why I did not have it for the origin-js schemas. 
Overall I don't really feel strongly one way or the other but I do think it would be good to be consistent in our naming for origin-js and dapp schemas. I'm happy to make the change to this new schemaName format on origin-js if you think it is better. Lmk your thoughts.

origin-js/src/ipfsInterface/adapters/listing/v1-listing-adapter.js
franckc: commissionPerUnit is an object with a amount and currency. See [listing schema](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/ipfsInterface/schemas/listing.json#L88).

origin-js/test/ipfs_store.test.js
franckc: Good point. Let me add a unit test to marketplace.

origin-docs/source/includes/_protocol_schemas.mdown
franckc: @jordajm Let me know if you have any input on the above. We can also have a quick video meeting to chat and settle on all this if that's easier...

jordajm: @franckc - thanks for the reminder. 

1. 👍 
2. I agree with your comment about "-" vs "." - I changed the dapp schemas back to "." in my PR.
3. I decided to leave the `.json` in the dapp schema IDs because without that, it doesn't work as a link when requesting the file directly from dapp.originprotocol.com/schemas. Here's an example of the new `dappSchemaId` format in my dapp PR: 
`"dappSchemaId": "https://dapp.originprotocol.com/schemas/forSale-toysGames_1.0.0.json"`

For reference the PR is here: https://github.com/OriginProtocol/origin/pull/1026

origin-js/src/ipfsInterface/adapters/listing/v1-listing-adapter.js
jordajm: Ahh, of course - thanks for explaining

origin-docs/source/includes/_protocol_schemas.mdown
franckc: Sounds good. Ok, let's add the .json extension to those origin-js schemas to be consistent. I'll do that as a separate PR.

I also see you have "_1.0.0.json" whereas in origin-js we have "_v1.0.0". I don't think there is really a good reason to have the "v". So i'll remove that as well.

franckc: Oh and also, let me use "https" rather than "http".

PR #1030, opened 2018-12-03
Nick/remove jquery

nick:
### Description:

- Removes jQuery as a dependency
- Bootstrap JS no longer included in bundle

### Checklist:


scripts/clean-package-locks.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
```
micahalcorn: 😭

micahalcorn:
@nick the one meaningful issue that I found is with the phone attestation modal. There is a dropdown there for selecting the country code.

I think it would also be preferable for the notifications dropdown to hide when a notification is clicked (and the route changes to purchase-detail); but not a big deal. And I don't know if it's preexisting or not, but the bottom link in some modals (such as ["Skip" here](https://github.com/OriginProtocol/origin/blob/nick/remove-jquery/origin-dapp/src/components/listing-detail.js#L330-L338)) should be smaller text (maybe `0.875rem`) and have some vertical spacing above. A concern there would be that style changes would affect other modals that currently look fine (such as the edit profile form).

micahalcorn:
@nick also, this one seems to have broken images.

screen shot 2018-12-03 at 1 18 20 pm

PR #1041, opened 2018-12-04
Enable hoisting and integration tests

tomlinton:

### Description:

- Synced modules across packages so hoisting is efficient. Generally minor version bumps. This included some changes to package versions in your experimental stuff @nick which perhaps weren't entirely necessary, but it makes `npm install` from root faster.
- Changed origin-box to use a single image for all services. This builds a lot faster than the previous setup of separate images due to the implementation of hoisting (and is what makes it possible to run integration tests in TravisCI in a reasonable time).
- Renamed the name of some packages in their `package.json` files to match the directory name. This was `origin-simple-demo` to `origin-graphql-simple-demo @nick and `OriginCatcher` to `origin-mobile` + `origin-mobile` to `origin-linking` @crazybuster and @micahalcorn. Let me know if this causes a problem for you. It makes it a lot easier when running commands with `lerna --scope` if the package names match the directory name.
- Updated some CSS imports in the DApp to be non-relative so they work with hoisted dependencies.
- Implemented a root config for `eslint` and extended it in each of the packages.
- Removed a bunch of `package-lock.json` files for dependencies where they are no longer used due to one existing in the root for hoisting.
- Add a new `origin-tests` that runs integration tests. Included basic tests for verifying the state of `origin-box` after it starts and a basic example of adding a listing and checking it gets indexed by the event listener and is retrievable from the discovery server.
- Updates TravisCI config file to run the new tests.
- Adding `TESTING.md` at root that describes the testing setup.
- Fixed some T3 linting errors.

Bit of a hairy PR. Hit me with any questions! I'll present some explanations of hoisting and how this all works in the engineering call tomorrow. I don't want to merge this until after I've given everyone a heads up. 

### Checklist:


tomlinton:
Integration tests take 9-10 minutes to run on TravisCI which isn't too bad. Of course the vast majority is building/starting `origin-box`. I tried to increase performance by caching the Docker images but `docker load` seemed really slow, I might revisit later.

origin-tests/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "description": "Integration tests for Origin Protocol",
+  "main": "index.js",
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
+    "test": "mocha --timeout 10000 --exit test"
+  },
+  "author": "Tom Linton <tom@originprotocol.com>",
+  "license": "ISC",
```
nick: Tiny thing but I think we're MIT for everything

tomlinton: Thanks, think this was the default for `npm init`, fixed it up!

tomlinton:
Thanks @nick, I'll go through and clean up some of the downgrades. I did get a bit lazy as I was just using the output of `lerna --hoist` which generates warnings about mismatched packages.

origin-tests/test/1-create-listing.js
```diff
@@ -0,0 +1,38 @@
...
+    // Wait to allow event-listener to process listing
+    return new Promise(resolve => setTimeout(resolve, 4000))
+  })
+
+  it('should allow created listing to be retrieved from discovery', async () => {
+    const listings = await this.origin.marketplace.getListings()
+    assert.equal(listings.length, 6)
+  })
+
+  it('should allow created listing to be searched from discovery', () => {
```
franckc: nit: add TODO or remove that test.

tomlinton:
What do you mean @franckc? There will be no issues to fix! :grin:

PR #1042, opened 2018-12-05
Multi-unit listing fixes

cuongdo:
### Description:

Most of the changes are a consequence of changing all listings, including single-unit listings, to deducting inventory as valid offers are made. Previously, buyers could make multiple valid offers on single-unit listings, and our marketplace tests were relying on that behavior.

Note that I'm not yet tackling the issue of vanishing offers, where an edit to a listing invalidates a previously valid offer. I wanted to split things up into smaller PRs and think a little more about how to fix that issue.

#### Decrement units available upon valid offer

This changes both single-unit and multi-unit listings so that units are
considered unavailable when the listing receives a valid offer in the
'created' state, instead of the 'accepted' state as we've traditionally
done.

#### Handle depleted listing commission for offers

This change handles the case where the listing commission doesn't cover
all units of the listing. Because commission validation now relies on
the full set of offers, this changes also moves detection of invalid
commissions from getOffer() to getOffers().

### Checklist:

- ~~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/origin-docs)~~


cuongdo:
@franckc Thanks for the review. I've added a TODO in getOffers. I may tackle that together with the change to filter offers based on their validity at creation time.

Issue #1052, opened 2018-12-06
[DApp] Failure to send message to buyer from transaction details page


franckc:
@AustinVirts Ran into this error while trying to message a buyer from the transaction details page
<img width="1612" alt="screenshot" src="https://user-images.githubusercontent.com/945910/49606510-6b6bf980-f948-11e8-8b4d-cfc86b635f23.png">

Notice the address of the recipient of the message is "undefined"

PR #1054, opened 2018-12-07
Address Improvements

micahalcorn:
This resolves an issue of the "contact" button not appearing on the public user profile and also may help prevent or diagnose the problem underlying #1052. 🤞


franckc:
FYI deployed this fix to staging so that it gets included in tomorrow's prod push.

Issue #1044, opened 2018-12-05
Multi-Unit Listings


micahalcorn:
Sellers should be able to create listings with a quantity available greater than one. Buyers should be able to purchase more than one unit in a single purchase. This functionality should be available for any non-fractional listing category that supports a non-zero price. This will soon equate to listings in the "For Sale" and "Services" categories.

The designs for multi-unit UI are integrated [here](https://app.zeplin.io/project/59fa2311bac7acbc8d953da9?seid=5be35b99d2521e3d0a7d3e09). The following changes are not reflected in the mockups:

- [x] There will be no "E-commerce" category.
- [x] The photo-related elements should remain at the bottom of the listing creation form.
- [x] The quantity field should default to a value of `1` for both the seller and the buyer.
- [x] The price field label should change from "Price" to "Price (per unit)" when the quantity field value changes from 1 to >1.
- [x] No "Get OGN" elements should be displayed.
- [x] All ETH values should contain exactly five decimal places.
- [x] The "Buy Now" button copy should remain "Purchase".
- [x] The transaction detail page should conform to the design prescribed [here](https://github.com/OriginProtocol/origin/issues/1038).
- [ ] ~Listing status should be displayed as "Pending" when the currently-available quantity is `0` and there _is_ one or more open, valid offers.~
- [x] Listing status should be displayed as "Sold Out" when the currently-available quantity is `0` and there _are no_ open, valid offers.

Important behavioural points:
- [x] do not let user edit the quantity of the listing below the number of accepted(?) offers
- [x] show the amount of the inventory according to the number of created(?) offers
- [ ] in elastic search listener set the boost to the remaining commission amount:  ```min(commission - commisionPerUnit * unitsSold, commissionPerUnit)```

Also keep in mind the following [corner cases](https://docs.google.com/document/d/19zrYx15LEdFzbkquH9AyoUEE3_fNkVqdH4uTJgSJuA0/edit#). 

**Pull Requests for this ticket:**
- https://github.com/OriginProtocol/origin/issues/1044#event-2006293854 (listing creation flow)
- https://github.com/OriginProtocol/origin/pull/1102 (everything else in the DApp)

sparrowDom:
thanks @micahalcorn for creating this issue

micahalcorn:
@sparrowDom thanks! 👏 Do you know if any of the items listed above should be retained as separate issues?

sparrowDom:
[...]

PR #1059, opened 2018-12-08
Multi unit DApp 

sparrowDom:
### Description:
This PR makes the required changes so that multi unit listings can be created in the DApp. This PR covers the createListing flow. 

Changes: 
- add quantity field when selected schema is `for-sale` or `services`
- add a boost limit field when listing is a multi unit listing (`quantity` > 1)
- the price field label changes from `Price` to `Price (per unit)` when there is a multi unit listing
- boost limit is automatically set to maximum boost limit (or maximum amount of funds in user's wallet) when user changes boost slider. 
- there is a progress bar in listing creation flow indicating

Zepplin designs: 
- https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5be35b63072fb97e8fe7b1ab
- https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5be35b65cc07e83f3df8abce
- https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5be35b631a44b33f2f36e8b7
- https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5be35b64cc07e83f3df8ab52

**Related issue:** https://github.com/OriginProtocol/origin/issues/1044#event-2006293854


sparrowDom:
@micahalcorn I am splitting the changes of multiunit listing DApp ticket to multiple smaller PRs in order to avoid long living giant ones.

origin-dapp/src/actions/Listing.js
```diff
@@ -8,11 +8,19 @@ export const ListingConstants = keyMirror(
   {
     FETCH_IDS: null,
     FETCH_IDS_SUCCESS: null,
-    FETCH_IDS_ERROR: null
+    FETCH_IDS_ERROR: null,
+    IS_MULTI_UNIT_LISTING: null
```
micahalcorn: @sparrowDom what do you think about storing this in the listing-create component state? Or better yet, can this be computed based on the the step 2 form state, or is it affected by more than just the quantity field?

origin-dapp/src/components/form-widgets/boost-limit-field.js
```diff
@@ -0,0 +1,113 @@
...
+      setTimeout(() => {
+        this.props.onChange(boostLimit)
+      })
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    const maxPossibleBoost = this.calculateMaxPossibleBoost(this.props)
+    // user is changing boost slider
+    if (this.props.formContext.formData.boostValue != prevProps.formContext.formData.boostValue &&
```
micahalcorn: Let's use strict equality unless there's a compelling reason not to.

origin-dapp/src/actions/Listing.js
sparrowDom: Yup good thinking. I remember hitting a wall, not knowing how to pass this information down to `price-field` since it doesn't get the full scope of `formData` to access the quantity value. But then I see there is a `formContext` field in the react form, that passes the object to all the children fields like `price-field`. Using that and avoiding storing values in Redux.

origin-dapp/src/components/form-widgets/boost-limit-field.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  componentDidMount() {
+    // If a price is passed in, we must call the onChange callback
+    // to set the price in the parent form
+    // Unfortunately, the setTimeout is needed to allow the parent
+    // form to render and be ready to handle the onChange event
+    const { boostLimit } = this.state
+    if (boostLimit) {
+      setTimeout(() => {
```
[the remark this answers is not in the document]
sparrowDom: Yup exactly `react-jsonschema-form` Form component is not receiving `onChange` updates in the first frame, that is why we wait one out. It is a hacky solution :/  

We are using the same approach in the `price-field`: https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/components/form-widgets/price-field.js#L24-L35

sparrowDom:
Thank you both for the review 🙏 All the comments have been addressed

origin-dapp/src/components/form-widgets/boost-limit-field.js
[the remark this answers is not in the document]
cuongdo: Are there any guarantees that `onChange` will fire after the `setTimeout` callback happens? I haven't noticed any issues testing on Chrome, but I didn't do thorough testing.

cuongdo:
If you think you can get the other parts of the multi-unit listing UI done soon, maybe we could merge this PR and other multi-unit PRs close to one another?

sparrowDom:
@cuongdo that sounds good to merge everything in master at the same time.

micahalcorn:
I'm working on resolving these conflicts. 🙁

cuongdo:
Thanks for taking a look, @jordajm !

micahalcorn:
@jordajm we intended services to _not_ be fractional.

micahalcorn:
@jordajm do we need to revert some commits in this branch for services-related changes?

jordajm:
Yes @micahalcorn - I'll make this match up with the latest dapp schemas that got merged last night - I should have time for that this morning

jordajm:
@micahalcorn actually the schemas already look good to me. I think the merge of master into this branch recently must have synced everything up.

micahalcorn:
@jordajm are we not adding `unitsTotal` to `public/schemas/services-*`?

micahalcorn:
@jordajm isn't that what we do we the price field for announcements since origin-js expects a value (`0`)?

jordajm:
Good point @micahalcorn - I hadn't thought about announcements needing a price of zero. We could do the same thing actually to make it clear to dapp developers that announcements need a price of `0`. (i.e. add a price field to the announcements dapp schema and set the default value to zero)

jordajm:
To avoid having a mess of category-based default value setting in the dapp, we should probably implement a helper that compares the form data to the schema and if the form data doesn't contain a value for one of the required fields in the schema, it sets the value in the form data to the default value from the schema. 

This is an implementation detail but I thought I'd mention it here because it's a departure from the way this currently works in `listing-create.js`.

franckc:
@jordajm Would it make sense to use a constant ? And when the DApp reads the schema, if it finds a constant it would not show it in the UI. But the constant value would still be passed to the method creating listing. This could be a general rule applies to any field.

jordajm:
That would cover most use cases @franckc... however, there is (or will be) a case in fractional listings where `slotLength` and `slotLengthUnit` are enums numbers with defaults that need to be compared to form data to see if they've been overridden, and the default needs to be applied to form data if not. But I like the idea of using `const` in the schemas where possible and then just looking for those in the dapp and hiding them if they exist as well as passing the `const` value right into the form data.

jordajm:
okay, after some more time thinking about this, I think we should probably:
1) on `services` schemas, add `unitsTotal` as a const field with a value of `-1`
2) on `announcements` schemas, add `price` as a const field with a value of `0`

In the `listing-create` component, we make a util that looks at the schema and hides all `const` fields from the form while adding their key/val to the form data. It would also compare all non-const fields to the form data and if there isn't a value for a given field in the user-generated form data, we assign the key/val of the schema field using the default value. 

how does that sound @franckc @micahalcorn ?

micahalcorn:
@jordajm yep, that was the case for the original announcements schema. 👍

jordajm:
@micahalcorn @franckc @sparrowDom - I added some commits that address the above comments

franckc:
@jordajm +1 for your proposal. Thanks for sweating those details ! Haven't looked at your code yet but will review tomorrow...

PR #1065, opened 2018-12-10
Update origin-js schemaId format

franckc:
### Description:
Updating origin-js schemas to be consistent with the format chosen for DApp schemas.

The old format was:
`http://schema.originprotocol.com/<dataType>_v<version>`
For ex:
`http://schema.originprotocol.com/listing_v1.0.0`

The new format is:
`https://schema.originprotocol.com/<dataType>_<version>.json`
For ex:
`https://schema.originprotocol.com/listing_1.0.0.json`

Notes: the code changes are designed to maintain backward compatibility since we don't want to break the existing couple hundred listings in production that are using old format schemaIds.

### Checklist:


experimental/origin-admin/src/pages/marketplace/mutations/_demoListings.js
```diff
@@ -1,8 +1,9 @@
 const HawaiiHouse = {
-  schemaId: 'http://schema.originprotocol.com/listing_v1.0.0',
+  schemaId: 'https://schema.originprotocol.com/listing_1.0.0.json',
+  dappSchemaId: "https://dapp.originprotocol.com/schemas/forRent-housing_1.0.0.json",
   listingType: 'unit',
-  category: 'schema.housing',
-  subCategory: 'schema.housing.vacationRentals',
+  category: 'schema.forRent',
+  subCategory: 'schema.forRent.housing',
```
jordajm: @franckc - in my PR here:
https://github.com/OriginProtocol/origin/pull/1026
I removed the second namespace in schema message IDs so they're all just `schema.<messageId>` because that allowed me to simplify the code quite a bit.

franckc: Ah ok. I can make that change. If we are going that route, do we even need to keep the "schema." prefix for category and subCategory ?

jordajm: Personally, I think it makes sense to keep the `schema.` prefix because it helps to differentiate translatable "message IDs" that exist in our dictionary from normal strings - open to feedback/ideas though.

franckc: Done. @jordajm ready for another round of review...

PR #1069, opened 2018-12-10
Use block epoch to speed up origin.js unit tests

cuongdo:
This change adds the optional blockEpoch parameter to the origin.js
resources we use in our tests. This results in a substantial speed
increase for our tests, because they don't have to scan over blocks
generated by previous tests.

**This change seems to decreased Travis CI Origin.js unit test runtime from ~8 minutes to ~4.5 minutes.**

### Checklist:


tomlinton:
This is brilliant, good find @cuongdo :grin:

PR #1073, opened 2018-12-11
Add support for multi-units in perf mode

franckc:
### Description:
Update getOffers logic to work in performance mode:
 - offers data is fetched either from discovery or blockchain based on perf. mode setting
 - filtering of invalid offers is subsequently applied in both cases

### Checklist:


sparrowDom:
Just wanted to raise a point (not related to the code of this PR) since we are filtering out the invalid offers (not enough unitsAvailable to satisfy offers) users of invalid offers are probably not going to be able to cancel those offers?

cuongdo:
@sparrowDom Offers with enough `unitsAvailable` at the time of the offer should be fine, because we [prevent updates of listings that would make unitsAvailable go negative](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/resources/marketplace.js#L392).

sparrowDom:
@cuongdo yup I agree that edge case is not a problem. I was thinking of another edge case @franckc mentioned the last time we were speaking, and that is if 2 users make an offer at the same time, and there is only 1 more `unitAvailable`. Both of those offers should be mined (since they were valid at the time of offer creation), but only the first one that gets added to blockchain will be valid in the end.

cuongdo:
@sparrowDom Yes, I believe that is an issue. I had some experimental code in a local branch for fixing that issue. Maybe I'll revive that code.

PR #1077, opened 2018-12-12
add package.json volume mappings

sparrowDom:
### Description:
Add package.json volume mappings to `docker-compose`


tomlinton:
Thanks @sparrowDom!

Issue #1087, opened 2018-12-13
[Discovery] Backfill offer.commission


franckc:
offer.commission were recently added to the origin-js model and therefore the old offers that have already been indexed do not have those fields.

There is a workaround to retrieve the data via the IPFS object (which is indexed) but it would be cleaner to backfill.

nick:
Don't think this is an issue post-rewrite

PR #1088, opened 2018-12-13
[Discovery] Fix offer commission

franckc:
### Description:
As part of the multi-unit feature, the origin-js getOffers method started relying on offer.commission.
In performance mode, this data is fetched from the discovery back-end.

Unfortunately offer.commission did not get indexed for older listings because the origin-js offer model did not have that field (it got added recently) and we did not backfill (I filed #1087 for that).
As a result, offer.commission was returned as null which then we causing getOffers to filter out the offer. See this [check](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/resources/marketplace.js#L285).

This CL fixes it by using the workaround to fetch the commission from the ipfs data that is luckily indexed. It also exposes these fields thru the graphql interface for easier debugging.

### Checklist:


franckc:
@sparrowDom Yes, a re-indexing script would be very useful. In the longer-term. An alternative design for the indexing server would be to store the raw immutable chain data, events and ipfs blobs rather than storing the output of origin-js methods. In performance mode, origin-js would use that raw data, while in regular mode it would query blockchain/ipfs. This way changes to origin logic js would not require a re-indexing... But that's obviously a pretty large rewrite of the indexing system and of origin-js code, and I don't think that's a priority right now since that is not a huge painpoint. Also it would not address the problem of re-indexing data in the search index. That would still need to happen. Anyway, just food for thoughts :)

PR #1096, opened 2018-12-15
[Identity] Profile registry contract

franckc:
### Description:

Simple profile registry contract to map users ETH address to IPFS hashes of their profile. Relies on logs for storage. The costs are:
 - profile update: 24,908 gas (~0.6 cent at current ETH price)
 - profile delete: 23,858 gas (~0.6 cent at current ETH price)

We could store the mapping address-> ipfsHash in the contract and that would increase the costs slightly:
- profile update: 65,690 gas (1.6 cent at current ETH price)
- profile delete: 21,102 gas (0.5 cent at current ETH price)
The advantage would be faster profile lookup. But given that a) we are already heavily committed to an event sourcing approach with the marketplace contract and b) we can index data in the back-end to solve performance concerns, I'm leaning towards the cheaper log only solution. 

### Checklist:


franckc:
Had a good discussion offline with @nick 
He suggested a possible alternative:
```
contract IdentityEvents {
    event IdentityEvent(
        address indexed sender, // Source account
        uint256 indexed topic,  // 0: Update, 1: Delete
        address indexed target, // Target account (0x0 for self)
        bytes data              // IPFS hash or other data
    );
    
    function emitEvent(uint256 topicId, address target, bytes data) public {
        emit IdentityEvent(msg.sender, topicId, target, data);
    }
} 
```

Pros
 - more open/generic. allows to support recording an action about another identity than myself. ex: initiating a conversation, or claiming some sort of association with another identity - like I belong to an organization or a group.

Cons
 - slightly more expensive
 - not clear there are strong uses cases that would benefit from the expanded capabilities
 - requires more logic to validate and interpret events on listener side

On the other hand the contract proposed originally in this PR:
Pros
 - can not be simpler and cheaper

Cons
 - not extensible. Probably ok to use in the short-term as part of our Q1 growth experiment to determine if dramatically lowering profile storage costs helps with user adoption or if a free solution is truly the only thing that will do it.

Everyone, feel free to comment with your thoughts ! :)

franckc:
Allright, going ahead with the dummiest approach then ! :)

PR #1102, opened 2018-12-17
Sparrow dom/multi unit ui2

sparrowDom:
### Description:
Continues to build the DApp support for multiunit listings initiated by Pull Request: https://github.com/OriginProtocol/origin/pull/1059

Changes in this Pull Request: 
- editing a quantity below units locked in active and sold offers not allowed
- `my-purchase` card look updated for multi unit listings
- `listing-card` and `listing-detail` badges logic updated and refactored inside a single util class
- quantity field moved above price field
- logic of loading offers in a listing moved to `origin-js`. This enables listing model from `origin-js` to be enriched with offer related information. (e.g. boostRemaining, unitsRemaining)
- changed `origin-js'` getListing API so adding additional options will not change the function signature
- when purchasing a multi unit listing a new `quantity` field is available
- in `listing-create` form step number displayed is now also used in the (steps) progress bar
- More complex logic with listing detail. For single unit listings everything stays the same, for multi unit: 
  - show sold badge only when all offers with state sold have used up the unitsAvailable
  - show seller information how many units have been bought, and how many are available. Also how much boost is remaining
  - besides the above point if there is a listing pending show that additional information to the seller
  - show the buyer the form to buy more units if any are available. Besides that information also show a link to a pending purhcase if one is active
- add Offer detail information to transaction detail view

### Things not yet resolved: 
- 2 buyers make an offer at the same time, and only the one that is created first is valid. - not sure how to solve this yet (This is also a problem with single unit listings, and fractional)
- editing a listing. Origin-js needs to check for offer validity against the version of the listing for which the offer was created. 

### Checklist:

**Related issue:**https://github.com/OriginProtocol/origin/issues/1044#event-2006293854
[Document](https://docs.google.com/document/d/1jSMlcqJqUQ7_EZQHJwV2mu9BdTii-29annKQ70BDHpA/edit) with listing detail states


origin-js/src/models/listing.js
```diff
@@ -24,8 +24,8 @@ export class Listing {
    *  - {string} seller - address of the seller
    *  - {string} display - 'normal', 'featured', 'hidden'
    *  - {Array<Object>} media
-   *  - {Object} comission - consists of 'amount' and 'currency' properties
-   *  - {Array} slots
+   *  - {Object} commission - consists of 'amount' and 'currency' properties
```
cuongdo: nit: (I forgot to do this) while you're here, could you add a comment for `commissionPerUnit`?

origin-js/src/resources/marketplace.js
```diff
@@ -176,17 +171,30 @@ export default class Marketplace {
    * @param {string} listingId
    * @param {{blockNumber: integer, logIndex: integer}} blockInfo - Optional argument
    *   to indicate a specific version of the listing should be loaded.
+   * @param opts: {loadOffers: boolean} also load offers for this listing
    * @returns {Promise<Listing>}
    * @throws {Error}
    */
-  async getListing(listingId, blockInfo) {
+  async getListing(listingId, blockInfo, opts = {}) {
```
cuongdo: Probably in a different PR, I would move `blockInfo` into `opts`, because the former is an optional parameter as well.

origin-js/src/resources/marketplace.js
```diff
@@ -244,52 +252,74 @@ export default class Marketplace {
...
-          remainingCommission,
-          commissionPerUnit.times(offer.unitsPurchased)
-        )
-        const offerCommission = offer.commission && BigNumber(offer.commission.amount)
-        if (!offerCommission || !offerCommission.isEqualTo(expectedCommission)) {
-          return
+      }
+      /* Validate units purchased against units available.
+       *
+       * TODO: Validate units purchased vs. units available at offer creation
```
cuongdo: We already validate units during offer creation. The code is here: https://github.com/OriginProtocol/origin/blob/master/origin-js/src/resources/marketplace.js#L443

origin-js/src/resources/marketplace.js
```diff
@@ -244,52 +252,74 @@ export default class Marketplace {
...
+       * time. This will handle the case where an edit of a listing invalidates
```
cuongdo: I don't understand how validating units during offer creation handles this case. We do handle this case in `updateListing` though. See: https://github.com/OriginProtocol/origin/blob/master/origin-js/src/resources/marketplace.js#L395

origin-js/src/resources/marketplace.js
```diff
@@ -244,52 +252,74 @@ export default class Marketplace {
...
+        return false
+      }
+
+      // listing has no commission
+      if (!isMultiUnit && (commissionPerUnit === null || commissionPerUnit.isEqualTo(0))) {
+        onValidOffer()
+        return true
+      }
+      else if (!isMultiUnit) {
+        if (offerCommission && offerCommission.isEqualTo(commission)){
```
cuongdo: is there corner case where the seller increases the commission through a listing update and invalidates previous offers that have the original commission? if so, a TODO would be good here

cuongdo:
Also, if you want to fix your CI, merge in `master`. #1110 fixed a broken test.

origin-js/src/models/listing.js
sparrowDom: sure 👍

origin-js/src/resources/marketplace.js
sparrowDom: Good thinking. Let me do it in this PR, so we have a minimal as possible `origin-js` API changes.

origin-js/src/resources/marketplace.js
sparrowDom: Tnx, this comment is totally off. Let me remove it.

origin-js/src/resources/marketplace.js
sparrowDom: Yeah this comment is off. Let me change it to: 
```
Validate units purchased against units available.

This check is required because of an edge case, where 2 users can make an offer at the same time for the remaining amount of units in a listing. Both offers will get mined, since they are both valid at the time transaction to the blockchain is issued. After they are mined, only the first one is valid.
```

origin-js/src/resources/marketplace.js
sparrowDom: good thinking. Currently a seller can not edit the overall commission (boostLimit) and commission per unit (boost) when editing a listing. Will add still add a comment, just in case we enable that in the future.

sparrowDom:
Thanks @cuongdo for the review. Have addressed all the comments, I will let you know once the IU changes are final. Also thanks for the CI tip :)

sparrowDom:
@micahalcorn and @cuongdo please take a look when you have time. 

There is some more complexity when dealing with multiunit listings that still have units available, but also offers that are relevant to the current user (current user is buyer / seller). Currently I have solved this by displaying the relevant offer information besides the buy form (for buyer) and units purchased statistics (for seller). See screenshots below: 
screenshot 2018-12-21 14 52 08
screenshot 2018-12-21 14 53 14

Do you think this is an ok solution?

micahalcorn:
@sparrowDom we should only show the status box if there is no available supply (and thus no "buy box"). This component should be all about listing detail rather than offer detail. We do need to account for offer(s) status(es) to know whether to show the overall status as "unavailable" or "sold out" (per a deeply nested, resolve comment reply [here](https://docs.google.com/a/originprotocol.com/document/d/19zrYx15LEdFzbkquH9AyoUEE3_fNkVqdH4uTJgSJuA0/edit?disco=AAAACXYrvac)):

_'Pending' no longer makes sense since we have multi-unit. When there is 0 available inventory, we will now use the language 'Unavailable' in the UI. We will still have 'Sold out' if all offers have been finalized and inventory is 0._

cuongdo:
@sparrowDom I'll take a look later today

origin-dapp/src/components/listing-create.js
```diff
@@ -165,7 +170,11 @@ class ListingCreate extends Component {
       try {
         // Pass false as second param so category doesn't get translated
         // because the form only understands the category ID, not the translated phrase
-        const listing = await getListing(this.props.listingId, false)
+        const listing = await getListing(this.props.listingId, { translate: false, loadOffers: true })
+        if (listing.isMultiUnit) {
+          listing.boostLimit = listing.totalBoostValue
+          delete listing.totalBoostValue
```
cuongdo: why is this `delete` needed?

origin-dapp/src/components/my-purchase-card.js
```diff
@@ -109,22 +138,21 @@ class MyPurchaseCard extends Component {
                   />
                 </Fragment>
               )}
-              <div className="actions d-flex">
+              {/*<div className="actions d-flex">
```
cuongdo: please remove this commented out code

origin-contracts/migrations/3_create_sample_listings.js
```diff
@@ -56,7 +56,7 @@ async function deployContracts(deployer) {
   // QmXx5Qo2xh5TWZXbsDmyquBpojpqoUVpgZt3kNgqsonw9j
   await createListing(
     marketplace00,
-    '0x8ecc70d4c694d22f72efbbe154fba7b4c9bae69f0b6f02d92ddcbcde24798b3a',
+    '0x929c6ef53a5fd00e51d935e228f4719b522b6b5d2076bf0d725b0b5ab2c6fe4c',
```
sparrowDom: thanks for this! 👍

origin-dapp/public/schemas/forSale-antiques_1.0.0.json
```diff
@@ -33,14 +33,14 @@
...
     },
-    "price": {
-      "type": "number",
-      "title": "schema.priceInETH"
-    },
     "unitsTotal": {
       "type": "integer",
       "title": "schema.unitsTotal"
     },
+    "price": {
```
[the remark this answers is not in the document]
sparrowDom: The order of fields in the schema files also dictates the order of fields in the `create-listing` form. There was a UX decision that it is more natural to the user to fill out number of units (quantity) before filling out the price because the label on the price field changes from `price` to `price per unit` when quantity is bigger than 1.

origin-dapp/src/components/listing-create.js
sparrowDom: Thanks for pointing to this. The logic was that we don't allow editing commission / commission per unit, and those fields should for that reason be removed. But now that I think about it that logic is flawed. We should load both commissions and leave them unchanged when submitting an edited listing.

removing this if condition

origin-dapp/src/components/my-purchase-card.js
sparrowDom: agreed, thanks!

sparrowDom:
@micahalcorn `unitsTotal` remains unchanged when offers are being created and accepted. That value can only change when user edits a listing and even then he can not bring quantity below the number of currently active offers. So If we have a listing with 3️⃣ units total and 2️⃣  offers have been made (are active) the seller can not decrease that number below 2️⃣ 

There is a property calculated in `origin-js/src/models/listing.js` that computes `unitsRemaining`. Which is `unitsTotal` - `NoOfUnitsInActiveOffers`.

With that in mind we can use `unitsTotal > 1 && listingType='unit'` to distinguish between multiUnit and non multiUnit listings.

cuongdo:
@sparrowDom Yes, we can save the quantity issue for another PR. Would you mind filing that issue?

Everything else looks good to me. I think we should test both single-unit and multi-unit on staging before launching this to prod. cc @franckc as a heads up

micahalcorn:
Thanks @sparrowDom 🙂 It sounds like there was a change made along the way that I missed or a miscommunication. But that's great that we can now distinguish between the two types. Have you already made the changes to the listings-details page that you proposed? I will take a look when they are ready.

sparrowDom:
@cuongdo agreed, a thorough testing on staging is definitely a good idea. Also filed the issue here: https://github.com/OriginProtocol/origin/issues/1141

@micahalcorn doing the changes now, will notify here once they are done.

sparrowDom:
btw in case it helps somebody running `npm run translations` inside the `origin-dapp` updates `translated-messages.json` and makes those "Missing message" errors go away.

franckc:
Thanks for the heads up. I'm happy to deploy this to staging for testing once ready... Just ping me...

cuongdo:
@micahalcorn feel free to merge once you’ve had a chance to look this over

micahalcorn:
Thanks so much for the screenshots @sparrowDom 👍 I think we've got quite a few things to optimize there, but we shouldn't hold up this merge. I'm going to run this locally and check on a few things before making it happen.

cuongdo:
@sparrowDom btw there's a minor JS lint error causing CI to fail

cuongdo:
I've pushed what should be a fix for the lint issue

sparrowDom:
thanks @cuongdo for the fix!

sparrowDom:
@micahalcorn this is ready to be reviewed, all the decisions from [the document](https://docs.google.com/document/d/1jSMlcqJqUQ7_EZQHJwV2mu9BdTii-29annKQ70BDHpA/edit) have been implemented.

sparrowDom:
ahh 🤦‍♂️ @micahalcorn sorry I must have missed your comment for the changes to not hold up the merge...

micahalcorn:
@sparrowDom haha, amazing 🏆 I just didn't want it to sit around for days and days.

PR #1109, opened 2018-12-19
Remove token CLI

cuongdo:
The functionality provided by the token CLI has now been subsumed by the
Gnosis multi-sig wallet, Etherscan, and other tools. So, I'm removing
this now dead code as well as some duplicate code.

As part of this, `origin-contracts/test-alt/*Lib.js` is moving to
`origin-faucet`, because the latter is the only user of the tested
functionality. Later, we will probably extract `lib/token.js` into another
package, but that's another battle for another day. 😄 

### Checklist:


joshfraser:
Yay for deleting code

origin-faucet/test/lib-token.js
```diff
@@ -0,0 +1,46 @@
+import assert from 'assert'
+import helper from './_helper'
+import Token from '../lib/token'
+
+// These tests are for the token library that the token CLI uses. We don't
```
franckc: super nit: update comment to remove mention of the token cli. also in the test description on line 8 below.

Issue #1113, opened 2018-12-19
Move sample listings creation out of contracts migration scripts


franckc:
There are couple things that are wrong with the way we currently create sample listings:
 - It happens as part of a truffle contracts [migration script](https://github.com/OriginProtocol/origin/blob/master/origin-contracts/migrations/3_create_sample_listings.js). Samples creation as nothing to do with contracts so it should be moved out of migration. My suggestion would be into a separate script in charge of populating test data and that gets called as part of starting origin-js. Perhaps [here](https://github.com/OriginProtocol/origin/blob/master/origin-js/scripts/serve.js).
 - It calls contract code directly rather than using origin-js marketplace.createListing - this has some weird effect when testing with sample listing since some data is not populated correctly. For example one can not make an offer on a sample listing.

franckc:
CC @wanderingstan as a FYI since we were chatting about this a day or so ago.

tomlinton:
Just a few further related things:

- It'd be great if the hashes of the listings were not hardcoded (i.e. listing uploaded to IPFS, and returned hash is then used)
- It'd also be great if there were an option to use an external IPFS store rather than the one that gets created by [serve.js](https://github.com/OriginProtocol/origin/blob/master/origin-js/scripts/serve.js), perhaps configurable via environment variables. This might be a separate issue though.

nihar97:
https://github.com/OriginProtocol/origin/pull/1116 attempts to address this issue but breaks becuase of other issues.

cuongdo:
Note that the GraphQL rewrite already has part of this:

https://github.com/OriginProtocol/origin/blob/master/experimental/origin-admin/src/pages/accounts/_populate.js#L187

micahalcorn:
Oops, I'll let someone reopen this, if necessary 🙂

PR #1116, opened 2018-12-20
Move sample listings creation out of contracts migration scripts #1113

nihar97:

### Description:

### Checklist:

This code should change the place where sample data is populated as discussed in the issue. However, in an effort to not break abstraction barriers I have to use the marketplace resource. Since this resource uses the ES6 import statement, the system is broken by this update. Once we figure out how to babelify the serve.js file it should work as expected.


origin-js/scripts/serve.js
```diff
@@ -17,4 +19,141 @@ const start = async () => {
...
+    ipfsService,
+    affiliate: accounts[3],
+    arbitrator: accounts[4],
+    store,
+    blockEpoch: await web3.eth.getBlockNumber()
+  })
+
+  //hawai-house
+  await marketplace.createListing({
+    listingType: 'unit',
```
franckc: As opposed to hardcoding the listing data, how about using the [sample fixtures](https://github.com/OriginProtocol/origin/tree/master/origin-js/test/fixtures) ?

origin-js/scripts/serve.js
```diff
@@ -17,4 +19,141 @@ const start = async () => {
...
+    ipfsApiPort: '5002',
+    ipfsGatewayPort: '8080',
+    ipfsGatewayProtocol: 'http'
+  })
+  const store = new StoreMock()
+  store.set(
+    'notification_subscription_start',
+    new Date().getTime()
+  )
+  marketplace = new Marketplace({
```
franckc: As opposed to instantiating directly a Marketplace object, how about creating an Origin object which is the base for using the origin-js APIs ?
See an example [here](https://github.com/OriginProtocol/origin/blob/master/origin-tests/test/1-create-listing.js#L19).
Also see our (still work in progress) documentation [here](https://docs.originprotocol.com/#origin-object).

In order to import Origin, could you try [this](https://github.com/OriginProtocol/origin/blob/master/origin-tests/test/1-create-listing.js#L8) and see if that works ?

nihar97: Yes. I will try that. I initially did use an Origin object but I ran into the same problem of imports.

nihar97: I changed the code to use and Origin object and also to use sample fixtures. However, that style of importing origin still doesn't work

franckc: hmm... what error do you get exactly ?

nihar97: ` /app/origin-js/src/index.js:1
origin                  | 
(function (exports, require, module, __filename, __dirname) { import ContractService from './services/contract-service'
origin                  |                                                                      ^^^^^^^^^^^^^^^
origin                  |
origin                  | SyntaxError: Unexpected identifier`

tomlinton: You'll have to use `babel-node` to get support for ES6. E.g. `babel-node scripts/serve.js --preset @babel/env`

nihar97: hmm should I add babel-node to the dependencies?

franckc: Sgtm. Please try adding babel-node and see if that helps...

franckc:
This issue has been fixed as part of DApp2 rewrite. Closing.

PR #1127, opened 2018-12-22
Creator DApp integration

tomlinton:

### Description:

- Adds configuration to redux
- Adds component responsible for loading configs from external URLs if override passed in URL or via IPNS
- Moves base config to JavaScript file for DApp creator UI to import

### Checklist:


tomlinton:
@wanderingstan this is the PR I mentioned. It's mostly just the ideas and work you did for the prototype with a bit of added redux.

origin-dapp/src/components/customize.js
```diff
@@ -0,0 +1,105 @@
...
+      return configUrl
+    } else if (this.isWhiteLabelHostname()) {
+      // Retrieve the config from config.hostname
+      const ipfsGateway = `${process.env.IPFS_GATEWAY_PROTOCOL}://${process.env.IPFS_DOMAIN}:${process.env.IPFS_GATEWAY_PORT}`
+      return `${ipfsGateway}/ipns/config.${window.location.hostname}`
+    }
+  }
+
+  getConfigOverrideUrl () {
+    const configUrlMatch = window.location.search.match(/config=([^#]*)/)
```
wanderingstan: This was (on my part) a pretty hacky way to get the query string var and it's not very smart. I.e. if `\config=XXX` appears _anywhere_ in the URL it will get parsed. Is this open to abuse/security risk? The "right" way I guess would be to use something like:
https://www.npmjs.com/package/url-parse

origin-dapp/src/components/navbar.js
```diff
@@ -37,11 +48,20 @@ class NavBar extends Component {
...
+                {!isWhiteLabel &&
+                  <img
+                    src="images/origin-logo.svg"
+                    className="origin-logo"
+                    alt="Origin Protocol"
+                  />
+                }
+                {isWhiteLabel && logoUrl &&
+                  <img
+                    src={logoUrl}
```
wanderingstan: There was some discussion about limiting this to `ipfs://` urls (which are then routed to our ipfs gateway) to prevent user tracking. (`file://` URLs would also be ok)

Not a huge concern at first, I think. Worth doing?

tomlinton: I think for now we'll want to allow people to use their own external URLs. I think if we did the `ipfs://` only route we'd need to modify the UI to support uploads of images to our IPFS gateway, otherwise it is a lot of friction for users to manually upload their images to IPFS.

IPFS would be nicer and would avoid the problem if invalid/404 URLs and user tracking but it is something we can do post MVP.

origin-dapp/src/components/customize.js
tomlinton: It'd be cleaner though I think, this is a bit ugly. But it should be enough for now.

I was really surprised React router doesn't do its own URL parameter parsing!

Issue #1099, opened 2018-12-16
Editing listings with a picture


sparrowDom:
When editing a listing that has a preexisting picture and the user does not update the picture with a new one, continuing on listing details step fails with form errors (the step where user inputs title, description, picture etc)
<img width="609" alt="screenshot 2018-12-16 01 14 27" src="https://user-images.githubusercontent.com/579910/50048591-aef4ff00-00d0-11e9-938b-fcd097b81f9e.png">

micahalcorn:
Resolved by #1135 ✅

PR #1135, opened 2018-12-31
Natasha/edit listing images

Natasha08:
### Description:
[Issue #1099](https://github.com/OriginProtocol/origin/issues/1099)

- When editing a listing, the image is an ipfs url, which triggers an error when submitting the form (the form expects a data url.
- When the component mounts, if there are images for the listing, convert them to data urls and save that to state instead.

### Checklist:


origin-dapp/src/utils/fileUtils.js
```diff
@@ -78,7 +78,11 @@ export const scaleAndCropImage = (props) => {
 
     const scaledImage = loadImage.scale(canvas, newConfig)
     scaledImage.toBlob(async (blob) => {
-      blob.name = imageFileObj.name
+      if (imageFileObj.name) {
+        blob.name = imageFileObj.name
+      } else {
+        blob.name = Math.random().toString(36).substring(4)
```
Natasha08: The image url will not have a name

PR #1138, opened 2019-01-01
Phone Verification: Use English for India Phone Numbers

### LINKED ISSUE #1124: Change default language for SMS verification in India
The Twilio API guesses a language for SMS verification based on the country of the phone number. We got a report in Discord that for India it defaults to Hindi which a large part of the population cannot read so it'd be better if we default to English for India. We can explicitly set the language to English for Indian phone numbers.

screen shot 2018-12-22 at 11 35 01 am

BrunoBernardino:
### Description:

Fixes #1124.

Hindi is Twilio's default for Indian phone numbers, but according to a user report (and after some quick investigation), English appears to be preferred.

This includes a regression test.

I'd love to know what should I do to get this merged.

### Checklist:


origin-bridge/logic/attestation_service.py
```diff
@@ -81,6 +81,10 @@ def send_phone_verification(country_calling_code, phone, method, locale):
             # If a locale is not set Twilio will use a sensible default based on
             # the country of the telephone number
             params['locale'] = locale
+        elif country_calling_code == '91':
+            # Locale not provided, and calling India, so we want to set English
+            # instead of Hindi, which is Twilio's default there (#1124)
```
BrunoBernardino: I haven't seen this pattern here of referencing issues, not sure if it's acceptable.

origin-bridge/tests/services/test_attestation.py
```diff
@@ -194,6 +194,28 @@ def test_verify_phone_invalid_code():
     assert(validation_err.value.field_names[0]) == 'code'
 
 
+@mock.patch('logic.attestation_service.requests')
+@responses.activate
```
BrunoBernardino: I didn't see in other tests a way to investigate the request via `responses`, but if there's one I'd love to learn about it as it's probably better than mocking the request.

origin-bridge/tests/services/test_attestation.py
```diff
@@ -194,6 +194,28 @@ def test_verify_phone_invalid_code():
...
+    args = {
+        'country_calling_code': '91',
+        'phone': '12341234',
+        'method': 'sms',
+        'locale': None
+    }
+    response = VerificationService.send_phone_verification(**args)
+
+    assert len(mock_requests.post.call_args_list) == 1
+    assert mock_requests.post.call_args_list[0][1]['params']['locale'] == 'en'
```
BrunoBernardino: I didn't use `.get()` here as if we don't get these dict keys it's ok to fail the test.

micahalcorn:
Thanks @BrunoBernardino 💙 I've added you to the GitHub organization so that you can push a branch directly to the repo in the future instead of forking. I'll ask @tomlinton to review this.

tomlinton:
This looks great @BrunoBernardino, thanks! I'll test it out later today then merge it if everything looks good.

origin-bridge/logic/attestation_service.py
tomlinton: It's something I think we should do more of 👍

origin-bridge/tests/services/test_attestation.py
tomlinton: You should be able to do something like `responses.calls[0].request` to get the original request. This is fine, but feel free to refine it if you want.

PR #1143, opened 2019-01-02
Moving from identicon.js to MEW blockies

### LINKED ISSUE #841: Origin Identicon does not match MetaMask
When MetaMask is set to use Blockies for identicons, the resulting icon does not match Origin even though we are also using a similar looking identicon. 

image

A goal of Identicons is to know at glance if accounts are the same, so something is awry. 

- We are using [identicon.js](https://github.com/stewartlord/identicon.js/tree/master) (our [package.json](https://github.com/OriginProtocol/origin/blob/7de3e3f336b2d12a1a58314ea0b303e2133a5559/origin-dapp/package.json#L64))
- MetaMask is using some code called `blockies.js` ([their PR](https://github.com/MetaMask/metamask-extension/commit/1b89ceb63aa7d96912eb32c8766ef566479dde41))

Not sure what the right path is, or what other wallets are doing, but we should double check that we're following what the standard is, or is emerging to be.

BrunoBernardino:
### Description:

For consistency with Metamask.

Fixes #841 (currently our different implementations result in different identicons, which defeats their purpose).

### Checklist:


tomlinton:
I tested this out and it looks great, except it looks like MetaMask changed the default. Left some comments in #841.

Issue #1112, opened 2018-12-19
Create listing workflow - image upload issue


jordajm:
As described by Alessandro in Discord:

>I'm experiencing some weird behaviours uploading listings.
It's related, I think, to the pictures.
I got the same error of few days ago that I thought was related to some weird character in the description.
I cleaned the text in Brakets for this new accommodation.
I inserted the title, description, price and pics.
Clicked to proceed and I got the error "Please fix errors before continuing."

>I tried to put only the text without uploading any image, tried to proceed and no error.
Then I went back, uploaded the images, tried to proceed and I got the error again.
I tried to delete the last pic, proceed and it went on.
So I went back, uploaded again the last pic and this time no error at all.

Natasha08:
I have been unable to reproduce this for a brand new listing. I have DM'd  Alessandro

PR #1153, opened 2019-01-04
send form only dataURis

Natasha08:
### Description:
[Issue # 1112](https://github.com/OriginProtocol/origin/issues/1112)

**Problem:**
- While creating a listing, if a user uploaded two image or more and changed the order of the images, they would receive an error. This was because the form was submitted with an array of objects (with keys of `originalImageFile` and `croppedImageUri`. Form submission expects an array of dataURi strings.

**Solution:**
- Use the photo-picker method `picURIsOnly` to only get the dataURi for the images when `onFormDataChange` is called in `origin-dapp/src/components/listing-create.js`.

### Checklist:

### Manual Testing:
**Note** There is a current issue w/ react-overlays, and so to manually test you will most likely have to do the following in the command line `origin` root:
```
lerna clean
rm -rf node_modules
npm install
cd node_modules/react-bootstrap
npm install react-overlays@1.0.0-beta.17
cd ../..
npm start
```
1. Go to the create a listing flow, and fill out the required details (Title etc)
2. Upload at least two images.
3. Change the order of the images (they are draggable)
4. When you press `Continue`, you should not encounter errors.

**Note** When the issue was first reported, the user uploaded 10 images and sorted them.


sparrowDom:
@Natasha08 sorry didn't get to this today... will check it tomorrow if there is no hurry?

PR #1160, opened 2019-01-04
List Badges (Featured and Sold)

### LINKED ISSUE #1148: Add badges to listings
Add 'Featured', 'Sold' and 'Pending' badges to the listing search page:

BrunoBernardino:
### Description:

This solves the problem, but not completely (some badges are not easy to calculate just yet).

For now, some questions below in the comments, before I continue, or we merge this and create another issue.

Fixes #1148

### Checklist:

### Screenshots of the colors

screen shot 2019-01-04 at 19 15 08


experimental/origin-dapp2/src/pages/_ListingBadge.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+
+const ListingBadge = ({ status, featured }) => {
+  if (status === 'sold') {
+    return <div className="badge badge--sold">sold</div>
+  }
+  // TODO: Unavailable (multi unit, pending)
```
BrunoBernardino: Should I add multi unit support to GraphQL in this PR, or a separate one? I can add it all in this PR, but just wondering.

experimental/origin-dapp2/src/pages/_ListingBadge.js
```diff
@@ -0,0 +1,38 @@
...
+  // TODO: Pending (single unit, pending)
```
BrunoBernardino: I started out building the logic for this by looking into the events, but it seemed too much business logic in there. I think we should consider returning a `pending: Boolean` in the graphql as well. That should probably be its own PR and then this can be implemented.

experimental/origin-dapp2/src/pages/_ListingBadge.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export default ListingBadge
+
+require('react-styl')(`
+  .badge
+    background: gray
+    color: white
+
+    &--sold
```
BrunoBernardino: Are these colors OK? I only saw the featured one in the screenshots.

experimental/origin-dapp2/src/pages/listing/ListingDetail.js
```diff
@@ -16,7 +17,13 @@ class ListingDetail extends Component {
 
     return (
       <div className="listing-detail">
-        <div className="category">{listing.categoryStr}</div>
+        <div className="header">
```
BrunoBernardino: I'd prefer to abstract this `header` wrapper with the category and badge into its own component (instead of `_ListingBadge`). Is it something you'd be OK with, or are there any dragons I'm not aware of?

BrunoBernardino:
A couple of other things:

1. The UI was very brittle. For the admin and the dapp (but worse for the dapp). Not sure if I have something badly setup, but I had to refresh at least 2-3 times every time I made a change to get the results (I'd just get an "Error" message otherwise, apparently failing to connect to ganache — I'm using `ganache-cli`).

2. To test `featured` I added a checkbox to the form, grahpql, and the IPFS data function, but it still didn't save. In the end I realized it might be a separate PR. Is there anything I should look out for? Should I work on that, or did we intentionally leave it out?

experimental/origin-dapp2/src/pages/_ListingBadge.js
nick: Lets do that in a separate PR...

experimental/origin-dapp2/src/pages/_ListingBadge.js
nick: How about we expand the `status` field on a listing to include 'unavailable', 'pending', 'sold' and 'soldout'. Agree we can do this in a separate PR.

experimental/origin-dapp2/src/pages/_ListingBadge.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
```
nick: Lets put this in `src/components` as it'll be used in multiple pages. You can then `import ListingBadge from 'components/ListingBadge'` from any other page

experimental/origin-dapp2/src/pages/_ListingBadge.js
nick: I think pending should be `var(--dark)`

experimental/origin-dapp2/src/pages/listing/ListingDetail.js
nick: Hmm this looks fine as is to me... unless we're going to be using the component in more than 2 places I'm not sure it's worth abstracting?

nick:
1. Hmm I haven't seen that before. Are you using your own ganache, or the one that runs with the admin app?

2. Featured listings are currently handled by fetching a list of IDs from a GitHub branch. We'll probably want to change this at some point to be decentralized. For testing we can edit origin-graphql/src/resolvers/marketplace/_featuredAndHidden.js and add this under line 3:

`if (net === 'localhost') return [1]`

That should mark listing 1 as featured.

3. Since GraphQL is so fundamentally different from the way we do things in the existing dapp, we're going to start from a clean slate and not merge the two projects.

4. We'll deal with translation later

experimental/origin-dapp2/src/pages/_ListingBadge.js
micahalcorn: I think the intent was for `Sold`, `Sold Out`, `Unavailable`, and `Pending` to all be the same color (`--dark`). They equate to a similar UX, which is that a buyer is unable to make a new offer.

experimental/origin-dapp2/src/pages/_ListingBadge.js
BrunoBernardino: Ok, created #1162 for that.

experimental/origin-dapp2/src/pages/_ListingBadge.js
BrunoBernardino: I'd prefer that as well! Created #1163 for that.

experimental/origin-dapp2/src/pages/_ListingBadge.js
BrunoBernardino: Makes sense! Will make that change.

experimental/origin-dapp2/src/pages/listing/ListingDetail.js
BrunoBernardino: Makes sense, won't touch it further. 😊

BrunoBernardino:
1. I'm using my own ganache, I didn't know the app ran its own, maybe that's why I was having issues!
2. Nice, I'll try that and let you know if it doesn't work. 😄
3. Ok, so which one will be the "truth"? The dapp2 or dapp? I know that for example, identicon is different in both, and dapp2 is a regression, for example.
4. Sounds good!

BrunoBernardino:
@nick can I merge this or will you?

Issue #1161, opened 2019-01-04
My-purchases on staging


sparrowDom:
There is an error when I open my purchases on staging. [This line](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/resources/marketplace.js#L69) seems to be failing with an error that `offerArr` is not interable. What is weird is that I can not reproduce this locally on Rinkeby with ipfs pointing to staging ipfs. 

Problem on staging: 
<img width="1000" alt="screenshot 2019-01-04 11 33 11" src="https://user-images.githubusercontent.com/579910/50695164-2e288380-103c-11e9-808b-afec74444ef0.png">

Can not reproduce locally: 
<img width="1000" alt="screenshot 2019-01-04 11 33 02" src="https://user-images.githubusercontent.com/579910/50695167-31237400-103c-11e9-9864-e94fb1ae0d8c.png">

**PR to solve the issue:** https://github.com/OriginProtocol/origin/pull/1165/files

sparrowDom:
Figured it out, the problem seems to be in performance mode. Debugging it....

sparrowDom:
Solved via https://github.com/OriginProtocol/origin/pull/1165

PR #1165, opened 2019-01-04
add offers to listings in performance mode

sparrowDom:
### Description:
Add offers to listings when running  performance mode

**Solves issue:** https://github.com/OriginProtocol/origin/issues/1161


origin-js/src/resources/marketplace.js
```diff
@@ -183,25 +201,23 @@ export default class Marketplace {
...
-      return listing
-    }
-
     if (this.perfModeEnabled) {
       // In performance mode, fetch data from the discovery back-end to reduce latency.
       let listing = await this.discoveryService.getListing(listingId, blockInfo)
-      if (loadOffers)
-        listing = await addOffersToListing(listing)
+      if (loadOffers){
+        listing.id = listingId
```
franckc: Why do we set listing.id only when loadoffers option is on ?
Also, why do we actually need to set this is since [initFromDiscovery](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/models/listing.js#L75) should set it ?

sparrowDom: We need to set it because `_addOffersToListing` requires `listingId` in order to be able to query offers for a listing.

I see how code is currently confusing and not that much readable. Will add listingId as a function parameter to `_addOffersToListing` function to improve code readability.

sparrowDom:
@franckc I've made the code a bit more readable (hopefully 😇) and added some tests for fetching offers alongside listings.

sparrowDom:
@franckc actually never mind the above comment. When writing tests I have come across some inconsistencies we have in the code base as to how offers are represented in the listing object returned by the origin-js. 

Calling [getListing](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/contractInterface/marketplace/v00_adapter.js#L263-L291) function an object is returned with `offers` property of type object where key is offer id and value is and object consisting of offer status + offer event. Example: 

```
{ 0: {
  event: {
    address: "0x4c22B592b68fb46E579F146b9e2dEA7fF5f9CaB7"
    blockHash: "0xf0ae06839f27189c840ac7be381c29da9167a7ed82716f9eb081c9c3689b1ea7"
    blockNumber: 546
    event: "OfferCreated"
    id: "log_d7821247"
    logIndex: 0
    raw: {data: "0x1907247fba4c4e6b3f5fc8bc8508e491e7fec6f50146e83efd33c47c722a5cc1", topics: Array(4)}
    returnValues: Result {0: "0x129CB381163dB4bC7ad4927d0d1EC88185C3EC91", 1: "5", 2: "0", 3: "0x1907247fba4c4e6b3f5fc8bc8508e491e7fec6f50146e83efd33c47c722a5cc1", party: "0x129CB381163dB4bC7ad4927d0d1EC88185C3EC91", listingID: "5", offerID: "0", ipfsHash: "0x1907247fba4c4e6b3f5fc8bc8508e491e7fec6f50146e83efd33c47c722a5cc1"}
    signature: "0x6ee68cb753f284cf771c1a32c236d7ffcab6011345186a30e57837d761e86837"
    transactionHash: "0x017e8d342b79d9bbe26af590d1c38a44c0ab555c0dc027cf365f0599210fa23d"
    transactionIndex: 0
    type: "mined"
    },
  status: "created"
}}
```

The above offer object is structured quite differently when calling [getOffer](https://github.com/OriginProtocol/origin/blob/master/origin-js/src/contractInterface/marketplace/v00_adapter.js#L458)... example: 
```
{
  0: "20000000000000000"
  1: "12000000000000000000"
  2: "0"
  3: "0x0000000000000000000000000000000000000000"
  4: "0x129CB381163dB4bC7ad4927d0d1EC88185C3EC91"
  5: "0x821aEa9a577a9b44299B9c15c88cf3087F3b5544"
  6: "0x0d1d4e623D10F9FBA5Db95830F7d3839406C6AF2"
  7: "31536000"
  8: "1"
  affiliate: "0x821aEa9a577a9b44299B9c15c88cf3087F3b5544"
  arbitrator: "0x0d1d4e623D10F9FBA5Db95830F7d3839406C6AF2"
  blockNumber: 546
  buyer: "0x129CB381163dB4bC7ad4927d0d1EC88185C3EC91"
  commission: "12000000000000000000"
  createdAt: 1546739653
  currency: "0x0000000000000000000000000000000000000000"
  events: [{…}]
  finalizes: "31536000"
  ipfsHash: "0x1907247fba4c4e6b3f5fc8bc8508e491e7fec6f50146e83efd33c47c722a5cc1"
  logIndex: 0
  refund: "0"
  status: "created"
  value: "20000000000000000"
}
```

I think representing `offers` so differently might be very confusing. Also I am leaning towards changing the `offers` property of a listing from object to array, seems more natural. Any thoughts on this?

Should we find a solution to unify this somehow?

origin-js/src/contractInterface/marketplace/v00_adapter.js
```diff
@@ -276,19 +276,19 @@ class V00_MarkeplaceAdapter {
       } else if (event.event === 'ListingWithdrawn') {
         status = 'inactive'
       } else if (event.event === 'OfferCreated') {
-        offers[event.returnValues.offerID] = { status: 'created', event }
+        offers.push({ status: 'created', event })
```
franckc: Couple comments:
- There can be multiple events for the same offerId (e.g. for each state of created, accepted, finalized, etc...). So adding all events to the list may not be what we want to do ? The dictionary was taking care of only keeping the latest event for each offerId.
 - How will the DApp get the list of offerIds associated with the listing ?
 - Also have you checked to make sure the DApp is not relying on doing look ups by offerId in the dict or relying on iterating over a dict rather than a list of offers ?

Overall we may want to make this change as a separate PR. Or punt on it since the graphQL rewrite will clean up a lot of the origin-js interface. Happy to discuss - feel free to ping me so that we can come up to a quick resolution on this and launch multi-units ! :)

sparrowDom:
Thanks for the good points @franckc. Lets talk about this on our today's regular call? Currently I have left the (previously present) inconsistencies in, and added a comment explaining them. 

The Dapp functions ok, and no APIs have changed. So we can deploy the `multi-unit` fixes, and perhaps solve this in another PR if need be - like you suggested.

Issue #1106, opened 2018-12-17
Impossible Seller-Initiated Dispute Invitation


micahalcorn:
We currently prompt the user to initiate a dispute if she starts to leave a poor rating (< 3️⃣⭐️). This isn't possible or relevant for the seller since the funds have already been released. The prompt should be conditioned based on the state of the offer.

bleushark1988:
Hi. I would like to fix this issues. Can you detail for work?

micahalcorn:
This is temporarily negated by https://github.com/OriginProtocol/origin/issues/1723 and hopefully will not resurface 🙏

PR #1195, opened 2019-01-08
Disabled impossible seller-initiated dispute invitation (#1106)

bshevchenko:


micahalcorn:
Thanks for this @bshevchenko 🙂 It appears that this implementation would cause an issue where the "next step" form would not appear if this component loads before some of the state is available (loaded directly instead of navigated to). Also, I think it would be preferable to condition the dispute invitation based on the offer state rather than just the wallet address.

Since our GraphQL rewrite is well underway and this is likely to get cleaned up there, I'm going to close this for now. Thanks again for your interest in Origin. 💙

PR #1201, opened 2019-01-09
Identity V2 implementation

franckc:
### Description:
Implement Identity V2. Notes:
 - Profile/attestation data is stored as a single blob in IPFS, using the schemas defined in [this doc](https://docs.google.com/document/d/1PYqe1ggL60xCQqX7C2gj0qZZeEsI0aqcd7cWvtenVuc/edit#).
 - No data is stored on the blockchain but events are emitted using [this contract](https://github.com/OriginProtocol/origin/blob/master/origin-contracts/contracts/identity/IdentityEvents.sol).
 - Any new identity created from now on uses Identity V2
 - Backward compatibility is preserved for reading legacy identities
 - If a legacy identity is edited, the changes will be saved under the new identity format (and items that were not edited will be resetted).

### Checklist:


franckc:
@tomlinton Nick reviewed the JS side of things. Do you mind looking at the Python side ? Thanks in advance - sorry it is such a big PR :(

tomlinton:
Will do @franckc! I think @micahalcorn is our resident Python expert/fan though?

origin-bridge/api/modules/attestations.py
```diff
@@ -1,10 +1,66 @@
...
+
+
+class AttestationData(Schema):
+    issuer = fields.Nested(Issuer)
+    issueDate = fields.Str()
+    attestation = fields.Nested(Attestation)
+
+
+class Signature(Schema):
+    bytes = fields.Str()
```
tomlinton: Both `bytes` and `hash` are reserved words in Python. I don't believe it will cause an issue but generally I try to avoid using reserved words.

origin-bridge/dev.env
```diff
@@ -29,6 +29,7 @@ TWITTER_CONSUMER_KEY=
 TWITTER_CONSUMER_SECRET=
 
 ATTESTATION_SIGNING_KEY=
+ATTESTATION_ACCOUNT
```
tomlinton: I'm not sure of the consequences of leaving off the `=` here, it might be nothing but it still might be good to add it.

origin-bridge/logic/attestation_service.py
```diff
@@ -166,27 +170,40 @@ def verify_phone(country_calling_code, phone, code, eth_address):
...
-            # TODO: determine claim type integer code for phone verification
-            signature = attestations.generate_signature(
-                signing_key, eth_address, TOPICS['phone'], data
-            )
+            data = {
+                 'issuer': ISSUER,
+                 'issueDate': current_time(),
+                 'attestation': {
+                     'verificationMethod': {
+                         'phone': True
```
tomlinton: Would it be worthwhile to record the verification method as `call` or `sms` rather than `phone`?

tomlinton:
Looks good @franckc! I noticed you switched a bunch of Python variables from underscored to camel cased. I always prefer underscored in Python but I think since we are going to replace this reasonably soon its no problem. Looking forward to the new identity!

origin-bridge/dev.env
franckc: Good eye. Fixed.

origin-bridge/logic/attestation_service.py
franckc: Good idea. Added.

origin-bridge/api/modules/attestations.py
franckc: Yes, that's good practice. Not causing an issue so will leave as is for now to keep code simple.

franckc:
@sparrowDom Looks like the integration tests are failing and the error seems legit (failure to create the query)... Can you take a look ?

sparrowDom:
@franckc the last comment was probably referring to this https://github.com/OriginProtocol/origin/pull/1205 PR?

sparrowDom:
yeah and they all look the same :D

Issue #717, opened 2018-10-10
Edit profile form doesn't pre-populate with saved data


jordajm:
Steps to reproduce:
1) Fill out name, description, etc
2) Publish profile
3) Refresh page
4) Click edit profile button

Notice that edit profile form isn't pre-populated with saved data

Natasha08:
When I open the Edit Profile Modal, and upload a photo, the photo never displays in the Edit Profile Modal (this is before publishing).

micahalcorn:
No longer an issue ☑️

PR #1207, opened 2019-01-09
fix: Edit profile form doesn't pre-populate with saved data

HighQualityCode:
Update ComponentDidUpdate to getDerivedStateFromProps function.

### DIFF
--- a/origin-dapp/src/pages/profile/EditProfile.js
+++ b/origin-dapp/src/pages/profile/EditProfile.js
@@ -36,18 +36,28 @@ class EditProfile extends Component {
     this.handleDescriptionChange = this.handleDescriptionChange.bind(this)
   }
 
-  componentDidUpdate(prevProps) {
-    const { data } = this.props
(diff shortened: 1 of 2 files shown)


micahalcorn:
Fix for https://github.com/OriginProtocol/origin/issues/717 ☝️

Natasha08:
@HighQualityCode great job! Manually testing looks 👍. Just one issue:
- When I open the edit profile modal, the input no longer focuses on the 'First Name' field.

Also @micahalcorn the warning below does not seem accurate on the profile page any longer. When I navigate away and return with unsaved changes, I still have them. I only lose them when I reload (and there is a separate warning for that). Seems like it should be a message to encourage publishing before navigating away.
image

tomlinton:
Thanks for your contribution @HighQualityCode, we totally understand it is not always easy to contribute to open source projects. We'll let one of our open source contributors or team members finish this one off for you.

Issue #1086, opened 2018-12-13
Create script for pinning ipfs data using ipfs-cluster API


tomlinton:
We need a script that can process a [listing json](https://github.com/OriginProtocol/origin/blob/master/origin-js/test/fixtures/listing-valid.json) and extract any IPFS hashes, and then use a REST API provided by `ipfs-cluster` to pin the hashes. The documentation for a `ipfs-cluster` API client written in Go is available [here](https://godoc.org/github.com/ipfs/ipfs-cluster/api/rest/client) but we will need a JavaScript implementation that supports the `pin` and `unpin` methods. Some API documentation for `ipfs-cluster` is available [here](https://cluster.ipfs.io/developer/api/).

The script will run as a [Google Cloud Function](https://cloud.google.com/functions/) and be passed JSON data with IPFS hashes from the `event-listener` via [Google Cloud Pub/Sub](https://cloud.google.com/pubsub/docs/overview). We have an example of a similar function [here](https://github.com/OriginProtocol/origin/tree/master/deployment/utils/gcb-discord-notifications) which does Discord announcements from Google Cloud Build events. This might help someone who wants to take this on.

Related discussion #912.

franckc:
@mouazq Would you be interested in working on this issue ? It would be pretty impactful since it will allow us to pin (e.g. "replicate") content that gets uploaded to our IPFS cluster into a separate 3rd party cluster that the Protocol Labs team made available for us.... Let us know :)

rohitverma007:
Hello, new here, would love to contribute and work on this issue :)

rohitverma007:
@tomlinton just a few questions before I begin implementing.
You mentioned "any" IPFS hashes. So, it is possible to have empty media arrays? If the url does exist, will it always be a ipfs location or can it be of other format? I saw references to allowing `data` and `dweb` urls as well.

For the script it self, will the HTTP endpoint of the cluster (e.g /ip4/127.0.0.1/tcp/5001) be an input to the script? If not, is there a default endpoint to be assumed?

DanielVF:
We currently allow data urls in user profiles, but in listings, the media urls should only be IPFS.

tomlinton:
Also on the URL front, it uses HTTP basic auth so there should be env vars for username/password. Thanks for your work on this, looking great so far!

mouazq:
@rohitverma007 

Did you work on any part yet ? we can pair program it if you want ?

rohitverma007:
[...]

PR #1215, opened 2019-01-10
Script for pinning ipfs data 

rohitverma007:
### Description:

Implements #1086 

### Checklist:


deployment/utils/gcf-pin-ipfs-hash/index.js
```diff
@@ -0,0 +1,57 @@
...
+    : null;
+
+  let pinnedHashes = [];
+  let unPinnedHashes = [];
+
+  if (!Object.is(data, null)) {
+    const hashesToPin = parseIncomingData(data);
+    for(let i = 0; i < hashesToPin.length; i++){
+      hashToPin = hashesToPin[i];
+      pinned = await ipfsClusterApiService.pin(hashToPin);
```
franckc: Should we have a reasonable number of retries before giving up ? Perhaps with some exponential delay.

deployment/utils/gcf-pin-ipfs-hash/index.js
```diff
@@ -0,0 +1,57 @@
...
+  } else {
+    console.log("Error retreiving listing data");
+  }
+  console.log("Pinned following hashes: ", pinnedHashes.join(", "), "\n");
+  console.log("Could not pin following hashes: ", unPinnedHashes.join(", "), "\n");
+};
+
+const parseIncomingData = data => {
+  let hashesToPin = [];
+  if(data.type == "listing"){
```
franckc: super nit: in our code base we try to always use the === operator.
Same comment applies to the other places where == is used in this PR.

tomlinton:
Looks great @rohitverma007. Nothing to add aside from what @franckc has said.

I'm happy with where it lives along side the other Google Cloud function we have. The functionality is a bit niche and I don't think we want a full blown package. Also GCF needs this specific non-package structure if I remember correctly.

rohitverma007:
@tomlinton @franckc 

I forgot to mention one key detail I just remembered as I was trying to implement the retry feature 😅

So, the `pin` method of the API simply sends the request to pin and returns a `true` or `false` if the request is accepted, and that's all I am checking to flag the pinning as a success: 

https://github.com/OriginProtocol/origin/blob/5fb19090e684c6f503b8ebd2e55c13377163d537/deployment/utils/gcf-pin-ipfs-hash/index.js#L33-L34

But, this doesn't mean the pinning was actually successful. There's a few other steps that the ipfs cluster service does to pin the data, the steps are outlined here: https://cluster.ipfs.io/documentation/internals/#pinning-an-item 

Referring to the link above, I am currently only returning after the 4th step highlighted in italics. So, right now I'm assuming all will be fine after the pinning request has been accepted.

The ideal way to check if the pinning was successful would be to call the `status` endpoint after pinning and checking if it returns a `pinned` status. The complexity with doing that is we don't know how long it will take to successfully pin. One solution would be something similar to how the retries would be implemented, to ping the `status` api a few times with increasing intervals and only flag it as pinned once a successful pinned status is returned. 

Let me know how that sounds and if you have any suggestions.

franckc:
Hi @rohitverma007 - thanks for these additional details regarding pinning status. Very useful.

I'd be inclined to keep things simple for now and just test that the request was successful but not call the status api. Though maybe add a comment so that i the future we can come back to it and will have the context.

I could anticipate that as an extra measure in the future we add a separate "verification" cron script that runs periodically and rescans our entire corpus of listings/offers/users to verify their associated IPFS hashes are pinned. And that script would call the status endpoint.

Thanks again for being so detail oriented !

rohitverma007:
@franckc 
Sounds good, and yea a separate cron script would be ideal in the future.

Should I still implement retrying the request? I imagine the main points of failure would actually be after  sending a successful request and before a successful pin. From the manual testing that I have done so far, one of the main reasons sending the request failed was due to an incorrect ipfshash and there isn't really a point retrying a request with an incorrect hash. Let me know your thoughts.

franckc:
@rohitverma007 I think would be worth having a simple retry in the script to guard against the case where the IPFS server might be shortly down for maintenance or there is a network blip.

rohitverma007:
> @rohitverma007 I think would be worth having a simple retry in the script to guard against the case where the IPFS cluster might be shortly down for maintenance or there is a network blip.

Done and pushed. I set the number of retries to 5, 500ms initial interval and interval doubles every retry. Let me know if that's alright.

PR #1223, opened 2019-01-10
Messaging Account Detection

micahalcorn:
This solves an issue of the "messaging requirement" modal persisting while Orbit is being synced. Now we will start by querying the central server for whether or not the account is in the global registry. We will still listen for it to be enabled later once everything is synced.

### DIFF
--- a/origin-dapp/src/actions/Activation.js
+++ b/origin-dapp/src/actions/Activation.js
@@ -5,6 +5,8 @@ import { createSubscription } from 'utils/notifications'
 
 import origin from '../services/origin'
 
+const MESSAGING_URL = process.env.MESSAGING_URL
+
 export const ActivationConstants = keyMirror(
   {
     MESSAGING_ENABLED: null,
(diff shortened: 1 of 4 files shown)


micahalcorn:
This basically combines https://github.com/OriginProtocol/origin/pull/982 & https://github.com/OriginProtocol/origin/pull/1100.

micahalcorn:
@tomlinton this will need a DApp environment variable for the messaging server.

tomlinton:
> @tomlinton this will need a DApp environment variable for the messaging server.

MESSAGING_URL to the express server you created? e.g. `https://messaging-api.staging.originprotocol.com/`?

micahalcorn:
@tomlinton yes, we will need one for production too if it's not already up. This is basically a hotfix for the broken messaging requirement that is plaguing everyone.

tomlinton:
Also requires CORS to be enabled from DApp to that URL by the looks of it. The production one isn't up yet, I'll get the DNS setup.

micahalcorn:
@tomlinton I naively added the CORS piece to get it to work on my machine. It is basically the same as the rules that we had committed in the notifications server. Feel free to tighten it up. 🙏

PR #1224, opened 2019-01-10
Identity modal for upgrade

franckc:
### Description:
Add a modal on the profile page to prompt user that have an identity version 0 to upgrade to version 1. 

If the user accepts the modal, attestations in published and provisional state are cleared, as well as profile in published state. The user can then go thru the attestation creation flow again and publish - this will cause the profile and attestations to be saved using identity version 1.

If the user declines, the modal should not be showned until the profile is fetched again (due to a page reload or a metamask user change).

### Checklist:


origin-dapp/translations/all-messages.json
```diff
@@ -819,6 +819,10 @@
   "ConfirmPublish.afterYouPublishNotice": "After you publish your changes to the blockchain, other users will be able to see that you have verified the following, but they will not be able to see your actual data:",
   "ConfirmPublish.letsDoIt": "Let's do it!",
   "ConfirmPublish.oopsNoWait": "Oops, no wait...",
+  "ConfirmReset.resetIdentity": "Upgrade your Identity?",
```
franckc: @micahalcorn I came up with those copies. I'd appreciate you putting your PM hat on to help making those better / more explicit for the user :)

micahalcorn:
@franckc no chance that we can just say "click here to upgrade" and take all of the existing data and republish it? 😕

franckc:
I'm a total newbie regarding REACT/Redux so feel free to let me know if this is not the right approach for showing a modal...

franckc:
@micahalcorn Unfortunately we can't just republish the attestations as is since their format has changed (from text to now JSON blob). We would have to make requests to the attestation server to resign them in the new format. This would involve creating a new route on attestation server for that special case. That would add quite a bit of extra complexity... not sure worth it ? @joshfraser and @matthewliu were ok with blowing out the current identities since we have only a few users.

franckc:
@micahalcorn Following up no standup discussion. May I ask you to suggest a copy for the modal ? Your written skills are far better than mine :)

Also I wonder if this modal should only show on desktop and not mobile ? If yes, what's the best way to test if dapp is on mobile ?

micahalcorn:
@franckc on it 👍 Maybe not best, but established: https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/utils/mobile.js#L3-L17

PR #1225, opened 2019-01-11
New documentation system

DanielVF:
New docs at last!

- Switch from single page only State to Jekyll.
- Add sections for guides
- Add a page for each software component
- New look and navigation

image


DanielVF:
Micah, this can be merged whenever - it's now live.

PR #1314, opened 2019-01-24
Natasha/message arbitrator

Natasha08:
### Description:
- Message the arbitrator when a dispute has been initiated.

### Checklist:


Natasha08:
@nick I could use some feedback on the approach here. Also, is this supposed to cover giving the arbitrator access to buyer/seller conversations if they exist?

PR #1317, opened 2019-01-24
Messaging Requirement

micahalcorn:
This removes the requirement to run a server and enable messaging when there is not `MESSAGING_ACCOUNT` defined. As a result, [the instructions in the notifications server README](https://github.com/OriginProtocol/origin/tree/master/origin-notifications#the-not-as-easy-way) should now be sufficient.


Natasha08:
@micahalcorn [video](https://drive.google.com/file/d/1Au3M8dlYprCjxopFeZzYfF1U-DqZDb8Z/view) on your branch, not running messaging

PR #1320, opened 2019-01-24
No More Seller Onboarding

micahalcorn:
The seller onboarding modals continue to cause UX problems and will soon be phased out with a more elaborate user activation flow. This removes the existing modals and associated state now instead of waiting around for the rewrite.

😢


Natasha08:
😿

PR #1327, opened 2019-01-25
[Search] Harden scoring function

franckc:

### Description:

Default boostCommission.amount to 0 in case that field is missing.

See Elastic doc [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-field-value-factor).

### Checklist:


sparrowDom:
@franckc was thinking of including this change in yesterday's fix and it would make search more resilient. The "advantage" I worry we will loose is the information that something unexpected happened. 

Each listing model should return a `boostCommission` value and failing helped us discover the error. If it were to silently fail, we might not have caught it right away.

micahalcorn:
At first glance, I thought this PR held the secret to [this](https://www.theringer.com/nba/2019/1/23/18193249/james-harden-scoring-history-houston-rockets). ⛹️

franckc:
@sparrowDom Yes, good point. Though i usually like when software goes into a degraded but not completely broken mode in case of an error... e.g. return results even if ranking is not perfect.
So what do you think of adding a warning message in the logs in case the boostCommission is missing ? This way we'll more easily detect these kind of issues in the future ?

PR #1334, opened 2019-01-26
[Search] Disable indexing of jCal fields

franckc:
### Description:

The new listing.availability and offer.timeSlots fields use a complex jCal type.
This is causing issues with ElasticSearch dynamic mappings functionality.
Here is the scenario:
 - First doc of the corpus with an availability field gets indexed, Elastic decides on a mapping.
 - 2nd doc comes, it has an availability field but the structure is different -> Elastic fails indexing that document since the mapping has changed.

Since for now we do not need to support searching by availability we can simply disable indexing of those fields.

In terms of longer term solution, dynamic mapping is pretty dangerous and we should probably turn it off in favor of defining explicit mappings ourselves.

### Checklist:


sparrowDom:
sorry for missing this @franckc 😇

PR #1339, opened 2019-01-28
attempt to support dapp2 as a docker container

sparrowDom:
### Description:
attempt to support dapp2 as a docker container. The idea behind this is that a developer does not need to have 2 checkouts or origin repo (one running in origin-box and another with local `npm run` command). 

Afaik same resolved npm modules do not work in 2 different environments (OSX and linux docker image)

### DIFF
--- a/docker-compose.experimental.yaml
+++ b/docker-compose.experimental.yaml
@@ -11,7 +11,6 @@ services:
       - ./experimental:/app/experimental
     ports:
       - "3000:3000"
-    # npm run start --prefix experimental/origin-dapp2
     command:
       >
-      sleep infinity
+      npm run start --prefix experimental/origin-dapp2


sparrowDom:
@tomlinton was trying to create a container for `dapp2` but somehow dapp2 can not resolve `origin-services` that is in the same experimental folder. Any quick ideas on this? Otherwise we can cancel this PR.

sparrowDom:
Hmm that is weird. This is what I get: 
screenshot 2019-01-28 19 48 49

sparrowDom:
I also don't see `origin-services` inside node_modules of origin-dapp2: 
screenshot 2019-01-28 19 52 08

Any ideas how to debug this?

tomlinton:
Did you forget to rebuild the container maybe? `docker-compose build --no-cache origin`. That is weird.

sparrowDom:
I did rebuild it several times with the command you posted, still the same problem... 

Also tried a fresh git repo checkout and container build (with --no-cache) options and I get the same error on startup. 

Should we close this PR or do you think it might be useful to someone?

sparrowDom:
@tomlinton pushed a commit where for development a separate image can be built (for those who would need it) and containers can be brought up with `docker-compose -f docker-compose.experimental.yaml up`. So the image Travis uses stays intact. 

Overall it seems practical to me to be able to run the whole stack in docker, so there are no special local dependancies. If this is not going to work, I am just going to fallback to local. Which would mean running `postgres` on host machine and connect with `growth` / `discovery` servers to that machine. 

I get this running lerna ls (interestingly it recognises `origin-services`): 
screenshot 2019-01-30 16 06 06

Btw how do you get lerna command to run from shell? It only works for me if I enter it in `scripts` part of `package.json`?

tomlinton:
@sparrowDom I don't see any new commits, forget to push it? 😁

PR #1355, opened 2019-01-29
Growth engine updates

sparrowDom:
### Description:
Some updates to the growth engine: 
- Add DateTime scalar to graphQl
- Create boilerplate `getCampaigns` resolver
- add primary keys, enum definitions and table names to model files
- add models index.js file

### Checklist:


sparrowDom:
@franckc was creating a getCampaigns resolver and needed to modify model files a bit. Just a heads up, so we won't both be implementing the same thing.

origin-growth/src/apollo/db.js
```diff
@@ -0,0 +1,41 @@
...
+
+  return {
+    id: row.id,
+    name: row.name,
+    startDate: startDate,
+    endDate: endDate,
+    distributionDate: row.distributionDate,
+    status: status,
+    actions: [], // TODO implement 
+    rewardEarned: { currency: 'OGN', amount: 0 } // TODO implement
```
franckc: note that there is a campaign.currency field. so let's use that rather than hardcode OGN.

franckc: Also rewardEarned will be by user so it's probably a method that won't be implemented in db.js but in a different layer ? In work-in-progress PR #1359 I have a Campaign class capable of evaluating rules per user. So maybe it should end up there ?

experimental/origin-graphql/src/typeDefs/Growth.js
```diff
@@ -127,6 +128,8 @@ module.exports =
   type Mutation {
     invite(emails: [String!]!): InviteResponse
     enroll(campaignId: Int!): EnrollResponse
-    log(event: JSON!): LogResponse
+    gasForIdentity(walletAddress: ID!): SimpleResponse
```
franckc: Spec is not finalized but I think we may give users enough ETH to pay for gas not only for Identity but also to create a listing. So maybe we give this mutation a more generic name like "sponsoredGas" ?

origin-growth/src/apollo/db.js
sparrowDom: Good point removing that in a separate PR

experimental/origin-graphql/src/typeDefs/Growth.js
sparrowDom: Good point 👍

PR #1359, opened 2019-01-30
[WIP] Growth engine rules implementation

franckc:

### Description:

Implement business logic for the growth engine rules.

### Checklist:

### DIFF
--- a/origin-growth/src/rules/rules.js
+++ b/origin-growth/src/rules/rules.js
@@ -1,5 +1,5 @@
 const db = require('../models')
-const { GrowthEventTypes, GrowthEventStatuses } = require('../src/enums')
+const { GrowthEventTypes, GrowthEventStatuses } = require('../enums')
 
 // Upper cap for number of rewards per rule.
 const MAX_NUM_REWARDS_PER_RULE = 1000
@@ -15,15 +15,30 @@ class Reward {
 }
 
 
+/**
+ * TODO: include logic for
+ *  - start_date: Date at which campaign starts
+ *  - end_date: Date at which campaign ends assuming cap not exhausted
+ *  - cap
+ *  - cap_used
+ *
+ */
 class Campaign {
   constructor(campaign, config) {
     this.campaign = campaign
     this.config = config
 
+    if (!this.config.numLevels ||
+      !Number.isInteger(this.config.numLevels) ||
+      this.config.numLevels <= 0) {
+      throw new Error(`Campaign ${campaign.id}: config has invalid or missing numLevels field.`)
+    }
+    this.numLevels = this.config.numLevels
+
     this.levels = {}
     for (let i = 0; i < this.config.numLevels; i++) {
       if (!this.config.levels[i]) {
-        throw new Error(`Campaign ${this.campaign.id} missing level ${i}`)
+        throw new Error(`Campaign ${this.campaign.id}: config with missing level ${i}`)
       }
       this.levels[i] = new Level(this.campaign.id, i, this.config.levels[i])
     }
@@ -31,6 +46,7 @@ class Campaign {
 
   getEvents(ethAddress) {
     const whereClause = { ethAddress: ethAddress.toLowerCase() }
+    // TODO: filter out events out of the campaign time window.
     const events = db.GrowthEvent.findAll({
         where: whereClause,
         order: [ ['id', 'ASC'] ],
@@ -88,7 +104,7 @@ class Level {
   getRewards(ethAddress, events) {
     const rewards = {}
     this.rules.forEach(rule => {
-      rewards[rule.name] = rule.getRewards(ethAddress, events)
+      rewards[rule.id] = rule.getRewards(ethAddress, events)
     })
     return rewards
   }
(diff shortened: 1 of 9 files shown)


origin-growth/src/rules/rules.js
```diff
@@ -0,0 +1,350 @@
...
+    }
+    return level
+  }
+
+  /**
+   * Calculates rewards earned by the user.
+   * Only considers events that occurred during the campaign.
+   *
+   * @param {string} ethAddress - User's account.
+   * @returns {Promise<Object>} - Dictionary with rewards grouped first by level and second by rule.
```
franckc: @sparrowDom Let me know if this grouping makes sense to you. I thought grouping by level and by rule would make it easier for the UI to show rewards per level/rule. But I'm totally open to suggestion. For example, if that's easier for you, we could simply return a flat list of Reward objects.

franckc:
@sparrowDom This is ready for a first round of review. Thanks ! :)

origin-growth/src/rules/rules.js
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+  /**
+   * Returns number of rewards user qualifies for, taking into account the rule's limit.
+   * @param {string} ethAddress - User's account.
+   * @param {List<models.GrowthEvent>} events
+   * @returns {number}
+   * @private
+   */
+  _numRewards(ethAddress, events) {
```
sparrowDom: I guess this is down to taste... A more functional alternative to this function:
```
_numRewards(ethAddress, events) {
  const tally = this._tallyEvents(ethAddress, this.eventTypes, events)
  if (Object.values(tally).length < this.numEventsRequired)
    return 0

 return Math.min(this.limit, ...Object.values(tally) // keep only event counts
    .sort((first, second) => second - count)  // order them by count descending
    .slice(0, this.numEventsRequired)  // keep only the first this.numEventsRequired entries
 )}  
```

origin-growth/src/rules/rules.js
sparrowDom: Thanks for asking :) Both are good, but if I can choose I would prefer the flat list.

franckc: Done

origin-growth/src/rules/rules.js
```diff
@@ -0,0 +1,356 @@
...
+}
+
+
+class Level {
+  constructor(campaignId, levelId, config) {
+    this.campaignId = campaignId
+    this.id = levelId
+    this.config = config
+
+    this.rules =[]
```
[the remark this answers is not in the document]
franckc: Ah yes. cleaner. Done.

origin-growth/src/rules/rules.js
```diff
@@ -0,0 +1,356 @@
...
+
+  /**
+   * Returns number of rewards user qualifies for, taking into account the rule's limit.
+   * @param {string} ethAddress - User's account.
+   * @param {List<models.GrowthEvent>} events
+   * @returns {number}
+   * @private
+   */
+  _numRewards(ethAddress, events) {
+    const tally = this._tallyEvents(ethAddress, this.eventTypes, events)
```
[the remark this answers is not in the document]
franckc: Added comment and changed to suggested code.

origin-growth/src/rules/rules.js
```diff
@@ -0,0 +1,356 @@
...
+
+  /**
+   * Calculates if the rule passes.
+   * @param {string} ethAddress - User's account.
+   * @param {List<models.GrowthEvent>} events
+   * @returns {boolean}
+   */
+  evaluate(ethAddress, events) {
+    const tally = this._tallyEvents(ethAddress, this.eventTypes, events)
+    return (Object.keys(tally).length === 1)
```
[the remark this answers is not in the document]
franckc: Defensive coding is always good ! :) Added.

origin-growth/src/rules/rules.js
franckc: Not certain the alternative work. Any combination of events can make a tuple.
Let's say numEventsRequired=2 and we get the following tally:
{
  'A': 10,
  'B': 5,
  'C': 2,
   'D': 1
}

The propose change would return 5.

But the correct result is 8 since we can make up to 8 pairs. For example:
5 * (A, B)
2 * (A, C)
1 * (A, D)

franckc:
Yes, you are right we'll need additional helper functions.

Also, I was thinking this code would only be used for the current active campaign.
For past campaigns, the data should be pulled from the DB table, more precisely from the growth_reward table. So I think we might need a higher level abstraction that stitches data from past and current campaigns.

If you are ok with that, I'd prefer doing all this as separate PR so that we keep PR relatively small and easy to review ?

sparrowDom:
For the past campaigns reading the data in a different manner, is that for performance reasons?

franckc:
Regarding reading from growth_reward for past campaigns, it is true performance will be a bit better especially when we star having lots of campaign. But more importantly, I thought it would be important to reflect *exactly* what was paid out to the user rather than re-computing rules which might in some cases be slightly different (ex: we fixed a bug in a campaign rule or we manually reviewed an account activity and decided it was fraud and marked growth_reward rows with status Fraud while maybe the growth_event rows may not be marked as such ? Happy to chat further and we can decide.

sparrowDom:
Aha I see, good point. 👍

PR #1363, opened 2019-01-30
Extremely simple page titles

DanielVF:
@nick, does this approach look good for page titles? 

Seems to work in Chrome, Firefox, and Safari. Will be checking IOS mobile next.


DanielVF:
Titles work on Android Chrome.

PR #1366, opened 2019-01-30
Add "About Origin Tokens" page

DanielVF:
Add "About Origin Tokens" page.

Using content from old DAPP. (Changed image though. 😞 )

image


micahalcorn:
waving

DanielVF:
So I did. Fixed

PR #1370, opened 2019-01-31
Prevent Undefined String Exceptions

micahalcorn:
This solves two cases where undefined strings cause origin-js to throw uncaught exceptions. Both occur when the DApp is accessed without a wallet, and one causes an error message to appear intermittently in the UI. I'm not sure if these are the ideal places to handle the missing strings, but I suspect this will all change soon anyway.


micahalcorn:
error

origin-js/src/contractInterface/users/v01_adapter.js
```diff
@@ -122,7 +122,7 @@ export default class V01_UsersAdapter {
    * @return {Promise<string|null>}
    * @private
    */
-  async _getIdentityIpfsHash(address) {
+  async _getIdentityIpfsHash(address = '') {
```
franckc: As opposed to setting a default value in this method, I would
1. Add a safeguard that returns null if !address. e.g.
if (!address) {
  return null
}

2. Add a similar safeguard in the async get(address) method above at line 84:
if (!address) {
  return false
}

micahalcorn: @franckc if you wouldn't mind, I'm mid-other-branch right now dealing with mobile. I'll come back to this a bit later if you haven't gotten to it. Thanks!

franckc:
No problem ! Done...

PR #1371, opened 2019-01-31
Nick/husky

nick:
Add husky pre-push hook for linting checks


experimental/origin-dapp2/package.json
```diff
@@ -16,8 +16,9 @@
     "build:js": "NODE_ENV=production webpack --progress",
     "build:css": "node scripts/getCss > public/app.css",
     "build": "npm run build:css && npm run build:js",
-    "lint": "eslint .",
+    "lint": "eslint . && npm run prettier:check",
     "prettier": "find ./src -iname '*.js' | xargs ./node_modules/.bin/prettier --write",
+    "prettier:check": "find ./src -iname '*.js' | xargs ./node_modules/.bin/prettier -c",
```
tomlinton: Is there any reason to do it this way instead of `prettier -c **/*.js`?

tomlinton:
What do you think about doing a project wide prettier run? I guess it might be quite slow if we do it on every push

experimental/origin-dapp2/package.json
DanielVF: I've definitely been bit in JS projects with shell expansions that can't hold all the matching .js files.

tomlinton: From a shell it needs to be `"**/*.js"` to prevent shell expansion, but if it is in `package.json` it shouldn't matter

nick: Updated this 🙂

nick:
I'm down for running prettier on all packages... or at least the ones we'll be using going forward. Perhaps we can add them one at a time?

PR #1372, opened 2019-01-31
Google Analytics page tracking for DApp2

DanielVF:
Google Analytics page tracking for DApp2

Also, this ensure that we have a pre GraphQL query, top level page title for pages that will later change their title as the result of a GraphQL query (listings, offers). For discussion: sometimes GraphQL instantly returns cached responses, meaning that the final page title is sometimes the one recorded, rather than the pre-query one. Probably okay...


experimental/origin-dapp2/public/dev.html
```diff
@@ -33,5 +33,17 @@
     localStorage.ognNetwork = 'localhost'
     localStorage.metaMaskEnabled = 'true'
   </script>
+  <script async src="https://www.googletagmanager.com/gtag/js?id=UA-106384880-2"></script>
```
nick: I think we can remove from dev.html as that only gets used for local development

DanielVF: Makes sense. We can still check that the analytics code is working by looking at `window.dataLayer` variable.

DanielVF: I've removed this.

DanielVF:
Anything else needed on this one, Nick?

nick:
Nope I think it's good... looks like a few merge conflicts though. Will merge once those are resolved.

PR #1383, opened 2019-02-03
Graphql upgrade and some other stuff

mikeshultz:
### Description:

This PR is fairly trivial but could use a second set of eyes. 

- README fix
- Test fix
- Upgraded GraphQL

It would be good if someone could review and confirm the test change.  It's unrelated to the GraphQL upgrade since it was throwing before that change, but `expect().to.be.an()` apparently doesn't see class instances as Javascript objects, even though `typeof GraphQLJSON` would still return `object`.  This change seems more accurate to me, but GraphQL is new to me, so I'd like to be sure.

Also, the GraphQL upgrade required no changes to `origin-discovery` to pass tests, even with the breaking changes listed for their releases. The scalar value changes could impact users, but it looks like it would require bad behavior in the queries to begin with.

### Checklist:

### DIFF
--- a/origin-discovery/README.md
+++ b/origin-discovery/README.md
@@ -22,7 +22,7 @@ Start the origin-discovery container.
 
     docker-compose up origin-discovery
 
-The server should start and you can point your browser to http://localhost:4000 to access the GraphQL playground.
+The server should start and you can point your browser to http://localhost:4000/graphql to access the GraphQL playground.
 
 ## Running tests
 
--- a/origin-discovery/package.json
+++ b/origin-discovery/package.json
@@ -22,7 +22,7 @@
     "envkey": "^1.2.4",
     "express": "^4.16.4",
     "express-prom-bundle": "^4.2.1",
-    "graphql": "^0.13.2",
+    "graphql": "^14.1.1",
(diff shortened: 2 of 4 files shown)


mikeshultz:
I ham-fisted some queries against an empty DB from the Playground, so sort of.   I can get some data in there and give it a better test.  I was mostly leaning on the unit tests to make sure things were working.

What's the best setup for manually testing this locally? Is there a set of test data sitting around somewhere, or should I just point the listener at mainnet?

franckc:
@mikeshultz Low ricj change so I merged your PR - let's validate on staging...

Regarding indexing production data, I usually point the listener at Mainnet.
Here are the [instructions to do so](https://github.com/OriginProtocol/origin/blob/master/origin-discovery/src/listener/README.md#configure-and-start-the-listener).

mikeshultz:
Looks good on my local, though I had to fix #1392 before the listener would process.  Ran some listing queries, including returning that `jsonb` data.

franckc:
Sweet. Thanks for testing your change @micahalcorn ! Appreciate it.

As a side note, we really ought to add better unit test to that discovery server... though it's going to drastically change  after the DApp2 work is done so probably not worth doing now.

Issue #1119, opened 2018-12-20
OriginToken migration needs to ensure that token contracts are paused


cuongdo:
Currently, the TokenMigration contract assumes that the source and destination token contracts are in a paused state but does not verify this. This can easily cause inconsistencies if transactions are done on either contract.

We should add safeguards to TokenMigration. This is a simple fix that I've already incorporated into an in-progress change to create the token migration script.

cuongdo:
note: this is audit finding 9

DanielVF:
@cuongdo Is this issue solved by the merged PR?

PR #1395, opened 2019-02-05
making sure tokens are paused when migrating a token (issue 1119)

yuval-weiss:
### Description:

As described in #1119, we need to make sure both the old token and the new token are paused during token migration.

### Checklist:

- added an internal function - `_migrateAccount`, which ensures both tokens are paused.
- added tests.


origin-contracts/contracts/token/TokenMigration.sol
```diff
@@ -67,5 +61,19 @@ contract TokenMigration is Ownable {
         emit MigrationFinished();
     }
 
+    // @dev Internal account migration function.
+    function _migrateAccount(address _holder) internal {
+        require(!migrated[_holder], "holder already migrated");
+        require(fromToken.paused(), "fromToken should be paused during migration");
```
cuongdo: I think that checking `paused()` should only be done in public functions such as `migrateAccounts()` and `migrateAccount()`. Reads of contract state cost gas when performed during a transaction. So, I'd rather incur that cost once per call to `migrateAccounts()` instead of once per migrated account.

origin-contracts/test-alt/TokenMigration.js
```diff
@@ -35,10 +35,39 @@ describe('TokenMigration.sol', async function() {
...
     })
+    await OldToken.methods.pause().send({from: owner})
+    await NewToken.methods.pause().send({from: owner})
+
     await NewToken.methods.transferOwnership(TokenMigration._address).send({from: owner})
     assert.equal(await TokenMigration.methods.finished().call(), false)
   })
 
+
+  it('disallow migrating when fromToken is not paused', async function() {
```
cuongdo: Thanks for adding this test. Could you also add an equivalent for ensuring that `toToken` must be paused?

origin-contracts/contracts/token/TokenMigration.sol
yuval-weiss: You're right, that was the reason I wanted to create the new internal function in the first place. The requires indeed should be in the public function. I'll fix it later today:)

yuval-weiss:
@cuongdo Pushed changes. also modified `assertRevert` to allow asserting a specific revert caused the error.

cuongdo:
@yuval-weiss I've added you as a collaborator for this repo. Could you push your branch to the origin repo, so that CI can run? When CI passes, I'll merge.

yuval-weiss:
@cuongdo cool:) pushed!

cuongdo:
Sorry for the inconvenience, but could you open a new PR with your new branch? That will enable PR to run. Pushing to branches on the origin repo then creating PRs based on those is our general development workflow. Thanks!

yuval-weiss:
@cuongdo no problem. https://github.com/OriginProtocol/origin/pull/1422

PR #1422, opened 2019-02-06
Ensure paused tokens migration issue 1119

yuval-weiss:
### Description:

As described in #1119 and discussed in PR #1395, we need to make sure both the old token and the new token are paused during token migration.

### Checklist:

* added an internal function - `_migrateAccount`, which ensures both tokens are paused.
* added tests.
* updated `assertRevert` testes utility function to be able to check for specific exception messages.


cuongdo:
Thanks @yuval-weiss! For future PRs, please prefix your branch name with your GitHub username. For example, the branch for this PR would be `yuval-weiss/ensure-paused-tokens-migration-issue-1119`.

PR #1425, opened 2019-02-07
Erc725 fix

hugooconnor:
### Description:

The current implementation of ERC725 in this repo does not allow a key to have multiple purposes.

This fix creates a new method to check that a key does not have a particular purpose.

This method is called in 'addKey'. Therefore, a user can call 'addKey' multiple times on the same key provided the purpose they submit is not in the array of existing purposes for that key.

### Checklist:

- [n/a] Wrap any displayed ETH addresses with [`formattedAddress`](https://github.com/OriginProtocol/origin/blob/master/origin-dapp/src/utils/user.js#L15-L17)
- [n/a] Wrap any new text/strings for translation
- [n/a] Run `npm run translations` if there are any changes to translated strings
- [n/a] Map any new environment variables with a default value in the Webpack config
- [n/a] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/origin-docs)

### DIFF
--- a/origin-contracts/contracts/identity/KeyHolderLibrary.sol
+++ b/origin-contracts/contracts/identity/KeyHolderLibrary.sol
@@ -73,9 +73,11 @@ library KeyHolderLibrary {
         public
         returns (bool success)
     {
-        require(_keyHolderData.keys[_key].key != _key, "Key already exists"); // Key should not already exist
+        // Check key purpose not established.
+        require(_hasNoMatchingPurpose(_keyHolderData.keys[_key], _purpose), "Key already has that purpose"); 
         if (msg.sender != address(this)) {
-            require(keyHasPurpose(_keyHolderData, keccak256(abi.encodePacked(msg.sender)), 1), "Sender does not have management key"); // Sender has MANAGEMENT_KEY
+          require(keyHasPurpose(_keyHolderData, keccak256(abi.encodePacked(msg.sender)), 1),
+                  "Sender does not have management key"); // Sender has MANAGEMENT_KEY
         }
 
         _keyHolderData.keys[_key].key = _key;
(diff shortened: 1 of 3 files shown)


origin-contracts/contracts/identity/KeyHolderLibrary.sol
```diff
@@ -73,9 +73,11 @@ library KeyHolderLibrary {
         public
         returns (bool success)
     {
-        require(_keyHolderData.keys[_key].key != _key, "Key already exists"); // Key should not already exist
+        // Check key purpose not established.
+        require(_hasNoMatchingPurpose(_keyHolderData.keys[_key], _purpose), "Key already has that purpose");
```
hugooconnor: This causes

origin-contracts/contracts/identity/KeyHolderLibrary.sol
```diff
@@ -73,9 +73,11 @@ library KeyHolderLibrary {
         public
         returns (bool success)
     {
-        require(_keyHolderData.keys[_key].key != _key, "Key already exists"); // Key should not already exist
```
hugooconnor: This means that a key can only ever have a single purpose. If that is the intention, there is no need to store purposes in an array.

origin-contracts/contracts/identity/KeyHolderLibrary.sol
```diff
@@ -204,4 +208,13 @@ library KeyHolderLibrary {
         }
         return isThere;
     }
+
+    function _hasNoMatchingPurpose(Key storage key, uint purpose) public view returns (bool hasPurpose) {
```
hugooconnor: This method could be more efficient if the set of purposes is known ahead of time.

nick:
We're not actually using these contracts any more. ERC725 is currently undergoing an evolution... please see https://github.com/ERC725Alliance/erc725 for the latest.

PR #1432, opened 2019-02-07
Identity v2 schema validation

nick:
Add Identity v2 schema validation to dapp2


experimental/origin-graphql/src/mutations/identity/deployIdentity.js
```diff
@@ -10,24 +11,46 @@ async function deployIdentity(_, { from, profile = {}, attestations = [] }) {
...
+  )
+
+  data.attestations.forEach(attestation => {
+    validator(
+      'https://schema.originprotocol.com/attestation_1.0.0.json',
+      attestation
+    )
+  })
+
   const ipfsHash = await post(contracts.ipfsRPC, data)
```
franckc: It would be nice to do some extra validations before recording on-chain. Things like:
 - checking the profile address matches with the current web3 account
 - checking the signature of the attestation

Let's add a TODO so that we remember to harden this later ?

Also the current origin-js code loads the current attestations and attempts to merge with the new one.
I'm not positive we should continue doing that (would rather keep things simple) since I assume DApp2 guarantees that all attestations that were loaded from the profile page are passed to this mutation ?

nick: Just modified the commit so we do attestation signature verification both going into and coming out of IPFS, so should be covered now :-)

But still need to add some tests...

franckc: Sweet ! Thanks for adding that signature validation :)

PR #1442, opened 2019-02-08
Docker implementation for experimental packages

tomlinton:
- Adds docker support for dapp2

### Checklist:

- [x] Test your work and double-check to confirm that you didn't break anything~~


experimental/origin-dapp2/package.json
```diff
@@ -44,6 +44,7 @@
     "graphql-tag": "^2.10.1",
     "identicon.js": "^2.3.3",
     "ipfs-api": "^26.1.2",
+    "jest-docblock": "^22.2.0",
```
tomlinton: This was necessary because babel-plugin-fbt wasn't installing jest-docblock which led to module not found errors. This might have been resolved with the hoisting support for babel-plugin-fbt.

tomlinton: Just checked and this is still required even with that hoisting change. 😞

PR #1446, opened 2019-02-10
origin-mobile Android support

mikeshultz:
### Description:

Added support for Android to origin-mobile

### Checklist:

This PR includes new env vars that will be necessary for deployment: 

origin-linking:
- `FIREBASE_SERVICE_JSON` - The service account JSON file used for firebase auth (file should be included as well)
- `FIREBASE_DB_URL` - The URL to the firebase DB


tomlinton:
Thanks @mikeshultz. To support this it looks like we'll need to update the linking server with `FIREBASE_SERVICE_JSON` and `FIREBASE_DB_URL`. Do you have a service account or should we create a new one? Is there anything else I am missing that needs to be updated on the deployments?

mikeshultz:
@tomlinton, @micahalcorn should be able to get you that service account JSON file(or point you to where to download it).  I don't have access to the Origin account.  This file should be considered a secret.  It contains private keys for the service, IIRC.

The default `FIREBASE_DB_URL` could probably be leaned on for prod.  Not entirely sure if we want a separate DB/account for staging.  I don't think there should be any harm in using the same.

Those should be the only changes necessary for deploy.

tomlinton:
Added the necessary stuff to Kubernetes so this should be good to go from a deployment perspective.

PR #1460, opened 2019-02-13
Dapp2 deployment to dev

tomlinton:
Container and Kubernetes config to run DApp2 on dev using Kovan

### DIFF
--- a/deployment/cloudbuild-deploy.yaml
+++ b/deployment/cloudbuild-deploy.yaml
@@ -88,7 +88,7 @@ steps:
           --build-arg ENVKEY=$${_ENVKEY} \
           .
 
-  # Push the origin-dapp container image
+  # Push the container image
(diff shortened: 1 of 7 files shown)


franckc:
Cool ! one potential caveat is that DApp2 will share same discovery server (for search) than DApp1. So discovery will return both our local blockchain listings (created on DApp1) and Kovan listings (created on DApp2). But may not be an issue since afaik nobody is creatoing DApp1 listings on dev...

franckc:
Well actually it depends on how we configure the listener on dev. We could configure it to only run against Kovan network.

tomlinton:
Yea you are right @franckc, the Kovan setup is pointed at the staging discovery server. For now though I just wanted to get it up and going, and figure out the discovery stuff later.

Issue #715, opened 2018-10-10
Link To Social Profiles


micahalcorn:
Once https://github.com/OriginProtocol/origin-js/issues/609 is completed, verified badges for social profiles should link to the user's Facebook, Twitter, or Airbnb account.

drewmo:
I can tackle this. We are ignoring Facebook links currently right since it was not decided how to handle that in #951

PR #1472, opened 2019-02-14
715 link to social profiles

drewmo:
### Description:

Worked to solve issue #715 to provide users links to other verified user's social media profiles if they have one tied to their profile. Specifically if you can verify from commit [line 15 in b18e938](https://github.com/OriginProtocol/origin/commit/b18e938b2cd4db5cac5370a1eda8adaf058df16d#diff-44607467bff009858f37536fc9a4ff84R15) as I wasn't sure if it was a good idea to hard code the object reference to the `siteName` and `userId.raw`

### Checklist:

- [x] Run `npm run translations` if there are any changes to translated strings


experimental/origin-dapp2/src/pages/user/User.js
```diff
@@ -1,13 +1,24 @@
...
 import IdentityQuery from 'queries/Identity'
 import Reviews from 'components/Reviews'
 import Avatar from 'components/Avatar'
 
 import UserListings from './_UserListings'
 
+function parseUserSocialID(attestations) {
+  const socialUrlObject = {}
+  attestations.forEach(attestation => {
+    const socialData = JSON.parse(attestation)
```
franckc: I don't think it is necessary to parse. This should already be an object ?

experimental/origin-dapp2/src/pages/user/User.js
```diff
@@ -1,13 +1,24 @@
...
+    socialUrlObject[socialData.data.attestation.site.siteName] =
+      socialData.data.attestation.site.userId.raw
```
franckc: A few comments:
 - attestation.site is only valid for social networks. not for phone and email attestations.
 - even if attestation.site exists, it is not guaranteed attestation.site.userId is (it is only currently only populated for Twitter and Airbnb. not for Facebook for example).
 - finally if attestation.site.userId, raw field may not be presented (id may be stored encrypted).

As a result, in some cases an exception will get thrown in the forEach loop and cause its execution to stop, which will cause the socialUrlObject to not be fully populated...

I'd suggest checking for the presence of the fields before dereferencing them.
e.g. something like this:
if (socialData.data.attestation.site && socialData.data.attestation.site.userId && socialData.data.attestation.site.userId) {
   socialUrlObject[socialData.data.attestation.site.siteName] = socialData.data.attestation.site.userId.raw
}

experimental/origin-dapp2/src/pages/user/User.js
drewmo: It actually returns as a string from IPFS, at least when I pull it from the `profile` object. It still has all the object structure but was a string.

experimental/origin-dapp2/src/pages/user/User.js
drewmo: Will do, thanks for the feedback! I'll make sure to condition out the loop in case any new social media comes through, I'll see if I can figure out a way to do it without having to hard code the references to .site.useId for example. That was it won't need to be touched in the future.

experimental/origin-dapp2/src/pages/user/User.js
franckc: Ah... The graphQL client side probably returns this as a string rather than an Object.
Might be worth updating the GraphQL schema to return structured attestations data.
@nick any thoughts on this ?

experimental/origin-dapp2/src/pages/user/User.js
```diff
@@ -1,13 +1,29 @@
...
+function parseUserSocialID(attestations) {
+  const socialUrlObject = {}
+  attestations.forEach(attestation => {
+    const socialData = JSON.parse(attestation)
+    if (
+      socialData.data.attestation.site &&
+      socialData.data.attestation.site.userId
+    ) {
+      socialUrlObject[socialData.data.attestation.site.siteName] =
+        socialData.data.attestation.site.userId.raw
```
franckc: I think you also want to check if socialData.data.attestation.site.userId and socialData.data.attestation.site.userId.raw.

For facebook attestations for example, socialData.data.attestation.site.siteName exists but there is no socialData.data.attestation.site.userId.

For other social sites in the future, socialData.data.attestation.site.userId may existing but not the raw field (for example if user opted for keeping that field encrypted).

Sorry, this is a bit complicated... Here is a doc with some more details on the attestation schema to help you understand: https://docs.google.com/document/d/1PYqe1ggL60xCQqX7C2gj0qZZeEsI0aqcd7cWvtenVuc/edit?usp=sharing

franckc: In particular note the schema for an Attribute. Can have either a raw or hash field.

drewmo: I will review the google doc schema. No worries on the complication, I want to be sure it's going to work for the project correctly.

nick:
Going to close this for now as I think we should expose these as a field on GraphQL

PR #1473, opened 2019-02-14
[draft] Additional GraphQL mutation testing + gas used table

DanielVF:
- Add test for addFunds mutation
- Add test for disputeOffer mutation
- Throw error in makeOffer if an ERC20 currency is used, since the method does not currently support it
- Add table of gas prices as used in GraphQL mutation tests
- Change addFunds "amount" field to "value" and use ETH, to match the name and behavior of all other offer mutations.


DanielVF:
I'm going to break this PR into smaller, more focused PR's. No need to merge this.

nick:
Are we OK to close this now?

PR #1474, opened 2019-02-14
Fix errors and token whitelist migration

crazybuster:

### Description:

- This is to fix errors in page
-  token whitelist migration

### Checklist:

### DIFF
--- a/origin-contracts/migrations/16_whitelist_vA_token.js
+++ b/origin-contracts/migrations/16_whitelist_vA_token.js
@@ -1,0 +1,18 @@
+const OriginToken = artifacts.require('./token/OriginToken.sol')
+const VA_Marketplace = artifacts.require('./V00_Marketplace.sol')
+
+module.exports = function(deployer, network) {
+  return createTokenWhitelist(network)
+}
+
+async function createTokenWhitelist(network) {
+  const token = await OriginToken.deployed()
+  const tokenOwner = await token.owner()
+
+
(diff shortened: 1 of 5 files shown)


franckc:
Adding @cuongdo since he is the most familiar with the token whitelist.

origin-linking/src/logic/hot.js
```diff
@@ -13,7 +18,23 @@ class Hot {
     origin.contractService.transactionSigner = this.account.signTransaction
   }
 
+  isAllowedCall(cmd, params){
```
franckc: Can you add a comment explaining the minFee logic ? Not knowing much about this it is not clear why this is needed...

origin-contracts/migrations/16_whitelist_vA_token.js
```diff
@@ -0,0 +1,18 @@
+const OriginToken = artifacts.require('./token/OriginToken.sol')
+const VA_Marketplace = artifacts.require('./V00_Marketplace.sol')
+
+module.exports = function(deployer, network) {
+  return createTokenWhitelist(network)
```
franckc: Does this need to be conditional to not run on dev environment ?

franckc: Similar to condition used in 7_create_token_whitelist:

>   if (network !== 'development' || process.env['SIMULATE_MAINNET']) {

PR #1477, opened 2019-02-15
Prevent USD field from focussing when using tab

wanderingstan:
USD field is not editable, so no point having it get focussed when user is using `TAB` to navigate through form. 

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex


micahalcorn:
Should this be merged?

wanderingstan:
@micahalcorn Yes, I thought it had been. Running a round of integration tests and then its good to go!

PR #1490, opened 2019-02-18
Marketplace creator fixes

tomlinton:
Fixes various issues on marketplace creator:

- Client/server side validation for invalid subdomains
- Handle switching of web3 account mid-flow and also where someone loads a configuration from localstorage with a different web3 account than the one it was created it (thanks @kayyoo23)
- Added some more logging to server

### Checklist:

### DIFF
--- a/origin-dapp-creator-client/src/pages/Configure.js
+++ b/origin-dapp-creator-client/src/pages/Configure.js
@@ -37,6 +37,21 @@ class Configure extends React.Component {
     this.setState({ redirect: '/metamask' })
   }
 
+  componentDidMount() {
+    // Handle the case where filter by own listings is enabled but the Ethereum
+    // address is different to the currently active web3 account (which will
+    // populate the config.marketplacePublisher attribute)
+    if (
+      this.state.config.filters.listings.marketplacePublisher &&
+      this.state.config.filters.listings.marketplacePublisher !==
+        web3.eth.accounts[0]
+    ) {
+      this.setListingFilters({
+        marketplacePublisher: web3.eth.accounts[0]
+      })
+    }
+  }
+
   // Retrieve the category object from the filter value in the config
   getCategoryFromConfig() {
     if (!this.props.config.filters.listings.category) return null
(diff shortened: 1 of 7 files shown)


origin-dapp-creator-client/src/pages/Configure.js
```diff
@@ -37,6 +36,21 @@ class Configure extends React.Component {
     this.setState({ redirect: '/metamask' })
   }
 
+  componentDidMount() {
+    // Handle the case where filter by own listings is enabled but the Ethereum
+    // address is different to the currently active web3 account (which will
+    // populate the config.marketplacePublisher attribute)
+    if (
+      this.props.config.filters.listings.marketplacePublisher &&
+      this.props.config.filters.listings.marketplacePublisher !==
```
franckc: is it worth doing comparison in canonial lower case form ?

origin-dapp-creator-server/src/middleware.js
```diff
@@ -11,8 +11,19 @@ export async function validateSubdomain(req, res, next) {
   const { address, config } = req.body
 
   if (config.subdomain) {
+    // Check subdomain doesn't contain invalid characters
+
+    // eslint-disable-next-line no-useless-escape
+    const subdomainRe = /[^a-zA-Z0-9\-]/
```
franckc: Corner case: looks like subdomain can now contain non-ascii characters:
https://domainname.shop/faq?id=8&session=106ee5e67d523298

But let's not worry about it for now.

origin-dapp-creator-server/src/middleware.js
```diff
@@ -11,8 +11,19 @@ export async function validateSubdomain(req, res, next) {
...
+    if (subdomainRe.test(config.subdomain)) {
+      return res.status(400).send('Subdomain contains invalid characters')
+    }
+
     // Check for subdomain blacklisting
     if (subdomainBlacklist.includes(config.subdomain.toLowerCase())) {
+      logger.warn(
```
franckc: nit: should this be an error rather than a warning ? same comment for similar logging below.

origin-dapp-creator-client/src/pages/Configure.js
tomlinton: Done, thanks!

origin-dapp-creator-server/src/middleware.js
tomlinton: I normally do warnings for things that are the users fault like this, but one below needs to be updated, good spot!

PR #1493, opened 2019-02-19
Run DB migration as part of Faucet deployment

franckc:
### Description:

Same as title

### Checklist:

### DIFF
--- a/deployment/dockerfiles/origin-faucet
+++ b/deployment/dockerfiles/origin-faucet
@@ -4,6 +4,12 @@ WORKDIR /app
 
 ENV NODE_ENV=production
 
+# Install envkey-source to make environment available for sequelize migration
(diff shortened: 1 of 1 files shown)


franckc:
Actually this is problematic since Faucet runs on staging and therefore those DB tables would be on staging... :( Since it is real ETH would be better to use the prod DB... 

A possible solution is that for the faucet, I could setup DATABASE_URL pointing to prod DB... a bit hacky but would work...

Alternatively we could setup the OGN token faucet on staging and the ETH faucet on prod under different domain. maybe faucet.staging.originprotocol.com and faucet.originprotocol.com ?

@tomlinton lemme know what u think...

tomlinton:
I think it makes sense to have the faucet on prod, I'll do that.

franckc:
Sounds, good. Having a prod faucet for ETH makes sense I agree. Thanks for the help on setting this up.

Thanks for the reminder on  _ENVKEY_FIELD !
I added it with value faucetEnvKey to the faucet trigger.

PR #1505, opened 2019-02-20
[ETH Faucet] Return txn hash right away rather than waiting for receipt

franckc:
### Description:
Waiting for a txn receipt on mainnet can take several minutes (it took over 2 min when I tested today)... This is not a great user experience since upon form submission, the use has to wait with no feedback as to what's happening.

Changes logic to return the txn hash right away, with a link to the pending transaction on etherscan so that the user can check there on the status.
In order to do this, I refactored the code to introduce a TxnManager which wraps calls to web3 for sending transaction, getting hash and receipt.

### Checklist:


franckc:
Good question @sparrowDom - I tested and even if the connection gets closed, Express finishes the execution of the route.

PR #1515, opened 2019-02-21
upgraded react-native to 0.59.0-rc.2

mikeshultz:
### Description:

This was necessary for Android 64-bit ARM support, related to this issue: 

https://github.com/facebook/react-native/issues/2814

This could probably use some iOS testing for verification.

### Checklist:


origin-mobile/src/components/transaction-item.js
```diff
@@ -75,7 +75,7 @@ class TransactionItem extends Component {
       <TouchableHighlight onPress={handlePress}>
         <View style={[ styles.listItem, styles.completed, style ]}>
           {!picture && <View style={{ ...styles.thumbnail, ...styles.imageless }} />}
-          {picture && <Image source={{ uri: picture.url || picture }} style={styles.thumbnail} />}
+          {picture !== '' && <Image source={{ uri: picture.url || picture }} style={styles.thumbnail} />}
```
micahalcorn: @mikeshultz did you find a problem resulting from this loose checking for a `picture` value?

micahalcorn: For example, if `picture` is `undefined` or `null` for some reason, won't we end up with both the `View` and the `Image`?

mikeshultz: Yeah, this was hasty, but it was to tackle an actual bug.  When the user had no stored avatar and wasn't setting one, `picture` was an empty string.  And for whatever reason RN was trying to render that empty string causing errors.  I don't really have a full picture of what its values could be, however.

How about an explicit boolean conversion, instead?  e.g. `Boolean(picture) && [...]`

mikeshultz:
Added explicit bool conversions.  Also, one of the checks was just wrong.  Thanks for pointing out this hasty change.

micahalcorn:
Great, thanks @mikeshultz 🙏 I'm actually about to add placeholder images in those cases, so that should take care of it too.

micahalcorn:
c560b74 was used for the latest TestFlight build (0.2.9). @crazybuster said I could merge this. 😬

PR #1517, opened 2019-02-21
Allow tests to be ran inside docker

sparrowDom:
### Description:

This allows tests to be run inside Dapp2 Docker container


sparrowDom:
that was fast, thanks @nick!

PR #1519, opened 2019-02-22
[Growth] Implement endpoint to retrieve referral data

franckc:
### Description:
This is now ready for review.

Implemented a new endpoint "invites" that returns the status of the invite for a given campaign
 - Completed invites are the ones that were completed during the campaign
 - Pending invites are the ones that are still outstanding.

Implemented a Referral rule which grants the referrer a reward upon the condition that the referee completed Level 2 during the campaign that is being evaluated.

Cleaned up a few things in adapter.

### Checklist:


experimental/origin-graphql/src/typeDefs/Growth.js
```diff
@@ -145,10 +145,11 @@ module.exports =
   type Query {
     # first property specifies the number of items to return
     # after is the cursor
-    campaigns(first: Int, after: String, walletAddress: ID!): GrowthCampaignConnection
-    campaign(id: String, walletAddress: ID!): GrowthCampaign
-    isEligible: EligibilityInfo
+    campaigns(first: Int, after: String): GrowthCampaignConnection
+    campaign(id: String): GrowthCampaign
+    invites(walletAddress: ID!, campaignId: Int!): ReferralAction
```
sparrowDom: I think it would be better to return this inside GrowthCampaigns's actions list

franckc: Ah sorry, yes you are right. this should be returned as an Action from the campaign endpoint. will update soon...

PR #1520, opened 2019-02-22
Private Key Disclaimer

micahalcorn:
This adds a blocking alert to tell the user to back up the private key under a couple of circumstances:

1️⃣ ETH is detected in the wallet.
2️⃣ We are about to recommend funding the wallet.

It can be permanently dismissed or just dismissed for the current session.

🔐


micahalcorn:
I also toned down the disclaimer about copying the private key to the clipboard and am attempting to automatically clear it after a minute as requested by @joshfraser.

PR #1522, opened 2019-02-22
Add sentry to origin-mobile

tomlinton:
Integrate Sentry with react-native. This is basically the output of the automated sentry integration script. I haven't attempted to test this at all, maybe you can try and spin it up @micahalcorn? There is an auth token you might need which goes in sentry.properties. You can ping me for it if you need it.

### Checklist:

- [] Test your work and double-check to confirm that you didn't break anything

### DIFF
--- a/origin-mobile/android/app/build.gradle
+++ b/origin-mobile/android/app/build.gradle
@@ -149,7 +149,7 @@ android {
 }
 
 dependencies {
-    compile project(':react-native-sentry')
+    implementation project(':react-native-sentry')
     implementation project(':react-native-push-notification')
     implementation project(':react-native-uuid-generator')
     implementation project(':react-native-webview')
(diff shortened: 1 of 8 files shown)


origin-mobile/android/app/build.gradle
```diff
@@ -148,6 +149,7 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-sentry')
```
mikeshultz: Instead of `compile`, this should be `implementation`.  It's just a warning right now but not for long.

tomlinton: Would you mind updating this if you think it is a good idea @mikeshultz? I have no idea what I'm doing with android builds :grin:

mikeshultz:
Made that change.  I'll give this a test build after either this or #1515 is merged, but eyeball parsing looks good.

micahalcorn:
Thanks @tomlinton 🙏 This worked for me after following [this](https://github.com/facebook/react-native/issues/23183#issuecomment-458106496). Do I just add your `auth.token` to my local sentry.properties before building a version to submit to the App Store?

PR #1526, opened 2019-02-23
Add sentry.properties files to .gitignore

tomlinton:
As per title.


tomlinton:
We haven't really decided about using a per package .gitignore or a root one for the monorepo. I don't really have much of an opinion either way, perhaps per package would make it easier to maintain as you wouldn't have one huge file.

PR #1532, opened 2019-02-25
Google cloud pub/sub integration for event-listener

tomlinton:
- Adds a hook to publish to Google cloud pub/sub for IPFS cluster pinning integration
- Adds service account credentials to event listener deployment for auth to Google cloud pub/sub

### Checklist:


origin-discovery/src/listener/handler_identity.js
```diff
@@ -266,6 +266,10 @@ class IdentityEventHandler {
   emailWebhookEnabled() {
     return true
   }
+
+  gcloudPubsubEnabled() {
+    return false
```
franckc: Identity data is stored in IPFS. So shouldn't we publish that IPFS hash to pubsub ?

origin-discovery/src/listener/handler_marketplace.js
```diff
@@ -379,7 +383,7 @@ class NoGasMarketplaceEventHandler extends MarketplaceEventHandler {
       throw new Error(`ListingIpfs and Id mismatch: ${listingId} !== ${listing.creator.listing.createDate}`)
     }
 
-    if(listing.creator != offer.buyer) 
+    if(listing.creator != offer.buyer)
```
franckc: nit: since we are changing this line, do you mind using the !== operator rather than != ?
same thing for line 388 below.

origin-discovery/src/listener/handler_identity.js
franckc: Ah... Perhaps the Google Cloud Function rohit implemented does not support extracting IPFS hash from identity data... Let's add a TODO or file a separate issue to keep track of this ?

origin-discovery/src/listener/handler_marketplace.js
tomlinton: I will do a prettier run after this PR is merged which should handle these.

origin-discovery/src/listener/handler_identity.js
tomlinton: Yep was going to do that as a separate PR. Also the unpinning if identity we discussed. Will file a separate issue.

franckc:
oh also please look at integration test failure. they seem relevant to this PR.

PR #1534, opened 2019-02-25
do not copy package-lock from host into docker

sparrowDom:
### Description:
I had a problem with my docker image build where the host outdated `package-lock.json` file would somehow prevent npm to resolve linked dependancies. Since docker's & host machine's `node_modules` and thus build environment are separate it would make sense to me to not copy `package-lock.json` json files over.


sparrowDom:
@tomlinton though I might be missing something... :)

PR #1549, opened 2019-02-26
No Messaging Prerequisite

micahalcorn:
This removes the requirement to enable messaging before buying or selling as discussed below. 🔥 

messaging


franckc:
It's kind of scary when IPFS chokes a monkey....

micahalcorn:
monkey

DanielVF:
GIF so good it makes me cry...

PR #1556, opened 2019-02-27
Add new origin-identity package

franckc:
### Description:
This is to solve the cyclical dependency issue between origin-discovery and origin-growth.

Currently origin-identity only includes the DB model for the identity table, but in the future we should move there contracts and business logic specific to identity.

### Checklist:


origin-discovery/src/listener/handler_identity.js
```diff
@@ -2,6 +2,7 @@ const Web3 = require('web3')
 const logger = require('./logger')
 
 const db = require('../models')
+const db2 = require('origin-identity/src/models')
```
tomlinton: Super nit: I wonder if it is possible to do something like:

```
const _discoveryModels = require('../models')
const _identityModels = require('origin-identity/src/models')
const db = { ..._discoveryModels, ..._identityModels }
```

I just find the naming a bit weird, it makes it look like two separate databases but they are the same database just different models.

franckc: Yes, this is much more elegant. Thanks for the suggestion. Updating...

PR #1557, opened 2019-02-27
Update spaceman sample listing

tomlinton:
- Add spaceman sample listing to origin-admin
- Update hash for listing in truffle migration. Not sure how this changed 😕


tomlinton:
@nick any idea what is up with this? With this PR (just the change of hash not the origin-admin bit) the listing will show up in origin-dapp2, but the integration tests fail. The integration tests work with the previous hash for the listing, but then the listing won't show in dapp2 because it is the wrong hash.

With the old hash it used to show up in dapp2. I'm confused :grin:

tomlinton:
I'll close this while I figure it out, but let me know if you have any ideas.

nick:
Yea not sure.. maybe just merge in master?

PR #1559, opened 2019-02-27
Dapp2 settings page

tomlinton:
First pass at implementing the settings page, a few things are different from the [mock](https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5bc779e5e0bf72191e77ec87)
- Notification radio won't work because there isn't a way to programatically revoke browser notification permissions
- Replaced yes/no indexing server option with a setting
- Didn't do messaging yet

Not too sure about this implementation @nick, keen to hear any thoughts. Not a priority though, so maybe just if you get bored. :grin:

screen shot 2019-02-27 at 4 07 32 pm


nick:
Looks good! Thinking we should have some validation on the inputs, for example valid URLs and perhaps even pinging the server to see if it's alive before actually saving the config. Since this is a 'secret' page that isn't linked to from anywhere directly we can always do that later though

PR #1561, opened 2019-02-27
Productionize growth server

franckc:
### Description:

Configuration for deployment of the new Apollo growth server. 

Note: the secrets changes are to add growthEnvKey.

### Checklist:


deployment/kubernetes/charts/origin/templates/growth.deployment.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "growth.fullname" . }}
```
tomlinton: This needs to be defined in `_helpers.tpl`. Same thing for `growth.host`. You can just copy and paste them for another service.

development/envfiles/origin-growth.env.js
```diff
@@ -0,0 +1 @@
+DATABASE_URL=postgresql://origin:origin@postgres/origin
```
tomlinton: No `.js` on the end of the file

deployment/dockerfiles/origin-growth
```diff
@@ -4,13 +4,22 @@ WORKDIR /app
...
 COPY ./scripts ./scripts
 
 RUN npm install --unsafe-perm
 
-CMD npm run start:growth --prefix origin-growth
\ No newline at end of file
+# Run migrations then start the growth server.
+CMD eval $(envkey-source) && \
+	npm run migrate --prefix origin-growth && \
+	npm run start:growth --prefix origin-growth
```
tomlinton: Not really related to this PR but I don't think we need the `npm run start:growth`, it should just be `npm run start` for everything. I think we should clean this up at some point.

deployment/dockerfiles/origin-growth
```diff
@@ -4,13 +4,22 @@ WORKDIR /app
...
 COPY ./origin-identity ./origin-identity
 COPY ./scripts ./scripts
 
 RUN npm install --unsafe-perm
 
-CMD npm run start:growth --prefix origin-growth
\ No newline at end of file
+# Run migrations then start the growth server.
+CMD eval $(envkey-source) && \
+	npm run migrate --prefix origin-growth && \
```
sparrowDom: Not related to this PR, just noticed something today, that `origin-growth` and `origin-discovery` use the same db and `sequalize` config. So you can't really run `$sequalize db:migrate:undo:all` in any of them, because sequalize migrations are located in two separate folders. 

Sometimes in the future we could either: 
 - have 2 different databases for these 2 services (`postgres://origin:origin@postgres/origin-discovery`, `postgres://origin:origin@postgres/origin-growth`)
 - or join the sequalize migrations into a separate module that accesses db layer. (`origin-database` or something similar)

deployment/kubernetes/charts/origin/templates/growth.deployment.yaml
franckc: ah yes. thank you for catching this.

development/envfiles/origin-growth.env.js
franckc: oops. good eye. fixed.

deployment/dockerfiles/origin-growth
franckc: Yes, currently db migrations and models are scattered. It gives us flexibility in future in case we decide to split into different DBs. But agreed right now it makes it a bit messy. We could discuss consolidating all migrations and models into a common origin-db module...

deployment/dockerfiles/origin-growth
franckc: Fixed.

deployment/dockerfiles/origin-growth
tomlinton: Good points @sparrowDom. The best approach is one database per service I think, reading/writing to the same database from multiple services isn't really a good separation of concerns. It is just a pain to configure separate databases so we went with the easy option for now.

PR #1562, opened 2019-02-27
[Growth] Add endpoint to send invite by email

franckc:

### Description:

Implemented endpoint that gets a list of recipients and send the referrer's invite code to them by email.

### Checklist:


experimental/origin-graphql/src/typeDefs/Growth.js
```diff
@@ -120,7 +120,6 @@ module.exports =
     code: String!
     success: Boolean!
     message: String!
-    invites: [Invite]
```
franckc: @sparrowDom Please confirm it's ok to delete this - I could not see why it would get returned ?

origin-growth/src/apollo/resolvers.js
```diff
@@ -76,11 +77,13 @@ const resolvers = {
     }
   },
   Mutation: {
-    async invite() {
+    async invite(root, args) {
+      // FIXME: Check the referrer against Auth token.
```
franckc: @sparrowDom My understanding is that when you implement the Terms signing, we'll then have in the DApp locale storage an Auth token that will be sent along as an HTTP header when calling any mutations on the server. This way we can check the referrer is really the one calling the mutation. Are we on the same page ?

origin-growth/src/apollo/resolvers.js
```diff
@@ -76,11 +77,13 @@ const resolvers = {
...
+      await sendInviteEmails(args.walletAddress, args.emails)
       return {
-        code: '418',
-        success: false,
-        message: 'I am a teapot'
+        code: '200',
```
franckc: @sparrowDom I was not sure what values to return here ? What is the intent of code, success and message exactly ?

experimental/origin-graphql/src/typeDefs/Growth.js
sparrowDom: Yes good point remove away

origin-growth/package-lock.json
```diff
@@ -71,6 +71,34 @@
 			"resolved": "https://registry.npmjs.org/@protobufjs/utf8/-/utf8-1.1.0.tgz",
 			"integrity": "sha1-p3c2C1s5oaLlEG+OhY8v0tBgxXA="
 		},
+		"@sendgrid/client": {
```
sparrowDom: We don't have package-lock in version control in a most of the modules. @tomlinton do we have any general guidelines whether to leave `package-lock.json` in the modules or leave it out? 

I know that docker environment ignores this file since it doesn't copy it into containers. Not sure what to do about local environments.

origin-growth/src/apollo/resolvers.js
sparrowDom: Yup exactly 👍

origin-growth/src/apollo/resolvers.js
sparrowDom: I think we can remove the `code` since that should be communicated with the `http.code`. The message I thought would serve as an error reporting field

origin-growth/src/apollo/resolvers.js
franckc: Excellent :)

experimental/origin-graphql/src/typeDefs/Growth.js
franckc: Sounds good. Thanks for confirming.

origin-growth/package-lock.json
franckc: Yes, package-lock.json is not used anymore. But I think that the idea is that when adding a new dependency, we want to commit it because otherwise the next person that runs lerna bootstrap on that scope will see package-lock.json has changed  and might get confused as to why it was not committed.

origin-growth/src/apollo/resolvers.js
franckc: Sounds good. Let me remove it.
I actually wonder if we should just rely on standard apollo responses for now and not bother about success and message neither ? My reasoning:
 - I don't think we are returning any user friendly messages from the back-end at this point ? Or do you have any specific use cases in mind ?
 - messages would need to get translated which would take some extra work/thinking
Perhaps we could add this later ? Let me know, I don't really feel strongly about this, just trying to keep things simple... :)

franckc:
CC @tomlinton as a heads up re the use of sendgrid

franckc:
Awesome. Thanks for that command - super useful. I ran it and verified origin-growth container is now building properly....

origin-growth/src/apollo/resolvers.js
sparrowDom: Yeah good point, agree with keeping it simple for now. And add complexity in case it is needed later.

origin-growth/package-lock.json
sparrowDom: Ah yes, this makes a lot of sense for local development. Another thing we could do is add all `package-locks.json` files to gitignore

origin-growth/src/apollo/resolvers.js
franckc: Ok. Let me take this out for now...

PR #1566, opened 2019-02-27
Fix growth server deployment

franckc:
### Description:
My test deployment build of origin-growth failed. [See logs](pulls most of the packages under experimental).

I think it is because I missed some dependencies in #1561.

origin-growth has the following dependencies:
 - origin-graphql (which pulls most of the packages under experimental)
 - origin-identity
 - origin-token

### Checklist:


franckc:
@tomlinton Is there actually a simple way for me to test this before I commit ?

tomlinton:
docker build . -f deployment/dockerfiles/origin-growth

sparrowDom:
Is it necessary to have a separate docker image built for `origin-growth`? I have been running growth successfully from [docker-compose.experimantal.yml](https://github.com/OriginProtocol/origin/blob/master/docker-compose.experimental.yml#L190-L201) and has been working ok for me. Currently it uses image built by the `Dockerfile-experimental` and has all the dependancies it needs there. 

If possible I would opt for as little separate images build as possible since from my experience using Docker in normal development I need to rebuild images 1-5 times a week. And for each of the images the linux image needs to be downloaded and lerna needs to do a bootstrap. Takes ~20 something minutes for me (With 16Mbit internet, optics should be faster). Another image will probably make that time longer. 

Except of course if I am missing something.

tomlinton:
This is the deployment container, which you shouldn't need to touch once it is working. It is better if it only includes the dependencies it needs for deployment (rather than everything like Dockerfile-experimental) because it'll be much faster to build/deploy on updates.

Now that it works you won't ever need to interact with it though, just keep keep using Dockerfile-experimental.

sparrowDom:
Ahh yes sorry missed that the file is located in `deployments` folder. Thanks for explanation!

PR #1568, opened 2019-02-27
copy enum_manifest file to translations directory

sparrowDom:
### Description:
From my experience for the translation enums to pick up changes the `.enum_manifest.json` needs to be copied to `translations` folder. 

I didn't go digging into code to find out why that is the case.

### Checklist:


sparrowDom:
@nick bump, in case you missed this one... it is only a 1 liner

PR #1575, opened 2019-03-01
Fix wallet dapp URLs

cuongdo:
The mobile wallet gets its URLs from the linking server. Right now,
we're returning bad URLs because staging's EnvKey has DAPP_URL set to
something that ends in `/#/`, which the linking server appends another
`/#/` to for the profile URL and other URLs.

This change uses Node's builtin `url` package to remove extra `/#/`s and
prevent double slashes. This should make the generated URLs a little
cleaner.

### Checklist:

### DIFF
--- a/origin-linking/src/logic/linker.js
+++ b/origin-linking/src/logic/linker.js
@@ -2,6 +2,7 @@
 import db from './../models/'
 import uuidv4 from 'uuid/v4'
 import { Op } from 'sequelize'
+import url from 'url'
 import { MessageTypes,EthNotificationTypes } from 'origin/common/enums'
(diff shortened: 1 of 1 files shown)


micahalcorn:
At least one way to find out ⏩

micahalcorn:
@cuongdo these URLs now work, but it's not actually pre-linking. When DApp2 loads, it always shows "Get Started". 😕

cuongdo:
Interesting. It works for me. Did you try unlinking all devices in the wallet app? It doesn't prelink if it thinks your mobile Safari is already linked, which is confusing but existing behavior.

PR #1577, opened 2019-03-01
Origin-bridge rewrite in JavaScript

tomlinton:
Herein lies the end of Python in our repository 😢 

This is a full rewrite of origin-bridge in JavaScript. It has the same level of testing as the existing Python code. I've tested it with origin-dapp2. It is untested with origin-dapp1 but it should work.

It is a big PR and a bit tough to review. It'd be great to get this merged so it can benefit from the thorough testing that dapp2 is going through on staging.


tomlinton:
Forgot to give a shoutout to @rohitverma007 who helped out with this, thanks for your contributions! 😎

micahalcorn:
❌ 🐍 ❌

origin-bridge/src/controllers/redirects.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+const express = require('express')
+const router = express.Router()
+
+router.get('/facebook', (req, res) => {
+  if (req.query.dappRedirectUrl) {
+    const dappRedirectUrl = req.query.dappRedirectUrl
+    res.redirect(dappRedirectUrl + '?origin-code=' + req.query.code)
```
sparrowDom: ultra nitpick: can be expressed without string concatenation
```
`${dappRedirectUrl}?origin-code=${req.query.code}`
```

origin-bridge/src/env.js
```diff
@@ -0,0 +1,8 @@
+'use strict'
```
sparrowDom: Just curious, is there a way to set environmental variables without having `envKey` running locally?

origin-bridge/src/controllers/twitter.js
```diff
@@ -0,0 +1,119 @@
...
+/* Get an oAuth request token from Twitter.
+ *
+ */
+router.get('/auth-url', async (req, res) => {
+  const dappRedirectUrl = req.query.dappRedirectUrl || null
+
+  let oAuthToken, oAuthTokenSecret
+  try {
+    // eslint-disable-next-line no-extra-semi
+    ;({ oAuthToken, oAuthTokenSecret } = await getTwitterOAuthRequestToken(
```
franckc: nit: why is the leading ; and the () needed ?
Couldn't this be written without them as:
` { oAuthToken, oAuthTokenSecret } = await getTwitterOAuthRequestToken(dappRedirectUrl)`

origin-bridge/src/logger.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
+
+const Logger = require('logplease')
+Logger.setLogLevel('DEBUG')
```
franckc: Wonder if we should log level by default to INFO rather than DEBUG ?

origin-bridge/src/logger.js
```diff
@@ -0,0 +1,7 @@
...
+module.exports = Logger.create('origin-bridge', {
+  color: Logger.Colors.Yellow
```
franckc: nit: I also tend to disable timestamp logging by setting option `showTimestamp: false` since we already get timestamp from GCP logs in prod.

origin-bridge/src/utils/attestation.js
```diff
@@ -0,0 +1,71 @@
...
+  return {
+    schemaId: 'https://schema.originprotocol.com/attestation_1.0.0.json',
+    data: data,
+    signature: signature
+  }
+}
+
+function generateAttestationSignature(privateKey, subject, data) {
+  if (!web3.utils.isHexStrict(privateKey)) {
+    throw 'Invalid private key, not a hex string'
```
franckc: nit: I think in our code base we tend to throw new Error(<msg>)

origin-bridge/src/utils/attestation.js
```diff
@@ -0,0 +1,71 @@
+'use strict'
+
+const eth = require('web3-eth')
+const Web3 = require('web3')
+const web3 = new Web3(process.env.PROVIDER_URL || 'http://localhost:8545')
```
franckc: We should not need to make any call to the blockchain, so why do we need to instantiate a web3 object with a provider ?
How about using utils method directly ? For ex: Web3.utils.isHexStrict() and Web3.utils.soliditySha3()

This way also we won't have to set the env variable PROVIDER_URL for the bridge environment.

origin-bridge/src/utils/index.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+const Web3 = require('web3')
+const web3 = new Web3(process.env.PROVIDER_URL || 'http://localhost:8545')
```
franckc: Same comment as above here. We should be able to use directly Web3, without instantiating an object.

origin-bridge/src/utils/index.js
```diff
@@ -0,0 +1,39 @@
...
+  let url = `${protocol}://${host}${relativeUrl}`
+  if (dappRedirectUrl) {
+    url += '?dappRedirectUrl=' + dappRedirectUrl
+  }
+  return url
+}
+
+function mapObjectToQueryParams(obj) {
+  return Object.keys(obj)
+    .map(key => key + '=' + obj[key])
```
franckc: keys and values should probably be URL encoded ? Perhaps using encodeURIComponent

origin-bridge/src/utils/attestation.js
```diff
@@ -0,0 +1,71 @@
...
+      // so the validation of the signature works.
+      stringify(data)
+    ),
+    version: '1.0.0'
+  }
+
+  // Save the attestation in the database
+  await Attestation.create({
+    method: attestationType,
+    ethAddress: ethAddress,
```
franckc: Should we follow the same convention we use for other tables and write the eth address in its canonical lowercase form ?

origin-bridge/src/env.js
tomlinton: You can just export each directly, e.g. `export TWILIO_VERIFY_API_KEY=abcd123` or you can `cp dev.env .env` and fill them out in there which should be read by dotenv when the package starts.

origin-bridge/src/controllers/twitter.js
tomlinton: That syntax isn't valid for destructuring objects, it only works if you also are declaring, e.g. `let { oAuthToken, oAuthTokenSecret } = await getTwitterOAuthRequestToken(dappRedirectUrl)` would work. It would work without the leading `;` but that gets added by prettier. Linting then fails because of an unnecessary semicolon which is why the eslint-disable is needed.

Crazy right? :grin:

origin-bridge/src/controllers/redirects.js
tomlinton: 👍Good spot, I prefer this too

deployment/dockerfiles/origin-bridge
```diff
@@ -1,14 +1,19 @@
...
-COPY ./deployment/dockerfiles/scripts/start-bridge-server.sh \
-	/usr/local/bin/start
+COPY ./origin-bridge /app
 
-EXPOSE 5000
+RUN npm install --loglevel notice
 
-CMD ["start"]
+CMD eval $(envkey-source) && \
+	npm run migrate && \
```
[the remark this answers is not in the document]
tomlinton: Good point, thanks for reminding me!

origin-bridge/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "origin-bridge",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
tomlinton: Good idea, seems like a good idea to keep it consistent 👍

origin-bridge/src/logger.js
tomlinton: Good plan!

origin-bridge/src/logger.js
tomlinton: Will do!

origin-bridge/src/utils/attestation.js
tomlinton: Right on, thanks! Fixed.

origin-bridge/src/utils/attestation.js
tomlinton: 👍

origin-bridge/src/env.js
sparrowDom: ok tnx 👍

origin-bridge/src/controllers/twitter.js
sparrowDom: Haha what a face plant by the linter/prettifier

origin-bridge/migrations/20190212002833-create-attestation.js
```diff
@@ -0,0 +1,39 @@
...
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('attestation', {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER
+      },
+      eth_address: {
+        type: Sequelize.CHAR(42)
```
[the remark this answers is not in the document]
tomlinton: Fixed 👍

origin-bridge/src/utils/attestation.js
tomlinton: Good catch!

deployment/dockerfiles/origin-bridge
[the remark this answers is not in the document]
tomlinton: Manually inserted the migration entries.

PR #1583, opened 2019-03-02
Sparrow dom/growth engine2

sparrowDom:
### Description:
This pull request: 
 - adds growth engine authentication: 
   - user signs terms using Metamask
   - when the server confirms validity of the signature it creates an auth_token
   - auth_token is stored in `growth_participant` table and returned to client
   - token is stored in browsers `localStorage` and passed as an `authentication` header each time a growth Apollo call is made
   - resolvers are passed `enrollmentStatus` context item, and reject access to some resources
 - implements 80% referral invite page functionality
 - resolves GraphQl Apollo bug present in current master

### Checklist:


sparrowDom:
When you have time @nick and @franckc if you can check this PR out. Perhaps it would make sense for @franckc to check out more of the `origin-growth` changes and @nick the `dapp2`/ `origin-graphql` part... or you can mix it up :)

experimental/origin-dapp2/src/pages/growth/Invite.js
```diff
@@ -0,0 +1,252 @@
...
+      subPage: 'sendInvites',
+      inviteCode: 'origin-invite-code',
+      showCopyConfirmation: false
+    }
+
+    this.inviteCode = `${location.protocol}//${location.hostname}/#/welcome/${
+      this.state.inviteCode
+    }`
+
+    this.handleNavigationClick = this.handleNavigationClick.bind(this)
```
nick: The way you're calling these methods, ie inside a `onClick={() => this.fn()}`, I don't think you need to bind them any more. The rest of dapp2 doesn't have binding in constructors. Not a big deal but thought I would mention :-)

experimental/origin-dapp2/src/pages/growth/Invite.js
```diff
@@ -0,0 +1,252 @@
...
+function NavigationItem(props) {
+  const { selected, onClick, title } = props
+  return (
+    <a
+      href="#"
+      onClick={e => {
+        e.preventDefault()
+        onClick()
+      }}
+      className="pt-4 pb-43 pr-4"
```
nick: is pb-43 correct?

experimental/origin-dapp2/src/pages/growth/Invite.js
```diff
@@ -0,0 +1,252 @@
...
+
+require('react-styl')(`
+  .growth-invite.container
+    max-width: 760px;
+  .growth-invite
+    margin-top: 70px;
+    .back img
+      width: 15px;
+      margin-right: 6px;
+      -webkit-transform: rotate(270deg);
```
nick: no need to add vendor prefixed styles - should get added automatically when the css is built

nick: Also no need for semi colons but I guess it's a pain to remove them if pasting styles in :-)

experimental/origin-dapp2/src/pages/growth/mutations/Enroll.js
```diff
@@ -0,0 +1,132 @@
...
+            <Mutation
+              mutation={GrowthEnroll}
+              onCompleted={({ enroll }) => {
+                if (enroll.error === null && enroll.authToken) {
+                  localStorage.setItem('growth_auth_token', enroll.authToken)
+                  this.props.history.push('/campaigns')
+                } else {
+                  console.log('Error occurred: ', enroll)
+                  this.setState({
+                    error: 'Can not enroll into growht campaign'
```
nick: typo

experimental/origin-dapp2/src/pages/growth/mutations/Enroll.js
```diff
@@ -0,0 +1,132 @@
...
+                this.setState({
+                  error: 'Problems enrolling into growth campaign.'
+                })
+              }}
+            >
+              {enroll => (
+                <Mutation
+                  mutation={SignMessageMutation}
+                  onCompleted={({ signMessage }) => {
+                    console.log('Message successfuly signed: ', signMessage)
```
nick: can probably take out these debug messages

experimental/origin-dapp2/src/pages/growth/mutations/Enroll.js
```diff
@@ -0,0 +1,132 @@
...
+                          autoPlay
+                          loop
+                        >
+                          <source
+                            src="images/growth/metamask_in_browser_dark_bg.mp4"
+                            type="video/mp4"
+                          />
+                          Your browser does not support the video tag.
+                        </video>
+                        <div className="title">Confirm Metamask Signature</div>
```
nick: Since the wallet provider can vary, probably need to take out MetaMask references. Maybe we can just add a todo for now

experimental/origin-graphql/src/index.js
```diff
@@ -6,7 +6,7 @@ import messagingSync from './messagingSync'
 import fragmentMatcher from './typeDefs/fragmentTypes'
 
 const cache = new InMemoryCache({ fragmentMatcher })
-const client = new ApolloClient({ link, cache })
+const client = new ApolloClient({ link: link, cache })
```
nick: how come this change?

experimental/origin-dapp2/src/pages/growth/Invite.js
sparrowDom: Good point, removing the `bind` and changing to current DApp2 style.

experimental/origin-dapp2/src/pages/growth/Invite.js
sparrowDom: tnx, a typo... correcting

experimental/origin-dapp2/src/pages/growth/Invite.js
sparrowDom: Nice, didn't notice when css compiles id adds vendor specific styles, thanks. Also removed all semicolons from growth components

experimental/origin-dapp2/src/pages/growth/mutations/Enroll.js
sparrowDom: Good eye tnx

experimental/origin-dapp2/src/pages/growth/mutations/Enroll.js
sparrowDom: yup, removed

experimental/origin-dapp2/src/pages/growth/mutations/Enroll.js
sparrowDom: 👍 added a todo

experimental/origin-graphql/src/index.js
sparrowDom: Ahh sloppy reset when I was testing something. Undoing change...

sparrowDom:
@nick thanks for a very quick over the weekend review. 🙏

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+  })
+
+  // don't override the existing invite code
+  if (existingInvite !== null) {
+    return
+  }
+
+  const code =
+    `${accountId.substring(2, 5)}-` +
```
franckc: How are ETH addresses generated. Don't we have a risk of collision ? Granted it is pretty small, but still a risk ?

One idea to generate unique invite code would be to leverage GrowthInviteCode.id since it is a sequence and will generate unique ids. It would go like this:
 - insert an entry in GrowthInviteCode for eth_address.
 - get the id
 - generate a unique code based on the id by using symmetric encryption on the id or some sort of hashing. we could use the id directly as code, but perhaps we don't want to expose it in the clear since it would reveal how many participants in the program ?
 - update GrowthInviteCode with the generated code.

Another possibility is to keep what you have but check for collision by trying to load any existing row that may have the same code. If found, the modify slightly the code and try again.

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,126 @@
...
+
+  const authToken =
+    participant === null
+      ? crypto.randomBytes(64).toString('hex')
+      : participant.authToken
+
+  const participantData = {
+    ethAddress: accountId,
+    status: enums.GrowthParticipantStatuses.Active,
+    agreementId: agreementMessage,
```
franckc: Should we have an growth_agreement table where the agreement text is stored and an id that is then referred to here ? One problem would be how to handle translation...
An alternative would be to have the user only signing an agreement ID (ex: "Campaign Agreement v1.0") which would not be translated, but it does look a bit trustworthy to user if we show them they are signing the agreement data ? We can chat and decide what's best...

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,126 @@
...
+    // if user uses growth from 2 devices let them share the same auth token
```
franckc: For my understanding, the user will have to agree to the terms on each device, right ? might be worth clarifying in the comment.

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,126 @@
...
+    throw new Error('This user is banned')
+  }
+
+  const authToken =
+    participant === null
+      ? crypto.randomBytes(64).toString('hex')
+      : participant.authToken
+
+  const participantData = {
+    ethAddress: accountId,
```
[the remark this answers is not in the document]
sparrowDom: good catch 👍

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,126 @@
...
+    signature
+  )
+
+  if (accountId !== recoveredAccountId) {
+    throw new Error('Recovered and provided accounts do not match')
+  }
+
+  const participant = await db.GrowthParticipant.findOne({
+    where: {
+      ethAddress: accountId
```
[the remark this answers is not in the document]
sparrowDom: 👍

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,126 @@
...
+    return enums.GrowthParticipantAuthenticationStatus.Banned
+  } else {
+    return enums.GrowthParticipantAuthenticationStatus.Enrolled
+  }
+}
+
+async function createInviteCode(accountId) {
+  const existingInvite = await db.GrowthInviteCode.findOne({
+    where: {
+      ethAddress: accountId
```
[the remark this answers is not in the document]
sparrowDom: :+1

origin-growth/src/resources/authentication.js
sparrowDom: Very nice catch thanks! When creating the growth invite code I now get the count of `GrowthInviteCode` table and convert that to human friendly hash, that represent the last characters of the invite code.

origin-growth/src/resources/authentication.js
sparrowDom: Yeah I have a `TODO` comment on a couple of places in the Dapp2 saying this should be read from db. Yeah lets talke about it today on our call.  Have made an agenda item [in our doc](https://docs.google.com/document/d/15nzudW56hQVIWHc7DRJI5XzxenM0-oc2p7z4vOqc8ms/edit?ts=5b7b483f)

origin-growth/src/resources/authentication.js
sparrowDom: good point adding a comment

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,157 @@
...
+  const inviteCodesCount = await db.GrowthInviteCode.count({})
+
+  const code =
+    `${accountId.substring(2, 5)}-` +
+    `${accountId.substring(accountId.length - 3, accountId.length)}-` +
+    `${toReadableHex(inviteCodesCount)}`
+
+  await db.GrowthInviteCode.create({
+    ethAddress: accountId.toLowerCase(),
+    // Consists of first 6 and last 3 ether address letters
```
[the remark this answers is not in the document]
sparrowDom: 👍 moved and updated comment

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,157 @@
...
+      ethAddress: accountId.toLowerCase()
+    }
+  })
+
+  // don't override the existing invite code
+  if (existingInvite !== null) {
+    return
+  }
+
+  const inviteCodesCount = await db.GrowthInviteCode.count({})
```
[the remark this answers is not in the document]
sparrowDom: Yup I think we should be safe here: 
- in theory if the same second 2 different accounts that share the same beginning and ending address letters do a growth enrolment then a collision would happen.

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,157 @@
...
+
+  const code =
+    `${accountId.substring(2, 5)}-` +
+    `${accountId.substring(accountId.length - 3, accountId.length)}-` +
+    `${toReadableHex(inviteCodesCount)}`
```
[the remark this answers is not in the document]
sparrowDom: Agree that random number encoded to readable string is better. Since I already have a function that converts a number to a readable string I have just generated a random number between 1 and a million like you suggest: 
```
const code =
    `${accountId.substring(2, 5)}` +
    `${accountId.substring(accountId.length - 3, accountId.length)}` +
    `${toReadableHex(Math.round(Math.random() * 1000000))}`
```
Are you ok with this solution?

origin-growth/src/resources/authentication.js
```diff
@@ -0,0 +1,159 @@
...
+    return
+  }
+
+  /* Consists of first 3 and last 3 ether address letters plus hex
+   * representation of a random number
+   */
+  const code =
+    `${accountId.substring(2, 5)}` +
+    `${accountId.substring(accountId.length - 3, accountId.length)}` +
+    `${toReadableHex(Math.round(Math.random() * 1000000))}`
```
franckc: toReadableHex does not not truly generate hexa. hexa uses alphabet [0-9, A-E].
how about simply using the toString(16) method ?
> (Math.round(Math.random() * 1000000)).toString(16)

sparrowDom: Thanks much better, and I can get rid of the `toReadableHex` function.

PR #1592, opened 2019-03-04
Add persistent session storage for origin-bridge

tomlinton:
- Stores session data in postgresql rather than in memory (which isn't suited for production due to memory leaks according to warnings when starting).
- Fixed an error message


tomlinton:
The `sessionStore.sync()` call ensures that the table is created/correct. I should add a comment explaining it.

Issue #1469, opened 2019-02-13
Make Listing Creation Modular for multiple listing types


wanderingstan:
Currently, the listing creation process supports Unit and Fractional listings by means of a nest of conditionals and `if` statements. On DApp1 We've seen how this yields spaghetti code that only gets worse with more listing types.   (e.g. [here](https://github.com/OriginProtocol/origin/blob/master/experimental/origin-dapp2/src/pages/create-listing/Step2.js#L65), [here](https://github.com/OriginProtocol/origin/blob/master/experimental/origin-dapp2/src/pages/create-listing/CreateListing.js#L55-L57), or [here](https://github.com/OriginProtocol/origin/blob/master/experimental/origin-dapp2/src/pages/create-listing/Step2.js#L163-L173))

The plan is to move to a modular system, where each listing type is free to define it's own listing creation components.

## Issues
- [x] Custom prices and availability for Fractional calendar not getting saved (or not always getting saved) -- on `Details.js`, the added custom fields arn't getting added with `fields: Object.keys(props.listing)`, presumably because the `setState()` hasn't finished adding the new fields by then..
- [x] No boost on Announcements, right? (check with Nick)
- [ ] Minor js and css changes across Details.js listing types..double check
- [x] Handle **edit** mode
- [ ] Set `step` in query string when navigating around creation flow
- [x] Remove `farts`
- [ ] Correct chaining of `onChange()` from [Details]-->[Announcementlistings]-->CreateListing ? 
- [x] Get feedback from aure on creation progress bar. 
- [x] See about removing .. from `import Price from '../../../../components/___'`
- [ ] `onNext()` seems to break react pattern. Ask about it. 
- [ ] Can we get the listing-specific knowledge out of `listing/Edit.js` and into the related listign-type files?

micahalcorn:
Should any of these be moved to a separate issue @wanderingstan? Or can this be closed? It looks like the meat of this has been resolved.

wanderingstan:
Completed in https://github.com/OriginProtocol/origin/pull/1648 
Thanks @micahalcorn !

PR #1598, opened 2019-03-05
Modular listing creation by listingtype

wanderingstan:
NOTE: Not to merge before DApp2 goes live. 

Per #1469 this PR abstracts the listing creation flow so that listing types can define their own arbitrary creation steps. It is also the first step to gathering all listingtype-specific logic in one place. 

Instead of a predefined number of steps, there is a directory for each listing type in the `listing-types` that contains:
- A component with the listing type's name that serves as the "controller", which dictates the order and flow between the steps. E.g. `<UnitListing />`
- Components for each step, e.g. `<Boost />`

Note that as is, this PR results in code duplication between the step components. We should work on abstracting out common cross-listingtype components.

@nick before you approve, can I go over these issues with you to double check my approach? : 
- [ ] Correct chaining of `onChange()` from [Details]-->[Announcementlistings]-->CreateListing ? 
- [ ] Set `step` in query string when navigating around creation flow
- [x] Minor js and css changes across Details.js listing types.
- [ ] `onNext()` seems to break react pattern. 
- [ ] Is it possible to get the listing-specific knowledge out of `listing/Edit.js` and into the related listign-type files?


wanderingstan:
Ignore this PR -- with the repo file structure changing I had to move things around, and am in a new branch here: https://github.com/OriginProtocol/origin/tree/stan/modular-dapp2-moved

wanderingstan:
Closed in favor of updated https://github.com/OriginProtocol/origin/pull/1648

PR #1599, opened 2019-03-05
[Growth] Placeholder for fraud logic

franckc:
### Description:

In dev we'll use this Fraud placeholder.
For prod I have implemented the full-fledge logic which I'm going to commit to a private repo.

### Checklist:


origin-growth/src/scripts/verifyEvents.js
```diff
@@ -12,6 +12,14 @@ const parseArgv = require('../util/args')
 Logger.setLogLevel(process.env.LOG_LEVEL || 'INFO')
 const logger = Logger.create('verifyEvents', { showTimestamp: false })
 
+// Load the production fraud engine if available, and fall back to the dev one otherwise.
+let FraudEngine
+try {
+  FraudEngine = require('../fraud/prod')
```
sparrowDom: I would prefer a hard fail in incorrect environment. This way it can not fallback to dev on production: 
```
if (process.env.GROWTH_FRAUD_DETECTION) {
  FraudEngine = require('../fraud/prod')
} else {
  FraudEngine = require('../fraud/dev')
}

```

origin-growth/src/scripts/verifyEvents.js
```diff
@@ -38,7 +43,7 @@ class VerifyEvents {
 
     for (const event of events) {
       let status
-      if (this._isFraud(event)) {
+      if (await this.fraudEngine.isFraudEvent(event)) {
```
sparrowDom: This can be done in parallel with
```
Promise.all(events.map(event => this.fraudEngine.isFraudEvent(event))) 
...
```

sparrowDom: Of course the order of promise execution is random here, in case that is important, above solution is no good.

origin-growth/src/scripts/verifyEvents.js
tomlinton: I think that @sparrowDom is probably right, it is better to hard fail if something goes wrong with pulling the production code. We could use `process.env.NODE_ENV === 'production'` so we don't have to introduce another environment variable. Maybe also add a few logs at this point to say something like `Loaded development fraud engine`, or `Loaded production fraud engine`.

origin-growth/src/scripts/verifyEvents.js
franckc: Yes, good point that could get parallelized... Though performance is not yet a concern since the volume of events will be low to start with. So if you ok with that I'd prefer processing in order since that could be a useful property at some point later when running the fraud detection heuristics.

sparrowDom: yup sure 👍

origin-growth/src/scripts/verifyEvents.js
sparrowDom: I was thinking of the `process.env.NODE_ENV === 'production'` solution also, but was concerned that setting that as an environmental variable might affects some other configuration in the container which might have undesired side effects.

franckc: Good point. Done.

franckc: NODE_ENV is already set at the container level (see [this](https://github.com/OriginProtocol/origin/blob/master/deployment/dockerfiles/origin-growth#L5)) so let me use that...

sparrowDom: Sorry I wasn't clear of what I meant to say. For production purposes `NODE_ENV` is perfect, for local setup (few developers who will have that) only `NODE_ENV` might cause side effects. 

Perhaps something like: 
```
if (process.env.NODE_ENV === 'production' || process.env.GROWTH_FRAUD_DETECTION)  {
...
```
So on local we can just set the `GROWTH_FRAUD_DETECTION` to true without the risk of side effects of setting `NODE_ENV = true`

franckc: Sounds good. Added.

Issue #1212, opened 2019-01-10
Remove duplicate GA script tag


joshfraser:
We're currently including Google Analytics twice. We're loading it via both analytics.js and googletagmanager. We should remove one of these (probably the tag manager one).

alexjspears:
@micahalcorn @joshfraser happy to look at this once prioritized

DanielVF:
@alex-j-s You can start on this anytime, no need to wait!

micahalcorn:
I think this is now invalid per https://github.com/OriginProtocol/origin/pull/1601#issuecomment-470406646 ✌️

PR #1601, opened 2019-03-05
Remove duplicate GA script tag (#1212)

AurhinDev:
Google Analytics is loaded through both analytics.js
and googletagmanager. Removing the load through googletagmanager.

### DIFF
--- a/origin-dapp/public/dev.html
+++ b/origin-dapp/public/dev.html
@@ -12,8 +12,7 @@
     <meta name="theme-color" content="#2585FB">
 
     <title>Origin Protocol DApp</title>
-    <!-- Google Analytics -->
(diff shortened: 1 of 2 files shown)


franckc:
@DanielVF You might be the one in the team who is the most familiar with GA :)
Do you mind reviewing and perhaps also making sure DApp2 does not have the same problem ?

DanielVF:
@franckc When I added GA to DApp2, I made sure to put only one script in!

https://github.com/OriginProtocol/origin/blob/491517f39195beb8990549c34ca84f9b88e2359d/experimental/origin-dapp2/public/template.html#L24

nick:
Since this is resolved in dapp2, going to close this

PR #1634, opened 2019-03-07
New dapp and docker fix

sparrowDom:
### Description:
In this pr: 
- docker images build a lot faster since they are built only once. Seems that `build` config triggers rebuilding of image and Lerna hoisting for each of the containers
- fix contracts path preventing `origin-dapp` container to start
- (slightly unrelated) enable `origin-growth` by default since this won't effect staging/production deployment

### Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ matrix:
         # Setting required for Elasticsearch to run
         - sudo sysctl -w vm.max_map_count=262144
       script:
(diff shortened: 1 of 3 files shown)


sparrowDom:
Cool thanks for sorting all this out. I removed them because the whole docker rebuilt took > 1 hour, since each image was doing `npm run bootstrap`... wasn't aware of the tests consequences.

PR #1661, opened 2019-03-11
Cleanup: use origin-bridge attestation model in listener

franckc:
### Description:
 - Use origin-bridge attestation model rather than raw query in the listener
 - origin-bridge now writes lowercased eth_address and eth_address values in dev, staging and prod have been lowercased so the query now uses lowercase.

### Checklist:


origin-discovery/src/listener/handler_identity.js
```diff
@@ -1,9 +1,10 @@
 const Web3 = require('web3')
 const logger = require('./logger')
 
+const _bridgeModels = require('origin-bridge')
```
tomlinton: Should this be `origin-bridge/src/models`?

franckc: doh ! I Fixed.

PR #1669, opened 2019-03-12
Pin service update + identity pinning

tomlinton:

### Description:

- This is the code that I updated to do the pinning of listings, I deployed it to GCF but forgot to submit a PR
- Extended it a bit for identity stuff, @franckc there is a minor change to the event listener so it sends along the IPFS hash that you might want to review.

### Checklist:


deployment/utils/gcf-pin-ipfs-hash/index.js
```diff
@@ -1,85 +1,105 @@
...
+    if ('media' in ipfsData.data && ipfsData.data.media.length > 0) {
+      const mediaData = ipfsData.data.media
+      mediaData.forEach(media => {
+        hashesToPin.push(extractIpfsHashFromUrl(media.url))
+      })
     } else {
-      console.log("No media ipfs hashes found in listing data");
+      console.log('No IPFS media hashes found in listing data')
     }
+  } if (eventName === 'IdentityUpdated') {
```
franckc: this should be a else if ?

origin-discovery/src/listener/handler_identity.js
```diff
@@ -258,6 +260,10 @@ class IdentityEventHandler {
       user.profile.avatar = user.profile.avatar.slice(0, 32) + '...'
     }
 
+    if (log.decoded.ipfsHash) {
```
franckc: Why is this condition necessary ? Is this to guard for future when this handler receives delete event that contrary to add/update event do not have an IPFS hash ?
https://github.com/OriginProtocol/origin/blob/master/origin-contracts/contracts/identity/IdentityEvents.sol#L17

If yes, maybe add a simple comment. It was not totally obvious to me at first.

origin-discovery/src/listener/utils.js
```diff
@@ -112,7 +113,13 @@ async function withRetrys(fn, exitOnError = true) {
   }
 }
 
+function bytes32ToIpfsHash(hashHex) {
```
franckc: Maybe we should have an origin-util package where we put this kind of common utility method ?
I'm not suggesting we do this in this PR, just thinking in the future...

origin-discovery/src/listener/handler_identity.js
tomlinton: Mostly because I was being lazy and the tests didn't pass a full log so this failed. I'll add it to the tests. :grin:

origin-discovery/src/listener/utils.js
tomlinton: Yea, agreed, we need to do that we have the reverse of this function floating around in a few places as well.

deployment/utils/gcf-pin-ipfs-hash/index.js
tomlinton: Thanks, good catch!

PR #1670, opened 2019-03-12
Dev environment and Docker Compose updates

tomlinton:

### Description:

- Removes origin-dapp from docker compose in favour of running the dapp outside of Docker (i.e. normally) and uses docker compose for running backend services (blockchain, ipfs, etc).
- Enables Origin testnet and adds instructions for getting Eth from faucet.dev.originprotocol.com
- Switches dev deployment to Origin testnet
- Removes dedicated ES container in favour of creating the index with @sparrowDom's `es-cli.js`. I added a bit of a hacky workaround for when the index already exists. @sparrowDom you might want to check it out/improve it
- Removed the old envfiles in favour of defining everything in `docker-compose.yml` so that the development `directory` is no longer needed.
- Update development docs with changes + origin-admin instructions

### Checklist:


origin-graphql/src/contracts.js
```diff
@@ -132,13 +132,26 @@ const Configs = {
...
   localhost: {
     provider: `http://${HOST}:8545`,
     providerWS: `ws://${HOST}:8545`,
     ipfsGateway: `http://${HOST}:8080`,
     ipfsRPC: `http://${HOST}:5002`,
     growth: `http://${HOST}:4001`,
-    bridge: 'https://bridge.staging.originprotocol.com',
+    bridge: 'https://bridge.dev.originprotocol.com',
     automine: 2000,
     affiliate: '0x0d1d4e623D10F9FBA5Db95830F7d3839406C6AF2',
```
sparrowDom: Could we read OriginToken , V00_Marketplace and IdentityEvents contracts addresses the same way it is done in docker configuration? They are read from the values from contracts build jsons (origin-contracts/build/contracts/...). Maybe we could auto deploy the contracts and read the addresses from the build files so there is no need to manually deploy contracts using `origin-admin`

sparrowDom: Also should we remove the `docker` configuration part?

tomlinton: The `docker` configuration reads the contract build files and the contracts are deployed the same way they have always been in the Docker setup (via truffle). This localhost config is for `origin-admin` based deployment.

So we don't want to remove the `docker` part for that reason.

sparrowDom: Sorry I might be confused a bit. I was thinking since `dapp` in this PR can no longer be ran from docker (only localhost) we don't need the `docker` configuration part in `origin-graphql` anymore

PR #1677, opened 2019-03-12
[Growth] Set event createdAt to blockchain event date.

franckc:
### Description:

Use date from blockchain rather than current date so that if the listener is down for a period of time or if we do some re-indexing, the createdAt field won't vary.

### Checklist:

### DIFF
--- a/origin-growth/src/models/growth_event.js
+++ b/origin-growth/src/models/growth_event.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const Sequelize = require('sequelize')
+
 const enums = require('../enums')
 
 module.exports = (sequelize, DataTypes) => {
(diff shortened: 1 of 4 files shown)


origin-growth/src/models/growth_event.js
```diff
@@ -10,10 +12,21 @@ module.exports = (sequelize, DataTypes) => {
       type: DataTypes.ENUM(enums.GrowthEventTypes),
       status: DataTypes.ENUM(enums.GrowthEventStatuses),
       ethAddress: DataTypes.STRING,
-      data: DataTypes.JSONB
+      data: DataTypes.JSONB,
+      createdAt: {
+        type: DataTypes.DATE,
+        defaultValue: Sequelize.fn('NOW')
```
sparrowDom: Are there any events that do not have `date` associated with them? If not perhaps we can remove the `defaultValue` in favour of hard failing when there is an insert attempt without a date

franckc: Currently there should always be a date. Implemented your suggestion to make date required.

PR #1678, opened 2019-03-12
Fix default value for attestation createdAt

franckc:
### Description:

I noticed this was not working while working on a different table...
I don't think it is exercises in the bridge so nothing is currently broken

### Checklist:


franckc:
See doc at http://docs.sequelizejs.com/manual/tutorial/models-definition.html

PR #1696, opened 2019-03-14
[Growth] Geolocation implementation

franckc:
### Description:

It turns out getting geolocation information from google infrastructure is trickier than we thought :(
For the time being, we decided to use an external geolocation service.

@sparrowDom had already written the code for integration with ipdata.co so I've just resurrected his code...

### Checklist:


sparrowDom:
Oh an probably this would be a good time to also add the key to  `envKey` in case it is not there already

PR #1718, opened 2019-03-15
console.log()-->console.error(),console.warn() where appropriate

wanderingstan:
- Makes it easier to spot errors and warnings in console
- Makes it easier to spot left-behind debugging console.log's in logs that might otherwise get committed.


sparrowDom:
this reminds me @wanderingstan I once tried to tackle the problem that whenever a Graphql server has an internal error just a useless message is reported instead of what server actually returns. (a stacktrace) Will get around to that at some point

Screenshot 2019-03-15 14 33 13

PR #1725, opened 2019-03-16
No JSONSchema

### LINKED ISSUE #24: Add support for loading sub-schemas by URL to the React JSONSchema library

micahalcorn:
I believe that we no longer use JSONSchema in DApp2. 🤔 This removes the reference from our docs and should close #24.

### DIFF
--- a/origin-docs/guides/overview/architecture.md
+++ b/origin-docs/guides/overview/architecture.md
@@ -7,7 +7,7 @@ category: "Overview"
 
 # Architecture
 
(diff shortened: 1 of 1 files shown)


micahalcorn:
Or rather, it's only [react-jsonschema-form](https://github.com/mozilla-services/react-jsonschema-form) that we no longer use.

PR #1730, opened 2019-03-18
Past event batching

nick:
The event cache finds all events in a given range and stores them in local storage. Previously this was all done in a single call which could take a long time to respond. This new code takes a range of blocks and splits it up into requests of 20,000 each. It then sends those requests in batches of 7.

For example, a range of blocks of 6,000,000 to 7,000,000 would result in 100 requests.These requests are then sent in batches of 7, so that the node is not hit with 100 requests at once.

I reached the 20,000 blocks in batches of 7 after a bit of experimentation... seems to be a sweet spot.

### DIFF
--- a/origin-graphql/src/contracts.js
+++ b/origin-graphql/src/contracts.js
@@ -32,6 +32,8 @@ if (typeof window !== 'undefined') {
 
 const Configs = {
   mainnet: {
+    // provider:
+    //   'https://eth-mainnet.alchemyapi.io/jsonrpc/FCA-3myPH5VFN8naOWyxDU6VkxelafK6',
     provider: 'https://mainnet.infura.io/v3/98df57f0748e455e871c48b96f2095b2',
     // providerWS: 'wss://mainnet.infura.io/ws',
     ipfsGateway: 'https://ipfs.originprotocol.com',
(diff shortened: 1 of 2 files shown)


nick:
This also lets us switch to using the Alchemy API instead of Infura (Alchemy has a timeout on requests which prevented long running getLog api calls)

origin-graphql/src/contracts.js
```diff
@@ -32,6 +32,8 @@ if (typeof window !== 'undefined') {
 
 const Configs = {
   mainnet: {
+    // provider:
```
franckc: We don't have to do this in this PR, but could we start using envkey for all these config variables ?

origin-graphql/src/utils/eventCache.js
```diff
@@ -1,10 +1,32 @@
...
 // import { post } from 'origin-ipfs'
 import uniqBy from 'lodash/uniqBy'
+import chunk from 'lodash/chunk'
+import flattenDeep from 'lodash/flattenDeep'
 import createDebug from 'debug'
 
 const debug = createDebug('event-cache:')
 
+// Given a contract and block range, break the range up into requests of no
+// more than 20,000 blocks, then send those requests in batches of 7.
```
franckc: nit: mention that after experimentation 7 is the sweet spot. otherwise is seems kinda random :)

PR #1732, opened 2019-03-18
[Growth] Update list of forbidden countries

franckc:
### Description:

Same as title

### Checklist:


joshfraser:
For the record, it makes me incredibly sad that our lawyers are telling us to do this. The idea that an entire country of people should be blacklisted is insane.

PR #1740, opened 2019-03-19
SDN blacklist implementation

franckc:
### Description:

Check user against SDN blacklist as part of token distribution.
If the first name and last name match, the user gets banned from the program and does not receive any token distribution.

### Checklist:


franckc:
Note: the SDN list does not have country data so I don't think we need to capture the country of the user in the growth_participant table for now.

infra/growth/src/util/sdnMatcher.js
```diff
@@ -0,0 +1,99 @@
...
+   * @param lastName
+   * @returns {boolean}
+   */
+  match(firstName, lastName) {
+    // Normalize the input.
+    const normFirstName = firstName.trim().toLowerCase()
+    const normLastName = lastName.trim().toLowerCase()
+
+    // Check against records of lastNames/firstNames.
+    const firstNames = this.lastNames[normLastName]
```
sparrowDom: I was just checking some balkan names and people have local letter variants that are simplified in the SDN document. For example a serb would put his name in as `Miroslav Radić` while in the SDN document it would show up as `Miroslav Radic` -> notice the different `c` letters.

There are probably similar simplifications od spanish `ñ` and german `ä`, `ö`, `ü` and a ton of other letters. 

I think we should simplify the users input `name` + `lastName` before we do the matching.

franckc: Yes, good points. There are lot of things we could do to improve the matching... Fuzzy matching would handle the case you mentioned. So perhaps we can focus on that first. Would you be ok if I add a TODO and land this as is ? Then we can iterate to make this better if needed...

sparrowDom: Yup that sounds good

PR #1744, opened 2019-03-19
Fix cron container to source envkey

franckc:
### Description:
 - I noticed the cron jobs where failing due to trying to connect to redis on localhost because REDIS_URL wasn't set. I think it's because we were not sourcing envkey.
 - I did check the proper envkey value is set for origin-cron in the secrets.

### Checklist:


franckc:
Yep, agreed we should standardize on require('envkey')...

PR #1649, opened 2019-03-11
Forked Wallet Onboarding

micahalcorn:
This implements a forked onboarding experience for first-time user of Origin Wallet. They can now import a wallet instead of having one automatically generated. It also adds support for managing multiple accounts within settings.

There are a couple of ui/ux issues that I will need to work out, but I think this branch is mostly ready to review.

PR #1747, opened 2019-03-20
Mobile Wallet messaging tab fix

### LINKED ISSUE #1682: Origin Wallet Messaging Tab
It's my understanding that the messaging tab does not work at all with DApp2 because of the way it links using an embedded script instead of a less-safe url parameter. This should be fixed as soon as possible. I see four options:

1️⃣ Use the `plink` proof-of-concept, which comes with security concerns
2️⃣ Fix DApp2 to work with the embedded script
3️⃣ Remove the promise of a messaging feature
4️⃣ Make the messaging tab a native experience once reading/writing is centralized

nick:
Closes #1682


micahalcorn:
Closing in favor of #1649 🙂

PR #1776, opened 2019-03-21
[Growth] Misc fixes

franckc:
### Description:
 - Fix bug where past events where included when computing a rule status
 - Fix bug where invalid invite code would cause welcome page to error out
 - Fix bug where invite code would sometime show as undefined

### Checklist:


franckc:
@sparrowDom Could you take another look ? I implement error management.
Since apollo already provides a few specific exceptions like UserInputError, AuthenticationError and ForbiddenError, I did not create growth specific errors. Happy to do that if you think of a reason it would be better.

sparrowDom:
@franckc I like your solution 👍

Issue #1117, opened 2018-12-20
ERC20 token address whitelist for marketplace contract


cuongdo:
A malicious or buggy ERC20 token could prevent an offer from finalizing by returning `false` from the calls to `transfer()` in paySeller ([call 1](https://github.com/OriginProtocol/origin/blob/master/origin-contracts/contracts/marketplace/v00/Marketplace.sol#L396), [call 2](https://github.com/OriginProtocol/origin/blob/master/origin-contracts/contracts/marketplace/v00/Marketplace.sol#L400)).

This is not currently an issue, because we only allow offers in ETH. However, we'll need to fix this before allowing ERC20 tokens to be used for offers.

A token whitelist is one way of fixing this issue.

Note: this is finding 3 in the audit

PR #1791, opened 2019-03-23
Whitelisting ERC20 token addresses

shahthepro:
### Description:

Changes made:

- Have added a new contract called `RestrictableContract` that inherits from `Ownable` contract.
- The contract contains two lists. One for whitelisted users and another one for whitelisted contracts. The methods `whitelistUser`, `whitelistContract`, `blacklistUser` and `blacklistContract` can be used add user/contract addresses to whitelist.
- The contract also adds the following modifiers on top of already existing modifiers from `Ownable` contract. 
    - `onlyWhitelistedContracts`
    - `onlyWhitelistedContractsOrOwner`
    - `onlyWhitelistedUsers`
    - `onlyWhitelistedUsersOrOwner`

### Some behavior by design:
- I wanted to keep the contract `RestrictableContract` as generic as possible so that it can be reused.
- Only owner of the contract can whitelist/blacklist user and contract addresses.
- I initially added `AddressUtil.sol` from zeppelin contract to check whether an address is a valid contract or not. However, ETH is internally represented using address 0x0 and 0x0 is not a valid contract. So, there is no direct way to whitelist 0x0. So I have commented out the code for now.
    
    We can manually whitelist address 0x0. But, since only owner of the contract can whitelist or blacklist, I don't think we have to be strict with whitelisting contract addresses. 

### Checklist:

- [ ] Write unit tests with 100% coverage
- [ ] Change name of the contract to a more reasonable one


shahthepro:
This addresses #1117

shahthepro:
Have created another draft PR with same changeset. #1792 

Closing this.

PR #1792, opened 2019-03-23
 Whitelisting ERC20 token addresses

### LINKED ISSUE #1117: ERC20 token address whitelist for marketplace contract

shahthepro:
### Description:

This changeset fixes #1117 

Changes made:

- Have added a new contract called `RestrictableContract` that inherits from `Ownable` contract.
- The contract contains two lists. One for whitelisted users and another one for whitelisted contracts. The methods `whitelistUser`, `whitelistContract`, `blacklistUser` and `blacklistContract` can be used add user/contract addresses to whitelist.
- The contract also adds the following modifiers on top of already existing modifiers from `Ownable` contract. 
    - `onlyWhitelistedContracts`
    - `onlyWhitelistedContractsOrOwner`
    - `onlyWhitelistedUsers`
    - `onlyWhitelistedUsersOrOwner`

### Some behavior by design:
- I wanted to keep the contract `RestrictableContract` as generic as possible so that it can be reused.
- Only owner of the contract can whitelist/blacklist user and contract addresses.
- I initially added `AddressUtil.sol` from zeppelin contract to check whether an address is a valid contract or not. However, ETH is internally represented using address 0x0 and 0x0 is not a valid contract. So, there is no direct way to whitelist 0x0. So I have commented out the code for now.
    
    We can manually whitelist address 0x0. But, since only owner of the contract can whitelist or blacklist, I don't think we have to be strict with whitelisting contract addresses. 

### Checklist:

- [ ] Write unit tests with 100% coverage
- [ ] Change name of the contract to a more reasonable one

### DIFF
--- a/packages/contracts/contracts/marketplace/v00/RestrictableContract.sol
+++ b/packages/contracts/contracts/marketplace/v00/RestrictableContract.sol
@@ -1,0 +1,78 @@
+pragma solidity ^0.4.24;
+
+import "../../../node_modules/openzeppelin-solidity/contracts/ownership/Ownable.sol";
+// import "../../../node_modules/openzeppelin-solidity/contracts/AddressUtils.sol";
+
+/**
+ * @title RestrictableContract
+ * @author Shahul Hameed <hello@hameid.net>
+ *
+ * Extends Ownable contract and maintains a whitelist of users and whitelist of contracts
+ */
+
+contract RestrictableContract is Ownable {
+    mapping(address => bool) public allowedContracts;
+    mapping(address => bool) public allowedUsers;
+
+    constructor() public {
+        // Set owner to contract creator
+        owner = msg.sender;
+
+        // Whitelist self
+        allowedContracts[address(this)] = true;
+    }
+
+    modifier onlyWhitelistedContracts(address contractAddress) {
+        require(allowedContracts[contractAddress], "Contract is not whitelisted");
+        _;
+    }
+
+    modifier onlyWhitelistedContractsOrOwner(address _address) {
+        require(_address == owner || allowedContracts[_address], "Must be a whitelisted contract or owner");
+        _;
+    }
+
+    modifier onlyWhitelistedUsers(address userAddress) {
+        require(allowedUsers[userAddress], "Must be a whitelisted user");
+        _;
+    }
+
+    modifier onlyWhitelistedUsersOrOwner(address userAddress) {
+        require(userAddress == owner || allowedUsers[userAddress], "Must be a whitelisted user or owner");
+        _;
+    }
+
+    // Add contract address to whitelist
+    function whitelistContract(address contractAddress) public onlyOwner {
+        // require(AddressUtils.isContract(contractAddress), "Must be a contract address AAA");
+        allowedContracts[contractAddress] = true;
+    }
+
+    // Remove contract address from whitelist
+    function blacklistContract(address contractAddress) public onlyOwner {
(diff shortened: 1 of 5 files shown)


packages/contracts/contracts/marketplace/v00/RestrictableContract.sol
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * @title RestrictableContract
+ * @author Shahul Hameed <hello@hameid.net>
+ *
+ * Extends Ownable contract and maintains a whitelist of users and whitelist of contracts
+ */
+
+contract RestrictableContract is Ownable {
+    mapping(address => bool) public allowedContracts;
+    mapping(address => bool) public allowedUsers;
```
franckc: What type of use cases related to a P2P marketplace do you anticipate would need a whitelist of users ?

shahthepro: I added that just in case if whitelisting of sellers/buyers are needed. Can remove it if it is not necessary.

shahthepro:
> Now, given that we already check at the DApp layer in JS against a whitelist of ERC20 token addresses, while nice, I don't think it is a very high priority to add those checks in the contract.

IMO, whitelisting of ERC20 tokens should be at contract level. Anyone can create a listing without using the DApp.

Doing this at the DApp level won't solve the problem.

joshfraser:
Just to be sure I'm understanding things correctly. If someone goes around our DApp and manually creates a listing or an offer that is denominated in an unrecognized token, it won't even show up in our DApp, correct?

While I appreciate the fact that people can read & write to our contracts directly, that's not a behavior we encourage. The DApp does a lot of work to try and keep people safe.

I think I agree that it's better to do this check at the contract level, but as long as there's no pressing security risk I'm okay punting on this for now for the reasons that @franckc outlined.

shahthepro:
@joshfraser 

> While I appreciate the fact that people can read & write to our contracts directly, that's not a behavior we encourage. The DApp does a lot of work to try and keep people safe.

I beg to differ. Users should be encouraged to both using DApp or interacting with contracts directly. DApp may do a lot of work to keep them secure. But unless it is hosted on IPFS or on Tor, the DApp isn't decentralized.

In this case, If the DApp goes down for some reason, the users will be forced to interact directly with the contracts (to accept/finalize/withdraw offers so that they can get their funds back).

However it is not something that will happen often. If there is no potential security risk, we can close this PR for now and reopen it when it becomes relevant.

nick:
I think we should close this for now and possibly come back to it in the future. I don't think we need to restrict ERC20 at the contract level... if users interact with our contract directly then they're on their own. Listings and offers with non-approved ERC20 just wouldn't show up in our dapp due to the JS level filters so this is only a problem if someone sets up an entirely separate marketplace contract and an entirely separate dapp.

PR #1809, opened 2019-03-27
Stan/auto pick chinese

wanderingstan:
Automatically switch to zh_CN if that's the language of user's browser.


micahalcorn:
Is this still necessary? 🇨🇳

franckc:
Yes I think this is relevant since @wanderingstan and I were discussing that we should turn on browser auto-detect now that we have pretty good coverage for translations in all languages...

wanderingstan:
Yes, once we have Anna's stamp of approval, we'll want to merge this so that Chinese users automatically see the Chinese version of the Dapp. 

...and we should add other languages to that whitelist array as we get good translations. I think Kay is doing Korean, and @franckc  do you want to do French? :)

franckc:
I pinged Anna and Mila on our internal Discord channel to ask them if we can whitelist Korean and Chinese...

franckc:
Checked with Anna and Mila. We are good to go to whitelist both Korean and Chinese.

wanderingstan:
@franckc ok, I added korean to the whitelist. 

Can someone give a ~~review~~stamp?

PR #1811, opened 2019-03-27
Conditionally show the rewards banner

joshfraser:
Avoid the infinite loop & don't show the rewards bannner if you're already on the rewards page.

Also removed the feature gate for Origin Rewards since it's now live.


micahalcorn:
Resolved a thing 🤷‍♂️

PR #1812, opened 2019-03-28
Find and make URLs in messages clickable

### LINKED ISSUE #591: Autodetect links in messaging
It may be a bit early but just wanted to mention this idea.  Any thoughts on autodetecting links so they are clickable in messages?  It may make the customer experience smoother if people are able to chat with each other and share clickable URL's.

Screenshot:
screen shot 2018-09-26 at 3 56 44 pm

(Note: SF team, feel free to remove this issue if you feel as though it is too early at this stage.  Happy to repost it in the future when we take a few more passes to improve the customer experience.)

shahthepro:
Fixes #591 

Have used a third-party package [react-linkify](https://github.com/tasti/react-linkify).


tomlinton:
Thanks @shahthepro! Just a note that the package doesn't need to be added to the `package.json` at the root of the monorepo. It'll get hoisted by lerna.

shahthepro:
Thanks @tomlinton! I have removed it from that file.

franckc:
@micahalcorn Looks good to you ?

micahalcorn:
Nice @shahthepro 👍 Can we easily change it so that the links open externally instead of navigating away from the conversation? (`rel="noopener noreferrer" target="_blank"`)

shahthepro:
@micahalcorn Done.

PR #1813, opened 2019-03-28
HTML templates for rewards emails

joshfraser:
Adds our new standard HTML template to the rewards emails. And sets the stage for when we add profile pictures to the emails.


infra/growth/src/templates/emailInvite.txt
```diff
@@ -1,7 +1,5 @@
 ${referrerName} has invited you to join Origin.
```
franckc: I think I had 2 newlines here on purpose. To force the content to show up in gmail with a newline. Otherwise gmail puts all the text all together, without any newline... Same comment for te reminder text email.

franckc:
Merging. Let's try this out !

PR #1830, opened 2019-03-30
Two-Week Finalization Window

micahalcorn:
As requested by @joshfraser and @matthewliu, this intends to shorten the finalization window from one year to two weeks. I recommend that we eventually prescribe different timeframes depending on the listing types or consider a different approach altogether.


micahalcorn:
As a short-term precaution, we could do a DApp-side condition on the seller's CTA for finalizing. We could hide it on fractional listings, for example, until two weeks after the end of the timeslot.

DanielVF:
I agree with a shorter finalization window, but we need to make sure it's really obvious in the DAPP UI that the timer is counting down and how much longer it has.

PR #1848, opened 2019-04-01
Verification Code Validation

### LINKED ISSUE #1834: Phone/Email Verification Code Validation
Once an email or phone verification code has been requested, users are often confused about the delay in receiving an email or text message. They sometimes end up typing their email address or phone number in the verification code field thinking that it did not get submitted. We should consider validating the input in this second modal so that only a six-digit number is a valid input.

shahthepro:
Fixes #1834 

Allows the user to type only numbers and restricting it to 6 characters at max.

### DIFF
--- a/dapps/marketplace/src/pages/identity/EmailAttestation.js
+++ b/dapps/marketplace/src/pages/identity/EmailAttestation.js
@@ -165,7 +165,7 @@ class EmailAttestation extends Component {
               if (this.state.loading) return
               this.setState({ error: false, loading: true })
 
-              const trimmedCode = this.state.code.trim();
+              const trimmedCode = this.state.code.trim()
 
               if (trimmedCode.length === 0) {
                 this.setState({
--- a/dapps/marketplace/src/pages/identity/EmailAttestation.js
+++ b/dapps/marketplace/src/pages/identity/EmailAttestation.js
@@ -164,6 +164,25 @@ class EmailAttestation extends Component {
               e.preventDefault()
               if (this.state.loading) return
               this.setState({ error: false, loading: true })
+
+              const trimmedCode = this.state.code.trim();
+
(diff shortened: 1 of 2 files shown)


dapps/marketplace/src/pages/identity/EmailAttestation.js
```diff
@@ -180,11 +180,14 @@ class EmailAttestation extends Component {
             <div className="my-3 verification-code">
               <input
                 type="tel"
+                maxLength="6"
                 ref={ref => (this.inputRef = ref)}
                 className="form-control form-control-lg"
                 placeholder="Verification code"
                 value={this.state.code}
-                onChange={e => this.setState({ code: e.target.value })}
+                onChange={e =>
```
nick: Can we show an error if the user tries to submit a non-numeric code instead of just disallowing non-numeric characters? User may wonder why they can't type otherwise

shahthepro: Sure. Will make that change

shahthepro: Done.

dapps/marketplace/src/pages/identity/EmailAttestation.js
```diff
@@ -164,6 +164,25 @@ class EmailAttestation extends Component {
...
+                this.setState({
+                  error: 'Verification code is required',
+                  loading: false
+                })
+                return
+              }
+
+              if (trimmedCode.length !== 6 || isNaN(trimmedCode)) {
+                this.setState({
+                  error: 'Verification code is incorrect',
```
nick: Perhaps say "Verification code should be a 6 digit number"

shahthepro: Done. :+1:

PR #1849, opened 2019-04-01
Stan/l10n cleanup

wanderingstan:
- Puts all translation files (except scripts) into `./translation` 
  - So no more `./translations` folder, instead fbt files are in `./translation/fbt`
- Decent README in that folder so we don't forget all this next time.
- Removed unused script and references to unused files.


micahalcorn:
@wanderingstan is https://github.com/OriginProtocol/origin/tree/stan/l10n-cleanup/dapps/marketplace#translation now redundant, or will it remain?

wanderingstan:
Good catch @micahalcorn , I'd removed that section but didn't get it in the commit. Just fixed it.

wanderingstan:
Thanks @franckc ! 

Crowdin only works in the `./translation/crowdin` folder (both reading and writing to), which this PR doesn't touch. The main thing was to create the `./translation/fbt` folder for the fbt-formatted files that are derived from crowdin. 

But this gets me thinking: there would be a decent argument to _not_ keeping `./translation/fbt` in source control at all -- they are just intermediate files used in generating the files in `./public/translations` (the folder the dapp actually uses). Any reason to keep them?

franckc:
Good point that since they are temporary files we don't need to commit those ./translation/fbt/*.json
The only reason to commit them would be that is developers run npm translate they'll see these files and may be unsure if those should be committed ? At the end of the npm translate commands should we clean up those temp files ? Maybe as par of npm fbt:clean ?

sparrowDom:
Yeah like that idea, either removing them at the end of `npm run translate` or perhaps adding them to `.gitignore` so it won't confuse developers when they will be running translate. I would chose the latter in case we anticipate we would need to inspect those intermediary files much. Otherwise both options seem equal.

wanderingstan:
Ok, I removed `./translation/fbt` from git and added a readme to the folder.

Added these files to be deleted by `run npm translate:clean`

@franckc : `translate:clean` is currently _not_ called as part of `translate`. I kinda feel thats okay, since it doesn't hurt anything, and people are always free to call clean afterwards. But could go either way.

PR #1797, opened 2019-03-25
Stablecoin

micahalcorn:
Opening this from @nick's branch to prepare for merge 😬 I'll also be committing here.

PR #1773, opened 2019-03-21
Stan/schemas to 2.0

wanderingstan:
Note: This PR should probably be merged in concert with (or as part of) @tomlinton 's work on updating EventListener to use GraphQL

- Update listing and offer schemas 
  - Support fractional listings
  - Change required fields. I.e. price no longer required, as it is absent from Announcements
  - Bump version to 2.0
- Remove "band-aid" code that re-wrote fractional listings in Dapp1 format
- Update sample listings (Zinc House, Scout, etc) to real fractional listings (2.0.0 format)
- Change sample listing creation to use mutations rather than hodge-podge of json

Issue #1417, opened 2019-02-06
Move listener code to separate origin-listener package


franckc:
Currently origin-discovery includes the code for both the discovery server and the listener.
It would be cleaner to separate the listener code out. I'm thinking the structure would look like:
 - origin-discovery: discovery server
 - origin-listener: listener server
 - origin-search: search specific code, used by both origin-listener and origin-discovery

franckc:
This could be a good meaty issue for someone looking at getting familiar with our back-end.

PR #1850, opened 2019-04-01
Update event listener for @origin/graphql

tomlinton:
This updates the event listener to use `@origin/graphql`. Important points:

- The data structure of what gets stored in postgres has completely changed for listings. These rows don't get used for anything right now so I think this is fine. It was easier to change the structure so it just stores what we get back from `@origin/graphql` returns rather than trying to coerce them into the old format.
- The listener still uses the same polling code, but calls `newBlock` from `@origin/graphql/src/contracts` to make sure that the event caches are kept up to date.
- The ES index has changed, the current one was wrong and needed to be updated.
- Removes no gas code from discovery and listener as we are taking an alternative approach and I think it will be too difficult/complex to maintain no gas via `@origin/js` while also switching to `@origin/graphql` for everything else.
- Some of this work is @wanderingstan's which updates the schemas to handle fractional hourly. I haven't tested fractional hourly yet. #1773 was the PR for that but the code is merged into this branch.
- Stablecoin is coming in #1797. I think this should all be OK, but something to keep in mind.

@DanielVF I know you are keen to help out. If you could take a general look at the approach that'd be great. Also testing it on staging or mainnet would be helpful. I think it might be less efficient now as both the listener and graphql are doing their own web3 stuff.

#1417 is a good follow up to this after it gets merged because dependencies between listener and discovery are probably at an all time low.


infra/discovery/src/listener/webhooks.js
```diff
@@ -40,11 +40,11 @@ async function postToDiscordWebhook(url, data) {
   }
 
   const icon = eventIcons[data.log.eventName] || ':dromedary_camel: '
-  const listing = data.related.listing
+  const listing = data.listing
```
franckc: The GCF for IPFS pinner relies on the related field so I think we'll need to update it.

packages/graphql/src/mutations/marketplace/createListing.js
```diff
@@ -4,13 +4,12 @@ import validator from '@origin/validator'
...
 import contracts from '../../contracts'
 import cost from '../_gasCost'
-import dapp1Compatibility from './_dapp1Compat'
 
 export function listingInputToIPFS(data, unitData, fractionalData) {
   const listingType = fractionalData ? 'fractional' : 'unit'
   const ipfsData = {
     __typename: data.typename || 'UnitListing',
-    schemaId: 'https://schema.originprotocol.com/listing_1.0.0.json',
+    schemaId: 'https://schema.originprotocol.com/listing_2.0.0.json',
```
franckc: Why is it necessary to change version from 1.0.0 to 2.0.0 ?
Is that going to create issues for reading/writing legacy listings created using 1.0.0 ?

Ahh.. is that @wanderingstan changes for hourly listings ?

wanderingstan: Yes, that was @nick and I. These are breaking changes to the schema for fractional listings. However, old listings should be automatically converted by the event sourcer, but we need to do a lot of testing on that.

infra/discovery/src/listener/handler_identity.js
```diff
@@ -125,20 +154,20 @@ class IdentityEventHandler {
...
+      identity.avatar.length > 0
     if (!validProfile) {
       return
     }
 
     // Record the event.
     await GrowthEvent.insert(
       logger,
-      user.address,
+      identity.address,
```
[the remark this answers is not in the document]
tomlinton: It is the result of the gql query defined in `queries/Identity.js`. It uses an id I think. Thanks for reminding me that I haven't tested this.

DanielVF:
Currently the event listener treats non-retryable validation errors just like it does transient errors. There's not much sense spending five minutes retrying "GraphQL error: Abstract type ListingResult must resolve to an Object type at runtime...".

It would be nice to just fail immediately. Essentially the only transient/retryable errors should be fetching from IPFS and fetching events from Infura?

tomlinton:
> Currently the event listener treats non-retryable validation errors just like it does transient errors. There's not much sense spending five minutes retrying "GraphQL error: Abstract type ListingResult must resolve to an Object type at runtime...".
> 
> It would be nice to just fail immediately. Essentially the only transient/retryable errors should be fetching from IPFS and fetching events from Infura?

Yep that makes sense. Do you want to implement that?

DanielVF:
Yeah, I'll implement that. I don't think that should hold up the rest of the PR though.

The listener code is nicely organized and the overall error handling seems very clean.

DanielVF:
Origin-js (at least a long time ago) would throw an error if it could not load the IPFS data for a listing. Our GraphQL will return everything it knows about a listing, even in cases when it can't load IPFS and almost all the listing data is empty. Or if the listing is missing required fields etc.

Because of this, it's important that when we store a the listing data returned from GraphQL, we also store if the listing data is valid. (our GraphQL provides a `valid` and a `validationError` field on each listing).

I've added storing these two fields in the database, and am now looking into filtering out invalid listings from results from the discovery server.

It's up for debate if the listener should notify webhooks and other systems about invalid listings.

My tasks:

- [x] Record valid/validationError in dB
- [ ] Verify that these are also being stored in ElasticSearch
- [ ] Make discovery server filter out invalid listings.

DanielVF:
We now ignore invalid listings from Discovery server searches. 

I've noticed that querying a listing price field in the discovery server results in errors. I don't think this holds up moving this PR forward, since the DAPP currently only queries the discovery server for listing IDs and loads the rest locally.

wanderingstan:
Nice work Tom 🎉!

PR #1907, opened 2019-04-04
Speed up listener with event caches

tomlinton:
This changes everything from processing logs to just processing events, it interfaces with the eventcaches directly to get new events. This is good so it uses the batching and optimisations of the eventcache rather than getting blocks itself.

WIP though!

### DIFF
--- a/infra/discovery/src/listener/listener.js
+++ b/infra/discovery/src/listener/listener.js
@@ -16,14 +16,12 @@ try {
   logger.warn('EnvKey not configured')
 }
 
-const express = require('express')
-const promBundle = require('express-prom-bundle')
 const Web3 = require('web3')
-
 const esmImport = require('esm')(module)
 const contractsContext = esmImport('@origin/graphql/src/contracts').default
 const { setNetwork } = esmImport('@origin/graphql/src/contracts')
 
+const { blockGauge, errorCounter } = require('./metrics')
 const { handleEvent } = require('./handler')
 const { getLastBlock, setLastBlock, withRetrys } = require('./utils')
 
@@ -34,7 +32,6 @@ class Context {
   constructor() {
     this.config = undefined
     this.web3 = undefined
-    this.networkId = undefined
   }
 
   async init(config, errorCounter) {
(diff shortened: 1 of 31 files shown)


infra/discovery/src/listener/listener.js
```diff
@@ -269,12 +127,63 @@ const errorCounter = new bundle.promClient.Counter({
...
+
+      // Flatten array of arrays filtering out anything undefined
+      const events = [].concat(...eventArrays.filter(x => x))
+      events
+        // Filter to only new events
+        .filter(event => event.blockNumber >= lastProcessedBlock)
+        // Process each new event
+        .map(newEvent => handleEvent(newEvent, context))
+
+      // Record state of processing
```
tomlinton: @nick would you mind checking out this as a way of getting new events?

nick: Looks good to me. I guess there is similar code inside the eventCache for checking toBlock, fromBlock etc.. so I'm not sure we need to duplicate it out here? Not sure though

tomlinton:
I think this is just about ready to go. If anyone wants to test you just need to copy `dev.env` from the listener directory to `.env` at the root of the discovery package directory and run `npm run start:listener` after adjusting the network setting in the `.env`.

tomlinton:
After testing on Rinkeby I've got 324 listings in my index with the new listener and https://dapp.staging.originprotocol.com shows 304. I will see if I can hunt down those last 20 tomorrow, or if anyone is keen feel free to take a look.

The listing order has also changed, but I think it is correct now. E.g. the [first listing](https://dapp.staging.originprotocol.com/#/listing/4-000-55) on staging appears to not be boosted at all, and there are boosted listings that should be higher.

infra/discovery/src/listener/listener.js
tomlinton: Yea it is similar but I don't think I can leverage it to just get new events like this, perhaps you know a way though.

DanielVF:
I love the new architecture.

tomlinton:
This is ready if someone wants to provide an approval!

PR #1944, opened 2019-04-05
Fix indexing of offer status

tomlinton:
Fixes offer status indexing which is `offerStr` not `offer` in graphql. The `offer` attribute is an integer representing the status.


franckc:
This is going out as part of a deploy I'm doing to staging/prod.

Should we re-index the data after deploy is done ?

tomlinton:
It is probably not worth it because we have to reindex for stablecoins which I think will probably happen on Monday.

PR #1945, opened 2019-04-05
Compress event cache to prevent localStorage overflow

nick:
the event cache is currently stored in localStorage but has recently gone over the limit, causing errors in the dapp.

This PR is a short term fix to compress the events before storing in localStorage.


nick:
I think we can push it but would want @tomlinton or @franckc to be around to check this over first

PR #1951, opened 2019-04-05
Hotfix

nick:
Hotfix

### DIFF
--- a/packages/graphql/src/resolvers/marketplace/_featuredAndHidden.js
+++ b/packages/graphql/src/resolvers/marketplace/_featuredAndHidden.js
@@ -27,7 +27,7 @@ export async function getFeatured(net) {
   if (queues.featured.fetching) await queues.featured.isDone()
   queues.featured.fetching = true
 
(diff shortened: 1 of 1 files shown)


packages/graphql/src/resolvers/marketplace/_featuredAndHidden.js
```diff
@@ -21,12 +21,12 @@ export async function getFeatured(net) {
   if (net === 'rinkeby') netId = 4
   if (!netId) return []
 
+  if (featured[netId]) return featured[netId]
+
   queues.featured = queues.featured || new Queue()
   if (queues.featured.fetching) await queues.featured.isDone()
   queues.featured.fetching = true
 
-  if (featured[netId]) return featured[netId]
```
franckc: Don't we actually need to keep this and put it in right after line 25 ?

nick: Ah yea you're right

Issue #1826, opened 2019-03-29
Google Attestation


micahalcorn:
Add the functionality to verify a Google account. It's been "coming soon" for about a year now. 😐 https://zpl.io/aMPeL3r

nick:
Done

PR #1964, opened 2019-04-08
Add Google Attestation

shahthepro:
Adds Google Attestation. #1826


nick:
Seems there's 1 failing test:

```
google attestations
    ✓ should generate a correct auth url
    x should generate attestation on valid verification code
    ✓ should error on missing verification code
```

shahthepro:
@nick Thanks. Have fixed the unit test and made those changes. :)

franckc:
Awesome ! Thanks @shahthepro for implementing this :)
Looking really good. Nice work.

Couple comments:
 - Let's store whether the user is Google verified in the identity table. Could you add a column in this table [here](https://github.com/OriginProtocol/origin/blob/master/infra/identity/src/models/identity.js#L14) and make sure it is populated by the listener [here](https://github.com/OriginProtocol/origin/blob/master/infra/discovery/src/listener/handler_identity.js#L132) ?
- Before we land this we need to figure what's the impact on Origin Rewards. Should users get rewarded for the verification of their Google account the same way as for other attestations ? And shall that be part of April campaign (e.g. we would need to update the campaign rules) or should it start in May campaign ? @matthewliu could you comment on this please ?

shahthepro:
@franckc 

> Awesome ! Thanks @shahthepro for implementing this :)
> Looking really good. Nice work.

Welcome :)

> Let's store whether the user is Google verified in the identity table. Could you add a column in this table here and make sure it is populated by the listener here?

I somehow overlooked it. Have made that change.

shahthepro:
@franckc will make that change :)

shahthepro:
@franckc Have made that change

dapps/marketplace/src/pages/user/Profile.js
```diff
@@ -186,7 +186,8 @@ class UserProfile extends Component {
                 )}
                 {this.renderAtt(
                   'google',
-                  fbt('Google', '_ProvisionedChanges.google')
+                  fbt('Google', '_ProvisionedChanges.google'),
+                  process.env.ENABLE_GOOGLE_ATTESTATION != 'true'
```
franckc: nit: in our codebase we use the !== operator rather than !=

franckc:
Thanks @shahthepro ! Minor nit inline otherwise looks great to me.
@nick if the DApp code looks good to you, let's merge this to master :)

Before we deploy to staging/prod, we'll need to add the following keys to EnvKey:
 - DApp: ENABLE_GOOGLE_ATTESTATION
 - Bridge: GOOGLE_CLIENT_ID and GOOGLE_CLIENT_SECRET
As our secrets guardian, perhaps @joshfraser could be help with creating and setting GOOGLE_CLIENT_ID and GOOGLE_CLIENT_SECRET ?

PR #1965, opened 2019-04-08
[Listener] Updates to main loop

franckc:
### Description:
 - Simply main loop by using do/while loop
 - Enforce processing of event in blockchain order
 - Move withRetry to event processing layer

### Checklist:


franckc:
I'm in the process of testing these changes but thought I would a preview to get feedback.

infra/discovery/src/listener/listener.js
```diff
@@ -101,66 +106,90 @@ const config = {
...
-      logger.debug(`Got ${newEvents.length} new events`)
-
+  do {
+    start = new Date()
+
+    // Compute the range of blocks to process,
+    // while respecting trailing block configuration.
+    const currentBlock = await context.web3.eth.getBlockNumber()
+    let toBlock = Math.max(currentBlock - context.config.trailBlocks, 0)
+    toBlock = Math.min(toBlock, processedToBlock + 10000) // At most process a chunk of 10k blocks at once.
```
tomlinton: The eventCache code out of `@origin/graphql` has its own chunking logic, so perhaps this is not needed here

infra/discovery/src/listener/listener.js
```diff
@@ -101,66 +106,90 @@ const config = {
...
-  check()
+    } else {
+      // Concurrency set to 1 -> Normal operation mode.
+      // Sort events by blockNumber and logIndex to process them in order.
+      // In normal operation (not backfill), ordering does matter for handlers.
+      // For example assume a buyer updates their identity to add their email
+      // then makes a purchase. If the handler would process the purchase
+      // event before the identity update, we may not be able to send the buyer
+      // a confirmation email about their offer.
+      newEvents = sortBy(newEvents, ['blockNumber', 'logIndex'], ['asc', 'asc'])
```
tomlinton: I think we should move this up so it happens in backfill mode as well. It's probably not a big performance hit and I think it is good if we keep the logic as close as possible for backfill vs non backfill. For example, if the listener fails and we need to run with concurrency to catch up and we are missing this logic the issue we describe in the comments will still arise.

infra/discovery/src/listener/listener.js
franckc: My intent was so that in backfill mode where large number of blocks need to be processed, we would persist the progress in the listener table regularly. But perhaps not worth the extra complexity. Removed.

infra/discovery/src/listener/listener.js
franckc: The problem is that there is no way to guarantee processing when using Promises, right ?
Like when Promise.all(...) is called, the events are processed in parallel without guaranteed ordering. Or maybe I'm missing some fancy JS trick ?

tomlinton: Yea you are right! Maybe just something to keep in mind then

PR #1968, opened 2019-04-08
More request optimizations

nick:
A few more optimizations and simplifications for reducing number of ethereum node requests


franckc:
Nicely done ! Love this simpler and faster code :)

Issue #1947, opened 2019-04-05
Improvements to Event Cache


nick:
The Origin marketplace dapp is able to function with only 2 dependencies: an Ethereum node and an IPFS server. Every time a user takes an action such as creating a listing or making an offer, an event is emitted on the [Marketplace contract](https://github.com/OriginProtocol/origin/blob/master/packages/contracts/contracts/marketplace/v00/Marketplace.sol#L24-L39) that follows the following structure:

`Event(index party, indexd listingId, indexd offerId, ipfsHash)`

The first 3 parameters as well as the name of the event are 'indexed', meaning they can be filtered on when requesting data from the Ethereum node. These indexed parameters are known as 'topics'. topic 0 is the hashed event name and topics 1-3 are the parameters marked as 'indexed'.

Having events indexed in this way is a useful property as we can now easily request events pertaining to a certain individual or particular listing or offer. 

For example, to retrieve all events related to listing 345, we would issue a query for all Marketplace events with topics `*, *, 345, *` (any event name, any party, listing 345, and any offer). To retrieve all events created by a particular user, we would query for `*, userWallet, *, *`. To retrieve all listings created by a particular user we would query for `ListingCreated, userWallet, *, *`.

In order to build up the current state of a listing or offer, Origin has an [event source](https://github.com/OriginProtocol/origin/tree/master/packages/eventsource) module which, given a listing ID and/or offer ID will first fetch all related events from the Ethereum node, then pull all related data from IPFS, then construct the current 'state' of the listing or offer by using what is essentially map/reduce.

This presents several problems:

- There can be a lot of duplicate events, eg requesting events for a particular user or listing can result in the same event being returned twice. It would be better if we could only request data we know we don't already have.
- Requesting events from long ago is inefficient due to the Ethereum node having to look in every single block to see if any events were emitted on the given contract. In order to combat this, events need to be queried in batches.

In order to address these issues, we have an [eventCache](https://github.com/OriginProtocol/origin/blob/master/packages/graphql/src/utils/eventCache.js) module which does the following:

- Requests all events ever emitted on the Marketplace contract in batches of [20,000 blocks](https://github.com/OriginProtocol/origin/blob/master/packages/graphql/src/utils/eventCache.js#L13)
[...]

PR #1975, opened 2019-04-09
event-cache package

mikeshultz:
### Description:

See: #1947

#### Usage

    import { patchWeb3Contract } from 'event-cache'
    
    const identity = IdentityEvents.deploy().send({ from: '0x0b44611b8ae632be05f24ffe64651f050402ae01' })
    patchWeb3Contract(identity)

    // Get all ident updates from a specific user
    const events = await identity.eventCache.getPastEvents('IdentityUpdated', {
      filter: { account: alice }
    })

### Checklist:


franckc:
Looking good @mikeshultz !

I wonder if we should re-use this existing event table ?
https://github.com/OriginProtocol/origin/blob/master/infra/discovery/migrations/20181023225253-event-create-table.js

Currently it is being populated directly by the event listener (see this [code](https://github.com/OriginProtocol/origin/blob/master/infra/discovery/src/listener/handler.js#L59)).
But the event cache could take care of that.
This table is not being used by any service. I initially added to help with debugging/troubleshooting.

mikeshultz:
Was thinking about using a simple schema for this, doing most operations on a JSONB column for the event object.  I can potentially use the `data` col for that, but that might buck the previous usage.  My first impulse says something like this:

    CREATE TABLE event_cache (
        event_id SERIAL PRIMARY KEY,
        address VARCHAR(40) NOT NULL,
        event VARCHAR NOT NULL,
        event_object JSONB NOT NULL
    )

Either way, using the JSONB blob or column for every event prop, I'll want to at a minimum alter this table schema for this use.

franckc:
Re event table, might be handy to have topic columns with an index on them. To be able to do fast filtering based on those, similarly to web3 providing filtering based on topics...

mikeshultz:
It's my understanding that since version ~9, JSONB queries are pretty performant.  Also, these event objects are decoded and event data can be directly used with their decoded type(though raw topics are in there as well).

That said, if you have a use case for it I can make sure that goes into the table.

franckc:
I think we have use cases where we want to filter by listingId, offerId and those are topics.
Yes, queries on fields within JSONB are pretty fast but the only thing is you can't have a DB index on those fields. So if we ever reach scale at some point we will likely need that otherwise a filter by listingId/offerId would cause a full table scan.

DanielVF:
Almost everything we'll ever do will be use these basic indexes - I'd be happy having them at the DB level.

mikeshultz:
Good point on the indexing, that didn't cross my mind for some reason.  I'll see about flattening the event object into a table schema.  The decoded event args will probably have to be in a JSONB column, however, since each event will have a different set of args.

When I get a table schema together I'll post it in discord and ask you guys to tear it apart.

mikeshultz:
I think this is pretty much ready for review.  I'd like to expand tests a little bit but don't want to burn too much of my allotted time on that without at least a first review.

There's also a change to the discovery server in this PR, so might want to pay special attention to that.

packages/event-cache/src/backends/PostgreSQLBackend.js
```diff
@@ -0,0 +1,205 @@
...
+   * Stores a single event
+   *
+   * For more info on the eventObject, see: https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#contract-events-return
+   *
+   * @param eventObject {object} A JS object representing the event
+   */
+  async addEvent(eventObject) {
+    const dbObject = eventToDB(eventObject)
+    await this._models.Event.upsert(dbObject)
+    this.setLatestBlock(eventObject.blockNumber)
```
franckc: This is a general question, not specific to the Postgres implementation.
Is it always the case that events will get added in blockNumber order ?
I wonder for safety purpose if we should check that eventObject.blockNumber > latestBlockNumber before calling setLatestBlock. Or alternatively have that check within the setLatestBlock method,

packages/event-cache/src/backends/PostgreSQLBackend.js
```diff
@@ -0,0 +1,205 @@
...
+    dbObject['data'] = dbObject['raw']['data']
+    if (dbObject['raw']['topics'] instanceof Array) {
+      for (let i = 0; i < dbObject['raw']['topics'].length; i++) {
+        if (i > 3) break
+        dbObject['topic' + i] = dbObject['raw']['topics'][i] || null
+      }
+    }
+    delete dbObject['raw']
+  }
+
```
franckc: How does the createdAt field get set ?
I think in the old implementation we had it set based on the block timestamp.

mikeshultz:
Thanks @franckc

> Regarding deploying the event-cache for use by the listener, I did not see how the EventCache would get configured in the listener ? Is that piece still under development ?

Check out [the README](https://github.com/OriginProtocol/origin/tree/mikeshultz/event-cache/packages/event-cache).  It will require integration into projects that use it, but it should be minimal.  I'd be happy to make those changes if people point me in the right direction.  Might be a good idea to start with that discovery listener, maybe?

Also, had a thought last night before going to sleep that there's probably an issue with installing this for client-side use right now.  I'm not super experienced in JS packaging, so maybe someone can help me out here.  I know at least `pg` is a native module, which can't be installed for browser.  While it doesn't get imported unless the pg backend is initialized, it could still be an issue when installing?

packages/event-cache/src/backends/PostgreSQLBackend.js
mikeshultz: I think for at least the postgres backend, that should be pulled from storage.  That was something I wanted to do but forgot about.

That's significantly less performant for the other backends.  Once the cache is "caught up", that shouldn't be an issue, but I agree in at least adding a check to not allow the block number to be reduced.  (e.g. `if (blockNumber < this.latestBlock) return`)

Thanks, I'll have a fix up this morning

packages/event-cache/src/backends/PostgreSQLBackend.js
mikeshultz: This was an explicit decision I made, for better or worse.  I did notice that the previous listener set the date, but didn't really think that was important here.  It doesn't match the API of web3's `getPastEvents` or that traditional use case, so I wasn't expecting anyone to query on it.

Right now the backends are also completely oblivious to web3.  To set the actual block stamp, I'll need to fetch the block.  Since the block stamp isn't technically an attribute of an event log, I was hoping not to do this, and that anything that needed to deal with time could fetch the block (or range) itself if necessary before making the request.

So, I guess the question is, is this functionality or data important?  Are there going to be other services/apps that are going to access this table directly?  If so, should I have a unique extra attribute added to these event objects before storage and stripped off upon fetch?

Ideally, or at least I think the intention of this package was to mimic the web3 API and be used in place of the regular un-cached calls.

mikeshultz:
> Regarding the migration to create the new event table, sicne event-cache is a package and not a service, we'll need to figure out as part of which container deployment its migration gets run.
I'm thinking perhaps we make it part of the listener container deployment here

Added the migrate command to the `event-listener` container.  I think that order should be good so the drop gets run before the create.

nick:
A quick note that we need support for fetching events for multiple event types at once. For example [this call](https://github.com/OriginProtocol/origin/blob/master/packages/graphql/src/resolvers/User.js#L133-L145) expects to return any event in a given list initiated by certain user, eg:

```
eventCache.getEvents({ event: ['OfferCreated', 'OfferFinalized'], party: "0xabc" })
```

or:

```
eventCache.getEvents({ listingID: ["5", "6", "7"] })
```

mikeshultz:
Ton of updates, including:

- Dapp/graphql integration
- Ordered output in chain history ascending order
- IndexedDB fixes

Travis is happy, and local tests of the dapp look good.

PR #1981, opened 2019-04-09
A few changes

nick:
Ensure fingerprinting only happens once. 
Add currency fields to growth query

### DIFF
--- a/dapps/marketplace/src/queries/Fragments.js
+++ b/dapps/marketplace/src/queries/Fragments.js
@@ -194,11 +194,19 @@ export default {
           status
           rewardEarned {
             amount
-            currency
+            currency {
+              ... on Currency {
+                id
+              }
+            }
           }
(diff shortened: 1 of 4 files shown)


dapps/marketplace/src/queries/Fragments.js
```diff
@@ -194,11 +194,19 @@ export default {
           status
           rewardEarned {
             amount
-            currency
+            currency {
+              ... on Currency {
```
franckc: Hmm... we also need to change the back-end to return Currency since it got changed from Price to Currency. I can send a separate PR

franckc:
We discussed this in Discord. The currency is encoded in the rules so it is a bigger changes than anticipated to switch to CurrencyResult in growth. For the short-term, we are going to use String rather than CurrencyResult - See #1983

PR #1994, opened 2019-04-09
Nick/stablecoin fixes

nick:
Fix issue where exchange rate causes too many decimal places


micahalcorn:
➕ Micah 😇

dapps/marketplace/src/components/WithPrices.js
```diff
@@ -2,6 +2,11 @@ import withCurrencyBalances from 'hoc/withCurrencyBalances'
 import withWallet from 'hoc/withWallet'
 import get from 'lodash/get'
 
+// web3.utils.toWei only accepts up to 18 decimal places
+function removeExtraDecimals(numStr) {
```
franckc: Works. An alternative could be to use
BigNumber.set({ DECIMAL_PLACES: 18 })

Issue #2010, opened 2019-04-11
Localize text for stable coin


wanderingstan:
Looks like none of the new strings got wrapped for stable coin release. 

I'm getting the low hanging fruit now, but don't have a lot of time, so we should do a full sweep later.

micahalcorn:
Really? I thought I did this @wanderingstan. Let me take it off your hands if so.

micahalcorn:
Lots of `<fbt` in [here](https://github.com/OriginProtocol/origin/pull/1797/files) but we might have missed some. 🙁

wanderingstan:
Ah, I just saw there was no wrapping done in here:
https://github.com/OriginProtocol/origin/blob/master/dapps/marketplace/src/pages/listing/mutations/Buy.js

@micahalcorn no worries, I'm already deep in it from trying to fix those "just a few fixes" :) on growth. I'll have a PR up in a sec. Just testing now.

PR #2011, opened 2019-04-11
Stan/l10n wrapping

wanderingstan:
Wrapping text primarily for stable coin and growth, but lots of other random strings found once I started doing global searches. 

Fixes: https://github.com/OriginProtocol/origin/issues/2010#event-2270141912

### Checklist:


dapps/marketplace/src/pages/transaction/mutations/WithdrawOffer.js
```diff
@@ -114,7 +114,7 @@ class WithdrawOffer extends Component {
           <div className="make-offer-modal">
             <div className="success-icon" />
             <div>
-              <fbt desc="WithdrawOffers.success">Success!</fbt>
+              <fbt desc="success">Success!</fbt>
```
micahalcorn: Thanks @wanderingstan 🙏 Did we move away from the convention of using the name of the component here?

wanderingstan: Along with @franckc , we agreed that for short, widely-used strings like "Success!", we just use the word itself as the `desc`. Otherwise translators would see the same string numerous times.

franckc: +1

wanderingstan:
I get so OCD on these things...but they are nice "small wins" on the side when slogging through bigger projects. :)

dapps/marketplace/src/pages/transaction/mutations/WithdrawOffer.js
micahalcorn: Oh, I thought CrowdIn was supposed to recognize the string itself and avoid redundant, potentially conflicting translations. 😕

wanderingstan: If the desc is different, crowdin assumes that the word/phrase must be being used in a different context, and thus creates a separate entry for translators. For things like button titles and boilerplate messages ("Success!") its best to keep them all under one entry, so things are consistent.

PR #2030, opened 2019-04-12
Email notifications

wanderingstan:
Update notifications server to send emails for events. Currently just handles same set of events as old notifications server, with the same limited messages. 

Unsubscribing from emails is just handled by sendgrid. 

Different events may be tested by using the sample events in the fixtures directory. The test_post.sh script will make a POST to http://localhost:3456/events with the json file passed.

Example:

    cd test/fixtures
    ./test_post.sh OfferWithdrawn.json

Also when testing, it can be useful to direct all emails to yourself. This can be done with the `--override-email=stan@originprotocol.com`

Example:

```
./node_modules/.bin/nodemon src/app.js --override-email=stan@originprotocol.com --email-file-out=rinkeby
```


tomlinton:
Looks good to me! It will be handy having those fixtures around. The only thing I might suggest is replacing the console.log calls with logging via the `logplease` package which is what we've used in some of the other packages.

If it is fine with you I'll work on the mobile part of this when this gets merged. The linking stuff is getting deprecated so it'll all get handled by this server.

infra/discovery/src/listener/listener.js
```diff
@@ -66,7 +66,7 @@ const config = {
   // Unique id. Used to differentiate between the several listeners instances
   // that may run concurrently (ex: main vs webhook vs re-indexing).
   listenerId: args['--listener-id'] || process.env.LISTENER_ID || 'main',
-  // Notification web hook URL.
+  // Notification web hook URL. e.g. http://localhost:3456/events
   webhook: args['--webhook'] || process.env.WEBHOOK,
```
sparrowDom: should we rename this param to `notificationWebhook`? It would seem more natural to me. Don't know what if that kind of change would require some `GoogleCloud` settings update.

infra/notifications/src/app.js
```diff
@@ -73,17 +107,23 @@ app.use(bodyParser.json({ limit: '10mb' }))
...
-    const subs = await PushSubscription.findAll()
-
-    markup += `<h3>${subs.length} Push Subscriptions</h3><ul>${subs.map(
-      s =>
-        `<li><pre style="white-space: pre-wrap;word-break: break-all">${JSON.stringify(
-          s
-        )}</pre></li>`
-    )}</ul>`
+  let markup =
+    `<h1>Origin Notifications v${process.env.npm_package_version}</h1>` +
```
sparrowDom: pretty cool, didn't know this env var is available  👍

infra/notifications/src/app.js
```diff
@@ -73,17 +107,23 @@ app.use(bodyParser.json({ limit: '10mb' }))
...
+    `<p><a href="https://github.com/OriginProtocol/origin/issues/806">Learn More</a></p>`
+
+  try {
+    if (app.get('env') === 'development') {
```
sparrowDom: In other places we are using `process.env.NODE_ENV`. Should we also here?

infra/notifications/src/app.js
```diff
@@ -123,152 +163,86 @@ app.post('/', async (req, res) => {
...
-        }
-      }
-    })
+  const party = returnValues.party.toLowerCase()
+  const buyerAddress = buyer.id ? buyer.id.toLowerCase() : null
+  const sellerAddress = seller.id ? seller.id.toLowerCase() : null
+
+  console.info(`Info: Processing event ${eventDetailsSummary}`)
+
+  console.log('party:')
```
sparrowDom: Besides this being logs like @tomlinton points out. It might be more practical to have all 6 lines in 1 line. 
```
logger.info(`party: ${party} buyerAddress: ${buyerAddress} sellerAddress : ${sellerAddress}`)
```
This way it is easier to parse logs in the future in case we will want to extract some kind of statistics

infra/notifications/src/browserPush.js
```diff
@@ -0,0 +1,83 @@
...
+//
+async function browserPush(
+  eventName,
+  party,
+  buyerAddress,
+  sellerAddress,
+  offer
+) {
+  console.log('🖥 Browser Push')
+  if (!eventName) throw 'eventName not defined'
```
sparrowDom: MobilePush, browserPush and emailPush all check for the presence of these properties. Might make sense to move it out into one place. Perhaps at the bottom of `src/app.js` where they are all called

infra/notifications/src/models/identity.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
```
sparrowDom: To not duplicate stuff I think it would be better to import this identity: https://github.com/OriginProtocol/origin/blob/master/infra/identity/src/models/identity.js

This way if the database model changes, we are less likely to cause errors since we only have to change it in 1 place

infra/discovery/src/listener/listener.js
wanderingstan: I agree. It even took me a while to be sure this was the particular webhook related to notifications. I'll look around to find places where this param/env vars are used. 
@tomlinton - Let me know if there are hiding places where I should look.

micahalcorn: For what it's worth, that wasn't the original intent of the webhook. I think notifications was just the first consumer of it after Discord, which ended up becoming its own param. 🙂

wanderingstan: @sparrowDom : changed name here: https://github.com/OriginProtocol/origin/commit/a9c8705343cd8157b5cdcdf4e3bd22e3d0bdac8d
Was torn, but went with `notificationsWebhook` (plural) to match the name of our package/directory.

infra/notifications/src/app.js
wanderingstan: Looks like I left in wayyy to many debugging logs. Removed the excess, back to previous logging levels. 
See bec330036d6d5a24eb3aa208b86f982dadd268fd

infra/notifications/src/app.js
wanderingstan: Agreed. Fixed in 6a6df603c50c9e382fad75267f549d4d47e96db4

infra/notifications/src/browserPush.js
wanderingstan: I think these guards are useful: E.g. in development I had a typo and was accidentally passing in a null value as one of the addresses. This caused things to fail in subtle ways that were a pain to debug. Better for these functions to protect themselves against being calling with invalid params.

infra/notifications/src/models/identity.js
wanderingstan: Agreed. @franckc and I talked about re-working how we handle models that are used across packages/apps. I lean towards having a central place where all our db models and migrations are stored. Right now its such a patchwork, making migrations hard, plus making it difficult to know what models are used in which packages.apps.

wanderingstan: Oh, and yes @sparrowDom , changed code to use the one Identity model in https://github.com/OriginProtocol/origin/commit/623fa7431d33a0fe02bbe597e4fff1bcb1858096

wanderingstan:
@tomlinton good call on Logger. Added in bf5b50cf54d5c7bac00ede7a101a6ec1553bf8b6

PR #2032, opened 2019-04-13
Commit ganache state

tomlinton:
This is a wacky idea I had. I'm having trouble with the graphql network configs because things like the event-listener and react-native can't work with the localhost config because it relies on `window.localStorage`. So I'm throwing this out as an idea: run contract deployment on Ganache and commit the Ganache database to our repo and then hardcode the addresses.

The advantages:

- Addresses can be hardcoded in `@origin/graphql/src/contracts` so the config works the same for all networks (no reading from localStorage or truffle).
- I can remove the truffle deployment/complexity from Docker startup.
- Easier to reason/test things because all contract addresses are known ahead of time.
- We don't have to keep truffle in sync. I.e. the contract population script Nick wrote now does more than truffle and truffle has not been kept up to date.

Disadvantages:

- A bunch of random files in the repo.

Basically to update this we just remove `packages/services/data/db`. Open up the DApp and hit the populate button and then commit what gets created in `packages/services/data/db`.

Any thoughts?


nick:
Yea we need a way to deploy contracts and persist the contract addresses for sure. My vague plan was to alter the [populate](https://github.com/OriginProtocol/origin/blob/master/packages/graphql/fixtures/populate.js) script to return all the deployed contract addresses at the end, then add an extra method to `@origin/services` to call that script and save the results to a .gitignored config file. `@origin/services` would then check for the existence of that file on startup and run populate if it wasn't there. contract.js would just append the config into it's 'docker' property.

Would something like that do the trick? Happy to work on it as a priority if this is a blocker.

tomlinton:
That sounds like a nicer solution, just for the docker config or also for the localhost config? 

If we could write the config to the root of the monorepo that'd work great, metro/react-native doesn't support symlinking of packages so if I could set the contract config path by environment variable that should work well.

It's a bit of a blocker as it makes it hard to test mobile and stablecoins with localhost/docker. I can do work using Rinkeby though (or Origin testnet, I think you deployed the uniswap contracts there?). Also happy to implement this if you've got stuff on your plate.

PR #2034, opened 2019-04-13
Add puppeteer test for editing a users profile.

DanielVF:
@nick I've got a puzzling thing going on when adding a the puppeteer test for identity.

I've added a test for a basic profile edit. However, the test always fails when it publishes the changes to the blockchain. After debugging, it looks like the problem is that the identity events contract address is not there when publishing the change, even though I see in the debug logs that the contract was created.

Some debugging logs here:

https://gist.github.com/DanielVF/00286830854f47860c9000820f9c08e0


nick:
Should be fixed now... I pushed a small change to this branch.

Thanks for writing these tests!

DanielVF:
@nick, I merged in master, which fixed the unrelated failing test.

PR #2048, opened 2019-04-16
Mobile 1.5

tomlinton:
About time to merge this and start putting out some test releases!

- Replaces @origin/linking with @origin/mobile-bridge which handles web3 transactions between the DApp webview and react-native using window.postMessage. 
- Removes all the linking code from the DApp.
- This is the end of @origin/js. RIP 💀 

Next work:

- Do test releases on iOS and Android.
- Notifications, we need to get #2030 in and I need to do some work on the endpoints for saving device tokens.
- Try to break it.


nick:
Looks good! I reviewed the dapp and graphql changes but didn't spend much time on mobile as it's such a big change. Going to merge it in now so we don't end up with too many conflicts.

PR #2056, opened 2019-04-16
Fix logger import

tomlinton:
Fix missing logger module error. @wanderingstan this probably worked locally because you had an ENVKEY variable set so it never hit the logger call.


wanderingstan:
@tomlinton it turned out that I hadn't committed my `logger.js` file at all, which is why things worked locally. 🤦‍♀️

But good to move logger import higher up.

PR #2062, opened 2019-04-18
Fix notifications container

tomlinton:
Needed a bit more than I thought @wanderingstan. This should make it go!


wanderingstan:
Thanks @tomlinton !

Issue #2042, opened 2019-04-15
Improve Search Results


micahalcorn:
Our listings search feature does not produce the most helpful results. For example, [a search for "Origin Event Shirt"](https://dapp.originprotocol.com/#/?q=Origin%20Event%20Shirt) reveals 265 listings, and the exactly matching "Origin Event Shirt" is the 12th result. I don't know whether the 11 results ahead of it are promoted for some other reason, or we're just not sorting results based their degree of match. Additionally, there are some [extremely fuzzy match](https://dapp.originprotocol.com/#/listing/1-000-1999)es that arguably should not be returned or should be buried.

DanielVF:
It should be easy to make this a bit better.

https://www.elastic.co/guide/en/elasticsearch/guide/current/proximity-matching.html

I'll update the query tonight or tomorrow.

DanielVF:
PR pushed that makes a search for "Origin Event Shirt" return just that listing.

DanielVF:
"Origin Event Shirt" now only shows one listing:

<img width="570" alt="image" src="https://user-images.githubusercontent.com/837/56434006-e81cc600-62a0-11e9-9afb-f0c8f034d7a6.png">

In the longer term, some more possibilities are:

- [ ] not show sold listings by default in DAPP search results.
- [ ] Reprioritize cheap items (say less than 50 cents worth)
- [ ] Inactive older items rank slightly lower after X time with no events.

micahalcorn:
Thanks @DanielVF! 💙

PR #2069, opened 2019-04-19
Reduce irrelevant search results

DanielVF:
### Description:

Currently our search results contain a lot of irrelevant listings, often ahead of the obvious choice. (see #2042)

Primarily this is because we were selecting any listing that contained *any* word from the search query. This placed many irrelevant featured listings ahead of real results, and adds in many listings with no relationship to the search phrase.

This PR makes two changes.  

1) When using multi-word searches, listings must contain almost of the query words in order to match. Searching for "origin event tshirt" will now require that a listing contains all three of those words, instead of matching both all listings with "tshirt" and all listings with "origin". I've set it to require all words, up to four words, then after allow 20% of the words not to match. Hopefully this number is a good tradeoff, but regardless it should be massively better.

2) We boost listings that contain the words from the search phrase in close proximity. A listing that said "Event T-shirt" would get boosted ahead of a listing that talked about "event pricing" in one area of the description, and "vendors will be selling custom T-shirts" later on.

### Checklist:

### DIFF
--- a/infra/discovery/src/lib/search.js
+++ b/infra/discovery/src/lib/search.js
@@ -134,7 +134,8 @@ class Listing {
         match: {
           all_text: {
             query,
-            fuzziness: 'AUTO'
+            fuzziness: 'AUTO',
+            minimum_should_match: '-20%' // most query tokens must be in the listing
           }
         }
       })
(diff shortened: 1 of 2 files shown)


micahalcorn:
@DanielVF it looks like this causes a test to fail, but otherwise, I trust it. 😁

franckc:
Very cool. Thanks @DanielVF for improving our search ranking !

DanielVF:
Thanks @tom for fixing that failing test!

sparrowDom:
nice fix @DanielVF

PR #2075, opened 2019-04-19
Display errors in html

tomlinton:
For debugging mobile where console can't be seen. Maybe hide by default and make it display on click?


micahalcorn:
Oh, @tomlinton [here is the existing way that errors are accessed](https://github.com/OriginProtocol/origin/blob/ab5c421aec5ed8f3ccd39d77baafaac302a36843/dapps/marketplace/src/components/TransactionError.js#L47-L51).

Issue #1928, opened 2019-04-04
Origin Rewards Banner Button Wrapping


micahalcorn:
The "Go" button wraps and gets partially cropped on very narrow screens or when the banner copy is translated into more verbose languages.

![Screen Shot 2019-04-04 at 4 17 17 PM](https://user-images.githubusercontent.com/273937/55594701-79166d80-56f5-11e9-8820-6c278a84bc16.png)

micahalcorn:
Resolve (roughly) by #2082 ✅

PR #2082, opened 2019-04-20
Rewards Banner Wrapping

micahalcorn:
This is a quick fix for #1928 that removes the word "Free" from the call-to-action as requested by @matthewliu (I believe). It also allows for wrapping of not only the button but also the text for more verbose languages. We will need a better pattern for allowing the user to dismiss the banner or something because it's going to consume a significant portion of the screen real estate on small devices.

I also generated new translation files, so it would be great if we could get those new strings translated (to Korean) before doing our upcoming workshops @franckc. I gave @kayyoo23 a heads-up.

🇰🇷


franckc:
@micahalcorn I'll make a PR on Monday morning 4/21 to pull strings from crowdin into the Dapp once they are translated. I have been holding off so far because it does not look like we got much of the new strings translated: we are currently at only 78% of strings translated to Korean...

PR #2087, opened 2019-04-21
Add purchase confirmation modal

### LINKED ISSUE #2027: Purchase Confirmation Modal
E-commerce buyers are accustomed to having an intermediate step between expressing intent to purchase and having a completed order. Even Amazon's "one click ordering" involves a confirmation modal. If we were going deep on the e-commerce use case, we might introduce a "cart" concept and buyers would add items to it before moving to "checkout". We're not quite ready for that, but @coleman415 and others have indicated that it is jarring to have MetaMask pop up and attempt to move a user's funds before she has a chance to "review" her purchase.

The design at https://zpl.io/a8wOl4g introduces a modal that would be displayed prior to MetaMask prompting the user for a `makeOffer` request. Since many users will be routed through user activation only to be dumped back on the listing detail page, ideally we would just present this modal if we know they had previously expressed intent to purchase during the same session. That would save them an extra click where we're introducing yet another.

shahthepro:
Closes #2027


micahalcorn:
Bump @nick @shahthepro 🙂

shahthepro:
@micahalcorn I have resolved merge conflicts. Can be merged if @nick is fine with it.

shahthepro:
@nick As discussed, closing this now. Will include the modal in my PR for #2413.

PR #2101, opened 2019-04-22
Stan/simple notifications

wanderingstan:
This adds links to the outgoing notification emails, taking user to the offer/purchase the email is about. 

Turns off mobile and browser push for now.

This is temporary minimal messaging template until I'm done with the aure-styled templates.

### DIFF
--- a/infra/notifications/src/app.js
+++ b/infra/notifications/src/app.js
@@ -62,18 +62,18 @@ const config = {
   verbose: args['--verbose'] || false
 }
 
+logger.log('Configuration:')
 logger.log(config)
 
 // ------------------------------------------------------------------
 
-// should be tightened up for security
+// TODO: Should be tightened up for security
 app.use((req, res, next) => {
   res.header('Access-Control-Allow-Origin', '*')
   res.header(
     'Access-Control-Allow-Headers',
     'Origin, X-Requested-With, Content-Type, Accept'
   )
-
   next()
 })
 
@@ -227,10 +227,10 @@ app.post('/events', async (req, res) => {
   res.status(200).send({ status: 'ok' })
 
   // Mobile Push (linker) notifications
-  mobilePush(eventName, party, buyerAddress, sellerAddress, offer)
+  // mobilePush(eventName, party, buyerAddress, sellerAddress, offer)
 
   // Browser push subscripttions
-  browserPush(eventName, party, buyerAddress, sellerAddress, offer)
+  // browserPush(eventName, party, buyerAddress, sellerAddress, offer)
 
   // Email notifications
   emailSend(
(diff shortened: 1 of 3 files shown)


infra/notifications/src/emailSend.js
```diff
@@ -85,6 +85,16 @@ async function emailSend(
           logger.info(`No message found`)
         }
       } else {
+        const vars = {
+          dappUrl:
+            listing.id[0] === '4'
```
franckc: This works only for staging and mainnet. Do we care about local and dev environments ? Might be nice for testing purposes to support those as well ?

infra/notifications/src/app.js
```diff
@@ -62,18 +62,18 @@ const config = {
   verbose: args['--verbose'] || false
 }
 
+logger.log('Configuration:')
 logger.log(config)
 
 // ------------------------------------------------------------------
 
-// should be tightened up for security
+// TODO: Should be tightened up for security
```
tomlinton: When deployed this is all handled by nginx, so this should only be applied in development, i.e. `NODE_ENV === 'development'`

infra/notifications/src/emailSend.js
wanderingstan: Ah yes, just added support for those. However, long term I feel that logic needs to get moved into the dapp itself: the network number is baked into the listing ID, so the dapp should do the right logic of getting the user to the right subdomain to display it. 

Likewise, the only reason we need custom logic on the ipfs gateway is because federating is currently blocked on staging. Once that is fixed we can always have the emails just point to our main gateway.

infra/notifications/src/emailSend.js
```diff
@@ -84,17 +84,20 @@ async function emailSend(
...
+          999: 'localhost.',
+          2222: 'dev.'
+        }
+        const listingNetwork = listing.id.split('-')[0] // First section of id is the network num
         const vars = {
           dappUrl:
-            listing.id[0] === '4'
-              ? 'https://dapp.staging.originprotocol.com'
-              : 'https://dapp.originprotocol.com',
+            `https://dapp.${networkSubDomains[listingNetwork]}originprotocol.com`,
```
franckc: for network 999, I don't think https://dapp.localhost.originprotocol.com will resolve to local dapp ? I think we use http://localhost on local...

wanderingstan: Ah crap you are right! Bit the bullet and have everything laid out explicitly.

wanderingstan:
Just finished the minimal HTML templates to aure's spec, so merged those into this PR. Please re-review if you can.

wanderingstan:
Closing in favor of https://github.com/OriginProtocol/origin/compare/stan/notifications-templates?expand=1 which has real styled templates per aure.

PR #2104, opened 2019-04-23
Update FractionalListing price calculation and booking availability

shahthepro:
The following changes have been made to FractionalListing type
- The price charged is now based on per-night basis. If the user books a listing for Thursday-Friday, he will be charged for Thursday night. The minimum number of nights a user can book is 1.
- Previously If an user checks out on a day, other users cannot book that night. That has been changed.


micahalcorn:
Thanks for this @shahthepro 👍 Does this change overfit for the homesharing case @matthewliu? There are surely some fractional cases where they are daily and not nightly. I’m not sure that we ever settled on this, but it may not be a priority right now.

nick:
Hmm I thought that housing was our only fractional category but I guess we have a bunch. However I think housing is the only one that's actually used right now so probably worth keeping nightly by default?

PR #2105, opened 2019-04-23
Mobile notifications

### LINKED ISSUE #1879: Log Wallet Creation
When a wallet (or account) is generated in our mobile app, we should post the public address and metadata to an in-house server and store it. This will allow us to learn about how many wallets are generated, what devices were used, how high the stakes are, etc.

### LINKED ISSUE #2049: Add endpoints to notifications server for saving device tokens
The mobile app needs to save device tokens to the notifications server so that it knows the device token to use for pushing notifications for a specific Ethereum address.

tomlinton:
This is a draft of the mobile push integration with the new notifications server. Check it out @wanderingstan and we can sync up.

- Adds an endpoint on the notifications server for saving device tokens
- Basic push logic (tested and functions on Android)

Things I have to do:

- [x] Handle wallet deletes
- [x] Add more tests for token save endpoint to notifications server
- [x] Test on iOS
- [x] Figure out how to handle URLs in notifications (open in WebView rather than browser?)

Closes #2049  and closes #1879.


micahalcorn:
Deep linking to webview would be very helpful. 🙏

tomlinton:
This is ready to go, some other changes:

- Simplified integration with @origin/graphql
- Add notifications server to @origin/graphql config
- Created a separate component for handling push notifications
- Made linking to pages in the WebView work. Doesn't work in development builds on iOS for the case where the app is closed and clicking the push notification opens the app. I think this might be because of the bundler and hopefully it works in test builds.

micahalcorn:
@wanderingstan did you want to take a look at this? If not, I'll probably just stamp it. 🤷‍♂

tomlinton:
Sure, whatever suits. Can do anything after 1pm your time.

This is a real hassle to test locally, you need to have a working docker-compose stack and an iOS/Android build running on a physical device (simulators don't support push notifications) and figure out how to connect to a DApp running on your machine over wifi. And you also need proper development APNS/FCM credentials. 😞

I think if we just push it to dev and use TestFlight builds to pick up any bugs I didn't find in my own testing it will be the easiest.

PR #2109, opened 2019-04-23
[Dapp] Pin webpack-cli to version 3.3.0

franckc:
### Description:

New version of some webpack npms were recently released and that caused our DApp to fail to start locally with the error: "Error: Cannot find module 'webpack-cli/bin/config-yargs"

The issue was that while some of the webpack npms are pinned to a specific version in package.json, others were not.

This PR pins all the webpack packages to the current most recent stable version to avoid such issue in the future. I tested locally and was able to start the DApp with this config.

Alternatively, we could always use most the recent versions. Please comment if you think it is better - I don't have much history on why we decided to pin only some versions and not others :)

### Checklist:


franckc:
Chatted with Nick. He had pinned webpack-dev-server due to this bug: https://github.com/webpack/webpack-dev-server/issues/1777
This issue was recently fixed in https://github.com/webpack/webpack-dev-server/commit/58d168237508bf419714d155496b954b98681d04#diff-bd664d218d585ff7bd6ceced46bd877aL229 but unfortunately the fix did not make it in the most recent release.

For now, pinning to previous stable version and once issue above is fixed we can upgrade...

PR #2115, opened 2019-04-24
Automatically enable messaging in mobile

### LINKED ISSUE #2063: Automatically Enabled Messaging
In cases where we are capable of under-the-hood signings (such as the app formerly known as Origin Wallet), we should automatically enable messaging without explicitly asking the user to confirm either of the two steps.

tomlinton:
- Generates messaging keys and calls the messaging client in the WebView via injectedJavascript so that every account using the mobile app has messaging automatically enabled
- Misc Docker fixes

This works, but I'm putting it up as a draft as I want to test it a bit more. Closes #2063.


tomlinton:
This one is ready to go, I'll merge it then put out some test builds!

Issue #41, opened 2018-01-07
Add `application/json` header to gateway responses


ryana:
We want to do this for security reasons so that people cannot host malicious content on the gateway.

ryana:
Adding this header may not be enough: https://stackoverflow.com/questions/37863357/why-do-browsers-execute-script-with-content-type-json

Investigating other ideas.

wanderingstan:
Any update on this, @ryana , or should we call this closed?

ryana:
Yikes no. No update here. I will circle back this week.

tomlinton:
This should no longer be an issue due to `origin-ipfs-proxy`.

Issue #38, opened 2018-01-07
Show other jsonschema fields on listing-detail page


wanderingstan:
Right now we only show photos, title, description, and price. 

We should at least show the other fields if present.

wanderingstan:
Right now all the listing fields (both ipfs and from the contract) are stored flat in the component state, along with housekeeping fields for the component like `step`.

I think we should instead have a `state.ipfsData` and `state.contractData`, which contain the correspondent fields. Then we can iterate over extra fields in ipfsData to display them (if we know how)

wanderingstan:
> contract and IPFS data could be stored separately
This is perhaps an issue unto itself. Currently we return ipfs+contract data merged together in one json ([docs](http://docs.originprotocol.com/#getbyindex)) 

For the purposes of this issue, we should just display any fields that are not handled by the GUI.

Aure (our designer) will surely have more to say about it, but for now just dumping them onto the display is a good first step. We'll sometime need to do a bigger design overhaul of extended data is shown on `listings-detail` page -- the seller address and "view on ipfs" sections are also just sort of tacked on right now. 

![image](https://user-images.githubusercontent.com/673455/38937337-382545fa-42e0-11e8-9a79-88b0c57b4e95.png)

ambertch:
Linked to #115

ambertch:
This is blocked by https://github.com/OriginProtocol/origin-js/issues/139

PR #2116, opened 2019-04-24
EventCache polish

nick:
Hopefully prevents those pesky fromBlock > toBlock errors


packages/event-cache/src/EventCache.js
```diff
@@ -18,18 +18,28 @@ const limiter = new Bottleneck({ maxConcurrent: 25 })
...
       try {
         debug('Loading event cache from IPFS', instance.ipfsEventCache)
         const cachedEvents = flattenDeep(
           await Promise.all(
             instance.ipfsEventCache.map(hash => get(instance.ipfsServer, hash))
           )
         )
-        fromBlock = cachedEvents[cachedEvents.length - 1].blockNumber + 1
-        debug(`Last cached blockNumber: ${fromBlock}`)
+        const lastCached = cachedEvents[cachedEvents.length - 1].blockNumber + 1
```
franckc: Nit: to make it a bit easier to reason about, I would not add +1 here and then remove -1 on line #38 and add +1 on line 41

packages/event-cache/src/EventCache.js
```diff
@@ -18,18 +18,28 @@ const limiter = new Bottleneck({ maxConcurrent: 25 })
...
         debug(`Loaded ${cachedEvents.length} events from IPFS cache`)
-        await instance.backend.addEvents(cachedEvents)
+        debug(`Last cached blockNumber: ${fromBlock}`)
```
franckc: Should this be lastCached rather than fromBlock ?

packages/event-cache/src/EventCache.js
```diff
@@ -18,18 +18,28 @@ const limiter = new Bottleneck({ maxConcurrent: 25 })
...
+        debug(`Latest indexed block: ${instance.latestIndexedBlock}`)
+        if (Number(lastCached) - 1 > Number(instance.latestIndexedBlock)) {
```
franckc: nit: would it make sense to cast lastCached value to Number when it gets assigned on line 34 and not do any casting here. Just being paranoid since on line #41 fromBlock gets assigned lastCached and it may cause unforeseen issues down the line if it is a string ?

packages/event-cache/src/EventCache.js
```diff
@@ -18,18 +18,28 @@ const limiter = new Bottleneck({ maxConcurrent: 25 })
 
 const getPastEvents = memoize(
   async function(instance, fromBlock, toBlock, batchSize = 10000) {
-    if (!instance.loadedCache && instance.ipfsEventCache) {
+    if (
+      instance.ipfsEventCache &&
+      !instance.loadedCache &&
+      (!instance.cacheMaxBlock ||
```
franckc: Under which scenario would instance.cacheMaxBlock not be set and we would still want to load data from the IPFS cache ? If instance.cacheMaxBlock is not set, doesn't it also mean the config would also not have IPFS hashes to load ?

PR #2118, opened 2019-04-24
Stan/notifications fix 2

wanderingstan:
- Fix problem where event listener would return without returning a http status code, causing eventlistener to hang

- Turn off mobile push and browser push, which currently aren't used


infra/notifications/README.md
```diff
@@ -87,5 +97,13 @@ For more (outdated) information, see [the original pull request](https://github.
...
-    npm run start:listener:development -- --identity --growth --marketplace --network=rinkeby --continue-file=../continue --webhook=http://localhost:3456/events --verbose --concurrency=4
+````
+export DATABASE_URL=postgresql://origin@localhost/origin
 
+npm run start:listener:development -- --identity --growth --marketplace \
+  --network=rinkeby \
+  --continue-file=../continue \
+  --notifications-webhook=http://localhost:3456/events \
+  --verbose \
+  --concurrency=4
```
franckc: I would discourage the use of concurrency, except when running backfills.
I think weird things could happen when concurrency is on - it makes listener process events without any ordering guarantee. This hasn't been tested well.

infra/notifications/src/app.js
```diff
@@ -178,6 +182,9 @@ app.post('/events', async (req, res) => {
     event.blockNumber
   } logIndex=${event.logIndex}`
 
+  // Return 200 to the event-listener without waiting for processing of the event.
```
franckc: That means the listener won't retry in case of a failure. It's probably ok for now since anyway email/mobile push  delivery is not guaranteed.

I think in longer term we should move to a queue model where the listener enqueues tasks that get consumed by the downstream services as opposed to using webhook. This gives more flexibility to the services to decide what to do in case of failure (let item in the queue for re-processing or not).

wanderingstan: Yep, that makes sense. I know there are a lot of good queuing libraries for this.

infra/notifications/README.md
wanderingstan: Sounds good, no need to complicate.  Removed.

PR #2124, opened 2019-04-26
Add ENABLE_GOOGLE_ATTESTATION to webpack.config.js

shahthepro:


dapps/marketplace/webpack.config.js
```diff
@@ -124,7 +124,8 @@ const config = {
       GIT_BRANCH: gitBranch,
       HOST: 'localhost',
       LINKER_HOST: 'localhost',
-      ORIGIN_LINKING: null
+      ORIGIN_LINKING: null,
+      ENABLE_GOOGLE_ATTESTATION: process.env.ENABLE_GOOGLE_ATTESTATION
```
franckc: This is a bit convoluted, but the way this works is that the key is the env var name and the value is the default to use in case that env var is not defined.
See the doc at https://webpack.js.org/plugins/environment-plugin/#usage-with-default-values

So I would just have:
`ENABLE_GOOGLE_ATTESTATION: false
`

shahthepro: Thanks @franckc. Have updated it.

PR #2133, opened 2019-04-26
Add services container dependacy

sparrowDom:
### Description:
I needed to add `origin-js` folder mapping in order for services to run. I couldn't find it as a dependancy in `services/package.json` though

Screenshot 2019-04-26 10 45 45

### Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,7 +46,6 @@ services:
       - ./packages/contracts/migrations:/app/packages/contracts/migrations
       # Core packages
       - ./packages/services/:/app/packages/services/
-      - ./packages/origin-js/:/app/packages/origin-js/
       - ./packages/graphql/:/app/packages/graphql/
       - ./packages/eventsource/:/app/packages/eventsource/
       - ./packages/event-cache/:/app/packages/event-cache/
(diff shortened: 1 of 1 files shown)


tomlinton:
I moved these fixtures into the services dir in #2102. A container rebuild should fix the issue in the screenshot. It is also an option for us to add that path to the volumes mapping.

sparrowDom:
I think that as long as we have lerna pull linker module and linker having `origin/js` dependancy the build is going to fail: 
https://github.com/OriginProtocol/origin/blob/master/infra/linking/package.json#L31

@tomlinton

tomlinton:
Do you get an error? `lerna bootstrap` only tries to bootstrap packages that exist, if the directories are missing it'll just ignore those. Nothing in the docker container should have a dependency on `@origin/js` or `@origin/linking` anymore.

sparrowDom:
Yeah I can not run `npm run bootstrap` inside services container because it fails with that error. I thought that happens because the linker is [mapped in docker-compose](https://github.com/OriginProtocol/origin/blob/master/docker-compose.yml#L65) and that lerna builds all the modules specifie in `lerna.json`

tomlinton:
With the same error in the screenshot? Did you do a container rebuild? That path is no longer referenced in master.

tomlinton:
Ahh I see, yea that line you linked needs to be removed. That is why the initial build would work but running it again in the container would cause it to fail.

sparrowDom:
Ah yeah that makes sense

tomlinton:
Thanks for finding that 👍

Issue #1795, opened 2019-03-25
Decouple profile picture from identity json blob


franckc:
Currently the user's profile picture gets saved as binary data within the JSON of the identity.
It would be better to save the profile picture as a separate blob in IPFS and have the identity JSON referencing the URL of the profile. In particular there are a few use cases that are emerging where having a separate profile URL would be useful:
 - including the profile picture in emails
 - generating custom images including the profile picture
Ideally we would have the back-end storing the profile picture URL in the identity DB table.

Fixing this involves the following:
 - On the DApp profile page: when the user publishes their identity, have the DApp first save the profile picture in IPFS and get the URL for it, then use that URL in the identity JSON.
 - In the event listener, extract the picture URL from the identity, save it in the identity table. 
 - Update the IPFS pinner to pin both the picture and the identity

For legacy accounts, we could write a script to backfill the data. It would:
 - scan all identities from the blockchain
 - read the identity JSON and extract the profile binary data
 - save the profile picture in IPFS
 - update the identity table with the profile URL

franckc:
CC @DanielVF since we were chatting about this today...

DanielVF:
In theory, given a string of bytes we can theoretically convert to the matching IFPS hash. This means that we can transparently transform old identities and return a single format from graphQL as long as the images are already stored separately.

Given this, it seems like the logical place to start is the backfill script.

DanielVF:
Here's some (possibly over generic) code to transform a buffer into its corresponding IPFS multihash.

https://github.com/alanshaw/ipfs-only-hash/blob/master/index.js

DanielVF:
From we will need to:

- [ ] Update the pinning service. (I figured we'd wait until after we've seen that GraphQL and the listener properly are passing on the avatarURL
- [ ] Modify the DAPP to display avatars using the new style avatar URL
- [ ] Remove storing the avatars the old way, and update the admin dapp

franckc:
Great progress @DanielVF ! In addition to the steps above, I think it would be handy to store in the identity table in the DB the IPF hash of the avatar. It would allow our back-end services (immediate use case being the notification service that sends email) to very quickly access this data as opposed to having to make a call to the blockchain to fetch it. Perhaps we could add a column "profile_hash" to the identity table ?

DanielVF:
Once the event listener re-runs, the avatarUrl for every identity should be in the identity JSON, just like the firstName and lastName. I think we should be good without adding another field.

[...]

PR #2135, opened 2019-04-26
Read IPFS stored profile avatars.

DanielVF:
Per issue #1795, we are working on moving away from storing the profile image data in the identity JSON. This PR takes care of the GraphQL side of this transition.

You can now fetch the avatarUrl from an identity GraphQL, even if that profile used an old style non-Url based avatar. In this case the avatar data in the profile is hashed by IPFS's splitting and hashing algorithms to create the actual IPFS url for the data.

In order that the avatar image will actually be at their IPFS URL, we also have a backfill script that works through the event history and for every non-deleted profile, uploads the avatar data to IPFS and checks that the expected IPFS hash and the actual uploaded hash match. It's single threaded for maximum obviousness, robustness, and debug-ability. (You could probably speed it up 30x by parallelizing the processing)

The listener should be storing the new `avatarUrl` field everywhere it stores, or webhooks, identities. In theory then, just rerunning the listener should update the database and ElasticSearch.

I've run it against Rinkeby, Mainnet, and Origin networks. In theory we should not need to run it again.

To run the backfill, go to the GraphQL package and `NETWORK=rinkeby npm run backfill_avatar`


DanielVF:
Backfill stats:

Mainnet
```
{ 'SUCCESS: Avatar uploaded to IPFS': 4604,
  'SKIP: No avatar': 1269,
  'SKIP: Already using new avatarUrl': 349,
  'SKIP: Not actual data in data url for avatar': 4 }
```

Rinkeby
```
{ 'SUCCESS: Avatar uploaded to IPFS': 79,
  'SKIP: Not data url for avatar': 2,
  'SKIP: No avatar': 216,
  'SKIP: Already using new avatarUrl': 6 }
```

Origin:
```
{ 
'SKIP: No avatar': 8,
'SKIP: Already using new avatarUrl': 2
}
```

packages/graphql/src/resolvers/IdentityEvents.js
```diff
@@ -105,6 +112,18 @@ export function identity({ id, ipfsHash }) {
       .filter(n => n)
       .join(' ')
 
+    // Make old style embedded avatars access by their IPFS hash.
+    if (identity.avatarUrl == undefined && identity.avatar != 'undefined') {
```
franckc: nit: use === and !=== operator to be consistent with the rest of our code base.

Same comment for the backfill script code.

DanielVF:
Changed to use '===' and '!===' !

franckc:
Thanks for fixing those nits !

PR #2144, opened 2019-04-27
Remove deprecated packages and cleanup

### LINKED ISSUE #1749: Delete old identity V00 code
Now that we retired DApp1, we should delete the old identity code under:
 - origin/packages/contracts/contracts/identity -> everything can be deleted except IdentityEvents.sol
 - origin/packages/origin-js/src/contractInterface/users/V00_adapter.js

### LINKED ISSUE #1752: Deprecate origin-js
Here are the tasks that need to be completed before we can finally get rid of origin-js:
 - [ ] #1749 - Delete old Identity code
 - [x] #1746 - Remove @origin/mobile dependency on @origin/js
 - [x] #1750 - Remove discovery server dependency on origin-js
 - [x] #1751 - Remove listener dependency on origin-js

tomlinton:
- Remove @origin/js
- Remove @origin/linking
- Remove @origin/linker-client
- Remove several workarounds required by other packages for building @origin/js
- Updated all Dockerfiles accordingly
- Committed basic built versions of contracts so that the build process doesn't require that step anymore (still .gitignored but I force added them to the repo)
- Removed all old identity contracts and migrations referencing those contracts
- Remove no gas contract

Standard `npm install` and docker builds still work, but it is hard to catch everything that could arise from removing this much code. :grin: If we remove it all anything that breaks should be a quick fix.

Closes #1752 and closes #1749.


tomlinton:
Thanks for checking it out @franckc!

You are right, I wasn't really sure how to handle that. What I've done now is restored those migrations but just replaced all the code with a placeholder function and a note that the migration is deprecated. I think it'll make it clearer in the future. Does that sound like a good approach?

franckc:
@tomlinton yes, empty migration scripts sounds perfect. thanks again for this dead code cleanup !

PR #2148, opened 2019-04-27
Save avatarUrl in identity database

DanielVF:
For #1795, we need to save the avatarUrl in the database. 

This PR makes migration to add the avatarUrl field, and the listener handler change to populate it.

This PR also allows you to specify an at-or-before blockNumber in the identity id when fetching from GraphQL, just like the marketplace listings do. So looking for "0xF17XXX732" gets you the latest version of an identity, while "0xF17XXX732-5005000" gets you the identity as it looked at the end of block 5005000. 

### Checklist:


infra/discovery/src/listener/handler_identity.js
```diff
@@ -207,7 +207,8 @@ class IdentityEventHandler {
       facebookVerified: decoratedIdentity.facebookVerified || false,
       googleVerified: decoratedIdentity.googleVerified || false,
       data: { blockInfo },
-      country: decoratedIdentity.country
+      country: decoratedIdentity.country,
+      avatarUrl: decoratedIdentity.avatarUrl
```
franckc: Is it worth just saving the IPFS hash vs the whole URL ? To save on DB space and possibly make it easier in future to change our IPFS gateway naming or what IPFS gateway to use to fetch those avatar pics ?

franckc: oh. Nevermind ! I just realized this is an IPFS URL (e.g. ipfs://<hash>) and not an URL pointing to our IPFS gateway.

franckc:
Going to merge this soon, after we deploy some changes from @mikeshultz to stabilize the listener....

franckc:
Merged this in master. Will go live next time we do a prod deploy.
CC @mikeshultz as a FYI

mikeshultz:
Out of curiosity, @DanielVF was there a specific reason [you added this](https://github.com/OriginProtocol/origin/pull/2148/files#diff-d5a210f834ba1e38ab5f65127f9dcaefR75)?  Did you see that occurring when you were testing?  

Looks like you branched after EventCache dropped with sorting, so I wouldn't expect this to be an issue.

DanielVF:
@mikeshultz This, and the other tiny change above in the same file,  allows us to get information about an identity at a specific block. In this case, for the work I was doing on avatar backfill, I wanted to be able to easily look using GraphQL at previous versions of an identity.

It's not discarding events unless you pass in a blocknumber to discard events after.

mikeshultz:
@DanielVF ah, okay.  I didn't take in the wider context and assumed this was a defensive check on order(kind of related to another issue I'm looking at). Thanks!

Issue #1727, opened 2019-03-17
Listing Image Gallery


micahalcorn:
The following changes should be made to the image viewing experience on the listing detail page:

- [ ] Thumbnails under the main image (shown [here](https://zpl.io/bJ7WX1r))
- [ ] Overflowing thumbnails with horizontal scroll

ldu2:
Hi, I just submitted a pull request, regarding this issue. You can check it out [here](https://github.com/OriginProtocol/origin/pull/2151). Let me know if anything.

micahalcorn:
Resolved by #2151 ✅

PR #2151, opened 2019-04-29
Change thumbnail listing to horizontal and overflow scroll bar

ldu2:
- Regarding issue #1727, changing the listing style of the thumbnails
- I simply modified some css to change the style
        -Thumbnails under the main image
        -Overflowing thumbnails with horizontal scroll
But since I am not invited in the Zeplin project, not sure how these two looks like:
       -Navigation arrows on the main image (shown here)
       -Closeable lightbox expansion of main image on click (shown here)
So I didn't attempt that.
This is how it looks like now:
Screenshot_2019-04-29 Zinc House - Origin
Let me know if anything.


micahalcorn:
Thanks @ldu2 🙏 Can you remove the horizontal margin on the thumbnails container? Since it's no longer being displayed vertically, the margin causes it to be unnecessarily narrow. Ideally, we would also reduce remove the left padding on the first image and the right padding on the last image so that their edges align with the main image.

ldu2:
I have removed the margin on both sides and padding on first and last images on left and right side, correspondingly. Sorry for the multiple commit. I was getting used to the formatting tools. Anyway, this is how it looks like now:
Screenshot from 2019-05-01 16-33-04

micahalcorn:
@ldu2 that looks good, but I think you should be able to accomplish the with just CSS. Want to take another stab at that? 🙂

ldu2:
Yeap. You are right. The CSS version is up there now. Let me know if anything. Appreciate the feedback.

micahalcorn:
Great, thanks 👍 I've also added you to the GitHub organization so that you can push a branch directly to our monorepo in the future. That will allow for CI tests to run and avoid the need for a fork.

Issue #220, opened 2018-05-25
Reference the same origin js source in develop and master


tyleryasaka:
We have a weird flow right now where in the `develop` branch, `origin` points to github. But in `master`, `origin` points to npm.

This is bad practice and requires us to manually revert this change every time we merge `develop` into `master`. If someone forgets or doesn't realize this is supposed to happen, things mysteriously don't work. @joshfraser ran into this yesterday.

I am proposing that we simply keep both branches pointing to `npm`. When a new version of `origin` is published to npm, we simply update the `package.json` to point to the new npm version; we do this in the `develop` branch. Then we test things, make sure everything works, and once we're happy we merge `develop` into master to create a new release.

As far as concerns about the `origin` in npm getting out of sync with the `develop` branch of `demo-dapp`: this is a non-issue because we are instructing people to run `npm run install:dev`, which does an `npm link` and bypasses whatever version is specified in `package.json`.

In other words:
- on `master`, people will use the `origin` specified in `package.json` - this should be the latest npm `origin`.
- on `develop`, it won't matter what's in `package.json`. People will be expected to use `npm run install:dev`/`npm link` and use the local `origin`. Thus it will also be safe to specify the latest published npm `origin` on the `develop` branch.

Additionally, our current setup actually doesn't add any value. The github `origin` will **never** be usable because it does not contain a built package with deployed contracts to test nets, and does not start ganache/truffle locally. There is absolutely no value in having a github reference in any branch.

@wanderingstan How does this sound to you?

joshfraser:
Thanks for walking through this @tyleryasaka & I'm in full support of this change.

wanderingstan:
Agreed! Fixed in: https://github.com/OriginProtocol/demo-dapp/commit/77f3434cdbefcb165754a8e93861bd26bbfa0ba2

jordajm:
BTW - @wanderingstan I think this is a much better solution to the issue we ran into last week re: multiple versions of origin-js in `package.json` that prompted me to recommend temporarily disabling `package-lock.json`. Much cleaner and long-term solution.
CC @tyleryasaka 👍

PR #2154, opened 2019-04-29
add the ability to use the latest block number from an event

mikeshultz:
### Description:

In case the fromBlock/toBlock logic wasn't convoluted enough, I added this!  If the EventCache property `useEventAsLatestBlock` is set truthy, it will use the block number from the latest retrieved event as `fromBlock` for future requests.  This should prevent any event loss from accidental future requests and as-is, this change will only effect the event listener.

### Checklist:


infra/discovery/src/listener/listener.js
```diff
@@ -206,8 +213,8 @@ async function main() {
      * event-cache
      */
     processedToBlock = Math.min(
-      contractsContext.marketplace.eventCache.lastQueriedBlock,
-      contractsContext.identityEvents.eventCache.lastQueriedBlock
+      contractsContext.marketplace.eventCache.lastEventBlock,
```
franckc: Marketplace and Identity events happen at difference pace.
So for example it's possible we would see a very few identity events but lots of marketplace events.
Which would then cause marketplace events to get reprocessed many times because processedToBlock would only get advances when we see an identity event...
The listener only provides at least once guarantees but still I wonder if there is something we could do to avoid this reprocessing ?

infra/discovery/src/listener/listener.js
```diff
@@ -125,6 +125,13 @@ async function main() {
   const tickIntervalSeconds = 5
   let start
 
+  /**
+   * Make sure we're using the block number from the latest retrieved event.  This is set to false
+   * in the graphql contracts initialization that's being used here.
+   */
+  contractsContext.marketplace.eventCache.useEventAsLatestBlock = true
```
franckc: Any reason we should not make this the default ?
The DApp could probably benefit from not skipping events as well ?

infra/discovery/src/listener/listener.js
mikeshultz: I went this direction for safety.  I can surely change this to max, if you think that's the better call.  The only other option I suppose would be refactoring the listener entirely to separate the two.

infra/discovery/src/listener/listener.js
mikeshultz: If there are no objections, I'll make this change.  It shouldn't cause any more requests, just open up the request window a bit(potential load for infura/JSON-RPC).

mikeshultz: I've gone ahead and made this change.  It simplified things a little.

infra/discovery/src/listener/listener.js
franckc: We could indeed refactor the listener to have 1 loop per contract type and keep separate block cursors for each loop... It's a slightly bigger change but wouldn't be super hard I think...
I'd say either we do that or use Math.max. Your call. What do you think ?

mikeshultz: Refactored.  Uses individual cursors now.

mikeshultz:
Listener refactor done.  If you guys are good with the changes, this PR should be good to go.

Also minor updates to docs and added a missing dependency(prettier).

infra/discovery/src/listener/listener.js
```diff
@@ -134,84 +136,92 @@ async function main() {
...
     start = new Date()
 
-    // Compute the range of blocks to process,
-    // while respecting trailing block configuration.
+    // Compute the range of blocks to process
     const currentBlock = await withRetrys(async () => {
       return context.web3.eth.getBlockNumber()
     })
-    const toBlock = Math.max(currentBlock - context.config.trailBlocks, 0)
+    const toBlock = Math.max(currentBlock, 0)
```
franckc: Is this needed ? would context.web3.eth.getBlockNumber() ever return a negative block number ?

infra/discovery/src/listener/listener.js
```diff
@@ -134,84 +136,92 @@ async function main() {
...
-    }
-    logger.info(
-      `Querying events within interval (${processedToBlock}, ${toBlock}]`
-    )
+    for (const contractKey of Object.keys(contracts)) {
+      let processedToBlock = await getLastBlock(
+        context.config,
+        `${contractKey}_`
+      )
+      logger.debug(`Processing from ${processedToBlock} for ${contractKey}`)
```
franckc: nit: maybe add in the debug the prefix so we know for which contract those queries are for

infra/discovery/src/listener/listener.js
```diff
@@ -134,84 +136,92 @@ async function main() {
...
-    // Listener is up to date. Nothing to do.
-    if (toBlock <= processedToBlock) {
-      logger.debug('No new blocks to process')
-      continue
-    }
-    logger.info(
-      `Querying events within interval (${processedToBlock}, ${toBlock}]`
-    )
+    for (const contractKey of Object.keys(contracts)) {
+      let processedToBlock = await getLastBlock(
```
franckc: nit: maybe add a debug statement to show which contract is being processed. like
logger.debug(`Processing ${contractKey}`)

infra/discovery/src/listener/listener.js
```diff
@@ -134,84 +136,92 @@ async function main() {
...
+      )
 
-    // Retrieve all events for the relevant contracts
-    const eventArrays = await Promise.all([
-      withRetrys(async () => {
-        return contractsContext.marketplace.eventCache.allEvents()
-      }),
-      withRetrys(async () => {
-        return contractsContext.identityEvents.eventCache.allEvents()
+      // Retrieve all events for the relevant contracts
```
franckc: super nit: contract should be singular

infra/discovery/src/listener/listener.js
```diff
@@ -134,84 +136,92 @@ async function main() {
...
-    // Flatten array of arrays filtering out anything undefined
-    const events = [].concat(...eventArrays.filter(x => x))
+      // Flatten array of arrays filtering out anything undefined and filter
+      // out events outside of interval (processedToBlock, toBlock].
+      const newEvents = [].concat(
+        ...eventArrays.filter(event => {
+          if (
+            event &&
+            event.blockNumber > processedToBlock &&
+            event.blockNumber <= toBlock
```
franckc: do we still need this filtering on toBlock ? wouldn't it be ok to process all events > processedToBlock ?

infra/discovery/src/listener/listener.js
```diff
@@ -134,84 +136,92 @@ async function main() {
...
+      } else {
+        // Have the handler process each event.
+        for (const event of newEvents) {
+          // Note: we purposely do not set the exitOnError option of withRetrys to false.
+          // In case all retries fails, it indicates something is wrong at the system
+          // level and a process restart may fix it.
+          await withRetrys(async () => {
+            try {
+              return handleEvent(event, context)
+            } catch (err) {
```
franckc: Catching error here will prevent retries. Let's let withRetrys logging the error (it should already do that I think) ?

infra/discovery/src/listener/listener.js
```diff
@@ -134,84 +136,92 @@ async function main() {
...
-     */
-    processedToBlock = Math.min(
-      contractsContext.marketplace.eventCache.lastQueriedBlock,
-      contractsContext.identityEvents.eventCache.lastQueriedBlock
-    )
-    await setLastBlock(context.config, processedToBlock)
-    if (context.config.enableMetrics) {
-      blockGauge.set(processedToBlock)
+      // Record state of processing
+      logger.debug(`Updating last processed block to ${toBlock}`)
```
franckc: Move this debug statement to line #220 and replace toBlock with processedToBlock

infra/discovery/src/listener/listener.js
```diff
@@ -134,84 +136,92 @@ async function main() {
...
+       * Don't assume it's the latest block known above, but use the one from
+       * event-cache
+       */
+      const latestIndexedBlock =
+        contracts[contractKey].eventCache.latestIndexedBlock
+      if (processedToBlock < latestIndexedBlock) {
+        processedToBlock = latestIndexedBlock
+        await setLastBlock(context.config, processedToBlock, `${contractKey}_`)
+        if (context.config.enableMetrics) {
+          blockGauge.set(processedToBlock)
```
franckc: Should we have a metric per contract as opposed to a single global metric. Perhaps add a TODO

infra/discovery/src/listener/listener.js
```diff
@@ -235,4 +245,12 @@ logger.info(
 )
 
 setNetwork(config.network)
-main()
+try {
+  main().catch(err => {
```
franckc: are both try/catch statement and .catch handler necessary ? Isn't it redundant ?

infra/discovery/src/listener/listener.js
```diff
@@ -91,6 +90,7 @@ const config = {
   // File to use for picking which block number to restart from
   continueFile: args['--continue-file'] || process.env.CONTINUE_FILE,
   // Trail X number of blocks behind
+  // DEPRECIATED
```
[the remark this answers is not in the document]
mikeshultz: Is it? One of Google's definition of deprecate: "another term for depreciate."  Odd, I didn't even know "deprecate" was a word until today...

infra/discovery/src/listener/listener.js
mikeshultz: It's there as `contractKey`, no?

infra/discovery/src/listener/listener.js
mikeshultz: Same as above

infra/discovery/src/listener/listener.js
mikeshultz: Honestly, I'm not really sure why this check was there to begin with?  I tend to try not to fiddle with defensive checks added by people before me, just in case they knew something I didn't.

infra/discovery/src/listener/listener.js
mikeshultz: Will add TODO, since at the moment, I know nothing about prom.  Let me know if I should dig into it.

infra/discovery/src/listener/listener.js
mikeshultz: I have no idea what happened here... thanks

infra/discovery/src/listener/listener.js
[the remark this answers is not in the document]
franckc: You would obviously know better. You are the native speaker :)

infra/discovery/src/listener/listener.js
franckc: yep. u r right...

infra/discovery/src/listener/listener.js
franckc: Do you mind removing it to clean up as much as possible this logic ?

Issue #2166, opened 2019-05-01
Error on profile page


franckc:
I get an error page when going to the profile page.
This is in production but I have seen similar issue on staging and dev.

Repro steps:
1. got to dapp
2. clear local storage
3. signup for origin reward
4. from campaign page, click on one of the attestation action
5. link goes to https://dapp.originprotocol.com/#/profile where I get blank page with error "Error! Please refresh the page". Clicking on Error shows a modal that says "TypeError: Cannot read property 'length' of undefined". No error show up in JS console.

franckc:
Mitigated by this bandaid that I committed and deployed to prod: https://github.com/OriginProtocol/origin/commit/d88b142c1f71264e2fa2a56ac18eabfa2e20ab51

Though I'd love for someone who understands react to tell me what the root cause is so that we can put in a better long-term fix... :)

tomlinton:
The issue is mostly [here](https://github.com/OriginProtocol/origin/blob/master/dapps/marketplace/src/hoc/withGrowthCampaign.js#L46-L58). That renders the wrapped component even when loading has not yet completed, so initially the `growthCampaigns` prop that gets passed will be undefined. It could be handled like [this](https://github.com/OriginProtocol/origin/blob/a8542d2ea7c9c0b6dbcc3d77d23d319701a79e6c/dapps/marketplace/src/components/GrowthCampaignBox.js#L139-L140) or how you've done it.

Thanks for figuring out the fix!

franckc:
ahhhh. you are so skilled in the React arts @tomlinton - thank you for enlightening me !

micahalcorn:
Resolved by #2170 ✅

PR #2170, opened 2019-05-01
Fix for growth box causing crash on profile page

franckc:
First pull request? Read our [guide to contributing](https://docs.originprotocol.com/guides/getting_started/contributing.html)

### Description:

Fix for #2166

### Checklist:


dapps/marketplace/src/hoc/withGrowthCampaign.js
```diff
@@ -43,8 +43,10 @@ function withGrowthCampaign(WrappedComponent) {
                     // action is completed
                     fetchPolicy="network-only"
                   >
-                    {({ data, error }) => {
-                      if (error) {
+                    {({ data, error, networkStatus, loading }) => {
+                      if (networkStatus === 1 || loading) {
+                        return ''
+                      } else if (error) {
```
nick: Ideally for these higher order components we should try to avoid returning a different component depending on the network or loading state, unless there's a hard error.

Instead a few lines below I think we can do this:

`growthCampaigns={get(data, 'campaigns.nodes') || []}`

franckc: Ah yes, probably better :)  Implemented your idea !

PR #2182, opened 2019-05-01
Use avatarUrl for all reads from DApp

DanielVF:
We are moving from using inline base64 blobs for our avatars to storing them in their own IPFS hashes (see #1795).

This PR switches the DApp **completely away from reading the old blob form avatars**. Instead the DApp now uses only the new avatarUrl and avatarUrlExtended attributes coming from GraphQL.

We still write both styles.

Since this does touch a lot of the DAPP (at least in tiny ways), it would be great to have another set of eyes test it out, particularly on mainnet and rinkeby.

--

I've made the avatar component be able to take several different props for display an avatar. Since this component is used so often in the DApp, it makes sense to put a little logic here and make the other calling components much simpler.

I found that I was not yet uploading the new style avatars from the onboarding flow. This means we will need to run a avatar backfill again after this PR goes live. (I've run another backfill again today, so the number of possible affected accounts should be tiny.)


dapps/marketplace/src/utils/makeGatewayUrl.js
```diff
@@ -0,0 +1,17 @@
+/**
```
franckc: As opposed to duplicating code, how about importing the @origin/ipfs package in the dapp and using that function from there ?

franckc:
@nick Does that look good to you ?

PR #2197, opened 2019-05-02
Notications for messaging, dupe checking, and template fixes

wanderingstan:
This PR brings (1) user-to-user messages now sent via email and mobile push and (2) simple duplicate (dupe) checking to make sure we don't send the same message within a given time frame. 

EDIT: and (3) fixes styling of transaction emails so images appear inline-vertical instead of across two columns, which looked terrible on mobile.

@tomlinton : I can't check the mobile push stuff, so please confirm if message notifications actually work and the dupe checker too. 

**Env Vars**

`DUPE_LOOKBACK_MS` sets how many milliseconds to look back in time for duplicates. Defaults to 30 minutes. 

`NOTIFICATIONS_ENDPOINT_URL` for messaging server needs to be pointed at the notications server, with path `/messages`


infra/notifications/migrations/20190502223858-create-notification-log.js
```diff
@@ -0,0 +1,38 @@
...
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface
+      .createTable(TableName, {
+        id: {
+          type: Sequelize.INTEGER,
+          allowNull: false,
+          autoIncrement: true
+        },
+        send_date: {
```
franckc: Do we need this send_at column or could we use created_at which typically stores the time the row was inserted in the DB ?

infra/notifications/src/app.js
```diff
@@ -50,6 +50,11 @@ process.argv.forEach(arg => {
 })
 
 const config = {
+  // URL for messaging on dapp
+  dappMessagesUrl:
+    args['--dapp-messages-url'] ||
+    process.env.DAPP_MESSAGES_URL ||
+    'https://dapp.originprotocol.com/#/messages',
```
franckc: Do we want this default value ? Is there a risk it could cause messages to be sent to production when developers test on their local environment ?

infra/notifications/src/app.js
```diff
@@ -219,6 +231,23 @@ app.delete('/mobile/register', async (req, res) => {
...
 
+/**
+ * Endpoint called by the messaging-server
+ * list of eth address that have received a message
+ */
+app.post('/messages', async (req, res) => {
+  res.status(200).send({ status: 'ok' })
+
+  const sender = req.body.sender // eth address
+  const receivers = req.body.recievers // array of eth addresses
```
franckc: nit: what's difference between receivers and recievers ? should we use same naming ?

infra/notifications/src/emailSend.js
```diff
@@ -19,10 +18,140 @@ if (!process.env.SENDGRID_FROM_EMAIL) {
...
+  getMessageFingerprint,
+  isNotificationDupe,
+  logNotificationSent
+} = require('./dupeTools')
+
+//
+// Email notifications for Messages
+//
+async function messageEmailSend(receivers, sender, config) {
+  if (!receivers) throw 'receivers not defined'
```
franckc: nit: I believe in the back-end codebase we've been trying to throw new Error() rather than just string

infra/notifications/src/dupeTools.js
```diff
@@ -0,0 +1,33 @@
...
+      messageFingerprint: messageFingerprint,
+      sendDate: {
+        [Op.gte]: new Date(Date.now() - config.dupeLookbackMs)
+      }
+    }
+  })
+}
+
+async function logNotificationSent(messageFingerprint, ethAddress, channel) {
+  return NotificationLog.create({
```
franckc: nit: can be simplified to
```
return NotificationLog.create({
  messageFingerprint,
  ethAddress,
  channel
})
```

infra/notifications/src/dupeTools.js
```diff
@@ -0,0 +1,33 @@
...
+    messageFingerprint: messageFingerprint,
+    ethAddress: ethAddress,
```
franckc: Are all those eth addresses lower case when writing/reading from DB ?
Same comment for all DB reads/writes in this PR

infra/notifications/src/emailSend.js
```diff
@@ -19,10 +18,140 @@ if (!process.env.SENDGRID_FROM_EMAIL) {
...
+  const emailTemplateHtml = _.template(
+    fs.readFileSync(`${templateDir}/emailTemplate.html`).toString()
+  )
+  // Standard template for text emails
+  const emailTemplateTxt = _.template(
+    fs.readFileSync(`${templateDir}/emailTemplate.txt`).toString()
+  )
+
+  const messageSender = Identity.findOne({
+    raw: true,
```
franckc: What's the reason for setting raw to true ? 
Afaik we don't use that in the rest of the codebase...

If it is really necessary, maybe add a comment explaining why.
It was not obvious to me reading the code :)

infra/notifications/src/emailSend.js
```diff
@@ -19,10 +18,140 @@ if (!process.env.SENDGRID_FROM_EMAIL) {
...
+              }),
+              html: emailTemplateHtml({
+                message: message.html(templateVars)
+              }),
+              asm: {
+                groupId: config.asmGroupId
+              }
+            }
+
+            if (config.verbose) {
```
franckc: As opposed to wrapping those debug statements with config.verbose, how about using logger.debug statement that can be enable via setting env var LOG_LEVEL to DEBUG ?

infra/notifications/src/app.js
tomlinton: Maybe not for this PR but we have DAPP_OFFER_URL and DAPP_MESSAGES_URL, all we really need is DAPP_URL because the paths can just be added as they don't change. And we already have all the DAPP_URLS for each network defined in emailSend.js.

We could just set a NETWORK_ID env var and then not have to parse that out of the listing in emailSend.js

@franck this is the URL that'll open when you click on the push notification, so the worst case is that this will cause someone to open messages on the wrong network on mobile.

infra/notifications/migrations/20190502223858-create-notification-log.js
wanderingstan: Ah yeah, that would work! (Designed the table structure before realizing the Sequelize always adds those date columns)

infra/notifications/src/app.js
wanderingstan: Love the idea of a `DAPP_URL` and/or `NETWORK_ID` var. It's a lot of gyrations to get it from the listing, and for messaging there is no clue what network we're on. 

(Frank, that URL was originally hard-coded..so at least it was moving in the right direction. :)

infra/notifications/src/app.js
wanderingstan: The difference is my inability to spell... 😬

infra/notifications/src/dupeTools.js
wanderingstan: Ah shoot, I of all people should have been checking that. Will do. 
Also realize I should convert them to the checksum'ed format when displaying, e.g. in an email.

infra/notifications/src/emailSend.js
wanderingstan: Good point. The whole `--verbose` thing started before @tomlinton showed me the light of our logger over console.log. Agreed that we should ditch verbose and just rely on `LOG_LEVEL`s.

infra/notifications/src/app.js
wanderingstan: Ha! Vidicated: it's actually because Yu Pan can't spell, and I blindly copied. That field comes from the messaging service, here:
https://github.com/OriginProtocol/origin/pull/2197/files#diff-8a57fbb88a1ad9dcf4f114b9e7efccb2L300
Will still fix it. :)

infra/notifications/src/emailSend.js
wanderingstan: This is me being new to Sequelize and trying to console.log() stuff for debugging. Removed in 977111f19e8d5e38162b1879b3ba345a4ae22859.

wanderingstan:
Okay, I think I got fixes in for everything but the DAPP_URL/ NETWORK_ID suggestion. (Which I think is a good one and will do in the morning.)

infra/notifications/src/emailSend.js
```diff
@@ -19,10 +19,133 @@ if (!process.env.SENDGRID_FROM_EMAIL) {
...
+    where: {
+      ethAddress: {
+        [Op.or]: receivers
+      }
+    }
+  })
+  Promise.all([messageSender, messageReceivers]).then(
+    ([senderIdentity, receiversIdentities]) => {
+      console.log(senderIdentity)
+      receiversIdentities.forEach(async s => {
```
tomlinton: This won't work for cases where the person has never published an identity as messageReceivers will be empty. The same thing applies to the mobile code. Not sure how we want to handle that. I know we want to encourage people to have an identity but I still think notifications from unnamed users should work.

infra/notifications/src/mobilePush.js
```diff
@@ -47,21 +59,95 @@ if (process.env.FIREBASE_SERVICE_JSON) {
...
+  if (!receivers) throw new Error('receivers not defined')
+  if (!sender) throw new Error('sender not defined')
+
+  // TODO: Move to config
+  const payload = {
+    url: config.dappMessagesUrl
+  }
+
+  // Get ID of sender
+  const messageSender = Identity.findOne({
```
tomlinton: These need an `await` in front of them because this should return a promise

tomlinton: Added those!

tomlinton:
Pushed up a few fixes I found running it in Docker @wanderingstan, hope you don't mind!

infra/notifications/src/mobilePush.js
wanderingstan: Thanks!

infra/notifications/src/emailSend.js
wanderingstan: Excellent point. Added code to follow the convention of using the eth address where a name is not available. 
Commit: 32d4a7e5a736cbfcab69cd94d60ee03e53538a93

infra/notifications/src/mobilePush.js
tomlinton: Sorry @wanderingstan I totally misread the code here, I see they don't need this because of the Promise.all, removed!

infra/notifications/src/emailSend.js
tomlinton: Users without any identity published can still be contacted - via messaging. All mobile users will have messaging enabled and be able to receive push notifications but have no identity published when the app first opens. I think it'll be reasonably common for someone to open the mobile app, see something they want to buy, and message the user without publishing an identity.

wanderingstan: Gotcha, true. We're okay for email (as in this particular case of `emailSend.js`) but for mobilepush with auto-enable messaging this will be an issue. I'll figure out the sequeleze for getting back a NULL in case of no identity.

wanderingstan: Ok, I pulled out the not-yet-used-anyway stuff around sender and receiver identities. This plain-jane approach should work for all mobile users: 2149e346aa60c477195b0c427fe5532c9c7ed478

infra/notifications/src/emailSend.js
wanderingstan: Yes, done!

infra/notifications/src/app.js
wanderingstan: Changed to just `ETH_NETWORK_ID`, and all other URLs are derived from that. _Much_ cleaner.

Fix is in: 96ff1fa62779ed4a0cc94e4032501f371dd25368

wanderingstan:
@tomlinton @franckc - I know you guys are swamped but please give this a looksie when you can; would love to get this merged and tested before end of sprint Wednesday! 

(..and Tom, thanks for the fixes..any time!)

PR #2199, opened 2019-05-03
Nick/tokenary

### LINKED ISSUE #2023: Fix our DApp to work with the Tokenary extension for Safari
[Tokenary](https://tokenary.io/) have been kind enough to feature us in their wallet. Unfortunately, our DApp doesn't work with their Safari extension.

The were able to track down the issue. It looks like Origin calls "ethereum.enable()" before their script has been injected (and they can't inject it earlier because of Safari restrictions). The following gist shows how it may be fixed: https://gist.github.com/grachyov/caee1ca2b59cbb8c3d8ff88526e7058c

nick:
Support for tokenary and local SSL

Closes #2023


nick:
Closes #2023

PR #2200, opened 2019-05-03
Data Concistency Checker

mikeshultz:
### Description:

This script verifies that proper DB records are being created by the listener and does cursory checking of the data.  Includes:

- `IdentityUpdated`
- `ListingCreated`
- `OfferCreated`

### Checklist:


mikeshultz:
I think this is reasonably complete and ready for testing against dev, staging, and/or prod.  Tested listing and offer checking by populating my local from mainnet, and tested identity/attestations by created attestations and triggering IdentityUpdated events.

I'd like to see the output from another system.

Here's an example command:

    LOG_LEVEL="INFO" DATABASE_URL="postgresql://mike@localhost:5432/origin" NETWORK="mainnet" node ./src/listener/tools/consistencyChecker.js --offers --listings --identity --from-block=6425000 --ipfs-gateway=https://ipfs.originprotocol.com

And [check the README for more detailed information](https://github.com/OriginProtocol/origin/tree/mikeshultz/consistency-check/infra/discovery/src/listener/tools).

Also, let me know if the location of this tool is okay.  I kind of packed it into `infra/discovery/src/listener/tools` with another script, even though this is a bit more modular.

infra/discovery/src/listener/tools/consistency/listing.js
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * Verify that listing DB records have been created for every ListingCreated
+ * event
+ *
+ * @param args {object} map of function arguments
+ */
+async function validateListings({ contractsContext, fromBlock }) {
+  const events = await listingCreatedEvents(
```
franckc: Would be valuable to also validate ListingUpdated events
Perhaps add a TODO. Let's do a first run of the tool as is and we can improve it afterwards.

infra/discovery/src/listener/tools/consistency/utils.js
```diff
@@ -0,0 +1,99 @@
+const Bottleneck = require('bottleneck')
+const flattenDeep = require('lodash/flattenDeep')
+const range = require('lodash/range')
+
+const db = {
+  //...require('@origin/bridge/src/models'),
```
franckc: cleanup

infra/discovery/src/listener/tools/consistency/utils.js
```diff
@@ -0,0 +1,99 @@
...
+  //...require('@origin/identity/src/models'),
+  ...require('../../../models')
+}
+
+const log = require('../../logger')
+
+const JSONRPC_REQUEST_BATCH_SIZE =
+  process.env.JSONRPC_REQUEST_BATCH_SIZE || 1000 // 10000 was too much?
```
franckc: Perhaps mention in comment web3 v34 chokes up id batch size > 1k

infra/discovery/src/listener/tools/consistency/utils.js
```diff
@@ -0,0 +1,99 @@
...
+ * Get latest block number for a specific event listener
+ *
+ * @param id {string} listener ID
+ * @param prefix {string} Prefix of the listener ID, if one
+ * @returns {number} block number the listener is at
+ */
+async function getListenerBlock(id, prefix) {
+  const listenerId = `${prefix}${id}`
+
+  const res = await db.Listener.findAll({
```
franckc: id is a pkey on listener table. so there can't be more then 1 row. so we should be able to just use findOne rather than findAll

infra/discovery/src/listener/tools/consistencyChecker.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * Sanity checker to verify all past events are in the database
+ *
+ * Blame: Mike Shultz
```
franckc: hehe

infra/discovery/src/listener/tools/consistencyChecker.js
```diff
@@ -0,0 +1,76 @@
...
+    await validateOffers({
+      contractsContext,
+      ...config
+    })
+  }
+  return commands
+}
+
+setNetwork(config.network)
+;(async function() {
```
franckc: is this needed ? calling main().then(...) directly does not work ?

infra/discovery/src/listener/tools/consistency/offer.js
```diff
@@ -0,0 +1,99 @@
...
+  return true
+}
+
+/**
+ * Verify that offer DB records have been created for every OfferCreated event
+ *
+ * @param args {object} map of function arguments
+ */
+async function validateOffers({ contractsContext, fromBlock }) {
+  const events = await offerCreatedEvents(
```
franckc: Would be good to test other events. Such as OfferAccepted, OfferFinalized, etc...
Primarily what I'd want to make sure is that in the db offer.status reflect the latest event for that offerId.
e.g. if latest event was OfferAccept, then offer.status should be "Accepted".
ok to add as a TODO that can be implemented once we run current script in prod.

infra/discovery/src/listener/tools/consistency/utils.js
mikeshultz: Good call, thanks.

infra/discovery/src/listener/tools/consistency/offer.js
mikeshultz: Adding TODO for now.  Would be happy to continue to improve this.

infra/discovery/src/listener/tools/consistency/ident.js
```diff
@@ -0,0 +1,237 @@
...
+    return false
+  }
+
+  // Make sure the record is sane
+  try {
+    validateIPFSToDB(ipfsJson, records[0])
+  } catch (err) {
+    // Handle Assertion errors
+    const errString = err.toString()
+    if (errString.indexOf('Assertion') > -1) {
```
franckc: nit: an alternative would be to define an AssertionError that would be thrown by assert and could be caught here.

mikeshultz: I was actually looking at this when I was writing it, but it seemed support was iffy.  There's [a lot of non-standard red boxes and considerations in the MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error).  You think that would normally be the way to go in this case?

infra/discovery/src/listener/tools/consistencyChecker.js
mikeshultz: hm, I don't remember exactly why I added this.  Just tested your suggestion and it works as expected.  I'll clean this up in my incoming commit.

infra/discovery/src/listener/tools/consistency/ident.js
franckc: I'm in no way a JS expert :) But my understanding was that custom exceptions are well supported and standard in Node.js. But not so much in browser/babel/etc...

mikeshultz: I'm going to go ahead and try/do this, since I've never done it in JS before.  Should have it up shortly.

mikeshultz: Scratch that.  I can't get this to properly work.  Comparing with `instanceof`, same as the MDN example, I just get this:

    [ERROR] event-listener: TypeError: Right-hand side of 'instanceof' is not an object
        at verifyIdent (/home/mike/dev/origin/infra/discovery/src/listener/tools/consistency/ident.js:173:13)
        at processTicksAndRejections (internal/process/task_queues.js:86:5)

And I couldn't find a way to actually get this to work.  If you know something I don't, let me know.

franckc: Yeah, looking a bit more into that it's actually a bit more convoluted than I thought. But this works:
```
class AssertError extends Error {
  constructor(msg) {
    super(msg)
    this.name = this.constructor.name
   }
 }

try {
  throw new AssertError('boo')
} catch (e) {
  if (e.name === 'AssertError') {
    console.log('AssertError !', e.message)
  }
}
```

mikeshultz: alright, that'll work

franckc:
Thanks for integrating my feedback. Merging in and will do a run in prod and let you know the output.

PR #2208, opened 2019-05-03
Meta tx readiness

nick:
Moving some things around to help facilitate the metatx work.


packages/graphql/src/mutations/_txHelper.js
```diff
@@ -26,44 +26,54 @@ export default function txHelper({
   mutation,
   onConfirmation,
   onReceipt,
-  from
+  from,
```
franckc: nit: since txHelper is going to a be a central function in the system, would be good to explicitly document the various arguments it takes as opposed to just relying on arg names.

packages/graphql/src/mutations/_txHelper.js
```diff
@@ -26,44 +26,54 @@ export default function txHelper({
...
 
-      try {
-        window.localStorage[`${contracts.net}Transactions`] = JSON.stringify(
-          contracts.transactions
-        )
-      } catch (e) {
-        /* Ignore */
-      }
+        try {
+          window.localStorage[`${contracts.net}Transactions`] = JSON.stringify(
```
franckc: Should the key be prefixed with the account's eth address ?
So that there is no conflict if user switches account in Metamask for example.

Issue #1217, opened 2019-01-10
Fractional Usage - Mobile Calendar Layout


jordajm:
The calendar component is not mobile-friendly. We need to design and implement mobile-friendly calendars for both buyer and seller.

PR #2211, opened 2019-05-04
Make Calendar component mobile-friendly

### LINKED ISSUE #2140: Fractional booking does not work on mobile
Reproduction steps:
 - Open Origin mobile app
 - Go to purchase page for a fractional listing
 - Try to select checkin date on calendar
=> Nothing happens

shahthepro:
Closes #2140 and #1217


tomlinton:
@nick did you have a chance to take a look a this? Do we want to take a different approach? It'd be cool to get this solved, I can look into it if we want to go a different way.

nick:
After talking to Aure we want to switch to a 'click for start date, then click for end date' instead of the 'click and drag for range' we have currently. That should solve the mobile issue I think

shahthepro:
@tomlinton  @nick 

> After talking to Aure we want to switch to a 'click for start date, then click for end date'

Have updated the component to support this behaviour on mobile. 

Still dates across months cannot be selected. That's on my to-do list. But that should not be a blocking issue. We can raise that as a separate issue.

nick:
Since we changed from the original idea of dragging on mobile, I was able to simplify the diff somewhat...

Issue #1760, opened 2018-09-28
Reviews Incentives & Finalization


micahalcorn:
In today's UX review, it became clear that we still are not completely settled on how to incentivize both buyers and sellers to leave reviews, protect the buyer from mistakenly releasing funds, and also minimize gas costs.

In our original purchase states, the review step was combined with each the buyer's confirmation of receipt and the seller's withdrawal of funds. As a result, both parties were required to leave a review containing at least a 1-5 star rating, and the number of blockchain transactions were reduced from four to two.

In the new marketplace contract, reviews are not required. The buyer's review is encouraged in the DApp by still having the review and finalization combined as one form submission, but the seller's review happens after the funds are already released. This presents two problems:

1. The buyer may instinctively leave a bad review and thereby unknowingly release the funds. We are addressing this for v1 by inferring the need for arbitration and intercepting the review process.
1. The seller has no incentive to return and leave a review. There is additional gas required AND the funds are already released.

Possible options are:

- require seller finalization after buyer finalization
- hold stake from both parties and release them in exchange for reviews
- settle for warning the buyer and create some incentive for the seller

This doesn't need to be settled for v1 but should be explored relatively soon.

cc @DanielVF @nick @matthewliu @joshfraser @wanderingstan 

Related: https://github.com/OriginProtocol/origin-js/issues/149 & https://github.com/OriginProtocol/origin-js/issues/365

sparrowDom:
I agree that combining leaving the review and releasing the funds into one step (though it seemed like a practical solution at one point) is confusing. Approaches that make sense to me: 
- reduce the current finalisation window from 1 year to something less conservative (2-3 weeks perhaps?) and give the seller an option to withdraw the funds after that period. 
- would be nice if we were to have solid notifications so we can message the buyer and the seller that finalisation period is coming to an end in case the offer is not finalised
- I really like @nick 's idea of incentivising the buyer to leave a review by giving him some seller's OGN tokens. Because this has another benefit of also incentivising sellers to boost listings with OGN since that will not only expose their listings but also increase their chances to get more reviews from buyers.

PR #2213, opened 2019-05-05
Add pagination to reviews

### LINKED ISSUE #2188: Only fetch data for visible reviews
All reviews are currently fetched for a seller - we should only fetch data for visible reviews

shahthepro:
Closes #2188


mikeshultz:
I was just poking around the reviews resolver working on something else and through this could really use some pagination.  Would be pretty great if the pagination in reviews actually prevented some of those `contracts.eventSource.getReview()` calls.  Each one of those makes an `eth_call` and multiple IPFS fetches that we could do without if those nodes aren't even being sent to the client.

Let me know if you have any questions, or I could add to this PR if you're busy.

shahthepro:
@mikeshultz
> Would be pretty great if the pagination in reviews actually prevented some of those contracts.eventSource.getReview() calls

I thought of doing that while working on this PR but skipped it since ratings are optional. The problem was that if there are any `OfferFinalized` event with no rating, the number of reviews returned for the requested page may be lesser or zero. PR #2143 will make the ratings required. Still we may face the issue with reviews left before merging #2143.

@nick @micahalcorn Any thoughts on this?

shahthepro:
@mikeshultz Here is the change you requested: https://github.com/OriginProtocol/origin/pull/2213/commits/4f71b18f12856b6aa2dd344b51c096c462c82629

micahalcorn:
Yeah, it's worth noting that we're soon going to separate ratings/reviews and finalization. They will no longer be combined in one event. See https://github.com/OriginProtocol/origin/issues/2422 and https://github.com/OriginProtocol/origin/issues/1760.

mikeshultz:
If I follow, a rating is required to even display the review?  So all the past reviews that people left without a rating are useless?  

Don't think I have a really good high level of what's going on here so I'll defer to everyone else.  Just saw you had work going on in this resolver and thought it might be an opportune time.  Perhaps not and it can be dealt with a little later.

But right now, the performance of this resolver is literally crippling.  The Origin Store user page won't even load because of it.  I'm working on some remediation techniques for this, but it's still an absurd amount of JSON-RPC calls and IPFS fetches to resolve this query.

shahthepro:
> If I follow, a rating is required to even display the review? So all the past reviews that people left without a rating are useless?

Yes, exactly. In the DApp, only reviews with ratings are shown as of now.

> But right now, the performance of this resolver is literally crippling. The Origin Store user page won't even load because of it. I'm working on some remediation techniques for this, but it's still an absurd amount of JSON-RPC calls and IPFS fetches to resolve this query.

I have added a changeset that calls `getReview()` method after filtering. It will not break anything on the DApp. However if the latest 5 reviews of any listing have no rating, the user will see blank list of reviews and a "Read More" button. Perhaps, in that case, we can fetch the next 5, by using `AutoMutate` component. The performance may still be affected if the listing has more number of reviews and none of it have rating.

PR #2216, opened 2019-05-06
Update withCurrencyBalances hoc to render after loading has completed

### LINKED ISSUE #2012: Insufficient Funds Flickering
When buyers land on a listing detail page, they often see the insufficient funds message for a brief moment before their balances are detected. We should just not render the CTA until loading is complete.

shahthepro:
Should close #2012

### DIFF
--- a/dapps/marketplace/src/hoc/withCurrencyBalances.js
+++ b/dapps/marketplace/src/hoc/withCurrencyBalances.js
@@ -30,12 +30,15 @@ function withCurrencyBalances(WrappedComponent) {
       variables={{ account: props.wallet, tokens: props.targets }}
       fetchPolicy="network-only"
     >
-      {({ data }) => (
-        <WrappedComponent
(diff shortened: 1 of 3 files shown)


dapps/marketplace/src/hoc/withCurrencyBalances.js
```diff
@@ -30,12 +30,15 @@ function withCurrencyBalances(WrappedComponent) {
...
       fetchPolicy="network-only"
     >
-      {({ data }) => (
-        <WrappedComponent
-          {...props}
-          currencies={get(data, 'currencies') || []}
-        />
-      )}
+      {({ data, loading }) => {
+        if (loading) return null
```
nick: Hrm in general I think we should avoid returning null from higher order components as it could cause items to not be rendered at all instead of showing a loading state or similar further down the chain

PR #2238, opened 2019-05-08
Notifications case fix & logging

wanderingstan:
- Fix problem where notification server wouldn't find Identities because of uppder/lower case differences in eth addresses. (Messaging server sends mixed-case eth addresses, while Eventlistener sends only lowercase)
- Add some logging to make future debugging easier
- Use sequelize `beforeCreate()` hook to force eth addresses in notification log table to always be lowercase. (Might consider using in our other Models?)
- Change test fixtures to have mixed-case eth address


infra/notifications/src/models/notificationlog.js
```diff
@@ -15,5 +15,12 @@ module.exports = (sequelize, DataTypes) => {
   NotificationLog.associate = function() {
     // associations can be defined here
   }
+  NotificationLog.beforeCreate(function(model, options) {
```
franckc: Cool - I did not know Sequelize supports that. Pretty handy.

PR #2250, opened 2019-05-09
Simulate standard browser user agent in webview

### LINKED ISSUE #2227: Google Attestation Not Supported In Mobile

tomlinton:
No idea what this one does 🙈 

Reconstructs a user agent that looks like the standard browser for the device. Closes #2227.

### DIFF
--- a/mobile/android/app/build.gradle
+++ b/mobile/android/app/build.gradle
@@ -162,6 +162,7 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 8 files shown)


micahalcorn:
This solves the issue for our webview, but we would still have the same problem in the various DApp browser wallets?

tomlinton:
That is true, assuming they use WebView (and don't do something similar).

PR #2251, opened 2019-05-09
Proxy Account and Meta Transactions

nick:
This PR adds:

- ProxyFactory contract (taken from Gnosis) for deploying new ERC725 Identity Proxy contracts with significantly less gas
- ERC725 Identity Proxy with Meta Transaction support. All of a user's interactions with the Marketplace and IdentityEvents contracts will be proxied via this. Users can also give permission for transactions to be run on their behalf by signing the transaction they wish to execute and passing the signature and transaction call data to a relayer server.

Since we're now supporting both direct wallet AND proxy identity interactions with our contracts, I have introduced the concept of an Account 'owner' in GraphQL. This will return the controlling wallet in the case of a Proxy account, or the same wallet address if the account is just a wallet.

For example, if the seller is a wallet `0xab` then both `listing.seller.id` and `listing.seller.owner.id` will return `0xabc`
If the seller is an account proxy `0xdef` with owner `0xabc`, then `listing.seller.id` will return `0xdef` and `listing.seller.owner.id` will return `0xabc`

The actual use of Proxy account is disabled by default. Use env var `ENABLE_PROXY_ACCOUNTS=1` to enable them.


packages/contracts/build/contracts/ProxyFactory_solc.json
```diff
@@ -0,0 +1,95 @@
+{
```
franckc: The other ABI files in this directory do not have the _solc suffix in their filenames.
Is it intentional that ProxyFactory_solc.json and IdentityProxy_solc.json do ?

packages/contracts/contracts/identity/IdentityProxy.s
```diff
@@ -0,0 +1,200 @@
+pragma solidity ^0.4.24;
```
franckc: For all the other contract files, we use the .sol extension.
Should we use that as well for IdentityProxy.s and ProxyFactory.s ?

packages/contracts/contracts/identity/IdentityProxy.s
```diff
@@ -0,0 +1,200 @@
...
+
+    function changeOwner(address _owner)
+        external
+        onlyOwner
+    {
+        owner = _owner;
+        emit OwnerChanged(owner);
+    }
+
+    function getData(bytes32 _key)
```
franckc: What are we planning to use getData and setData for ?

packages/graphql/src/utils/hasProxy.js
```diff
@@ -0,0 +1,38 @@
+import contracts from '../contracts'
+
+export default async function hasProxy(address) {
```
franckc: nit: add a bit of comments to explain this magic :)

packages/graphql/src/mutations/_txHelper.js
```diff
@@ -21,6 +24,12 @@ export async function checkMetaMask(from) {
 // to hang
 const isServer = typeof window === 'undefined'
 
+// function useRelayer({ mutation }) {
```
franckc: Is the plan to gate this with an env var as well ? like ENABLE_RELAYER ?

infra/bridge/relayer.js
```diff
@@ -0,0 +1,164 @@
+'use strict'
```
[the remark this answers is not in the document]
nick: Yea that what I was thinking

packages/contracts/build/contracts/ProxyFactory_solc.json
nick: It's because Truffle can only support either version 4 or 5 of Solidity but not both, and the new Proxy contract requires version 5. I've appended _solc to the contract names so we know they were compiled with Solc instead of Truffle.

packages/contracts/contracts/identity/IdentityProxy.s
nick: I had to name them .s so truffle doesn't recognize them and try to compile them. Not a great solution but the alternative was to to put them in a separate directory. I searched everywhere but unfortunately it's not possible to make truffle ignore certain contracts.

packages/contracts/contracts/identity/IdentityProxy.s
nick: These come as part of the ERC725 v2 standard. We could potentially use them for storing the IPFS profile instead of the EventLog method we're currently using. Since it's only one slot it shouldn't be too expensive.

packages/graphql/src/mutations/_txHelper.js
nick: Yep

packages/graphql/src/utils/hasProxy.js
nick: Will do

packages/contracts/contracts/identity/IdentityProxy.s
```diff
@@ -0,0 +1,200 @@
...
+        // uint256 txGas = 1 ether;
+        // solium-disable-next-line security/no-inline-assembly
+        assembly {
+            success := call(gas, to, value, add(data, 0x20), mload(data), 0, 0)
+        }
+    }
+
+    // copied from GnosisSafe
+    // https://github.com/gnosis/safe-contracts/blob/v0.0.2-alpha/contracts/base/Executor.sol
+    function executeCreate(bytes memory data)
```
shahthepro: Do we need this method? We are not deploying contracts with Proxy right now. Will it be needed in future? Unless we want the user's proxy to be not specific to Origin, I guess we can remove contract creation from this file.

packages/contracts/contracts/identity/IdentityProxy.s
```diff
@@ -0,0 +1,200 @@
...
+        bytes _offer,
+        address _exchange,
+        bytes _swap,
+        address _token,
+        uint _value
+    )
+        public
+        payable
+    {
+        owner = _owner;
```
shahthepro: Why override owner? Are we going to deploy a single proxy contract and make all users use that?

This function doesn't have `onlyOwner` modifier. Won't this allow any person to exploit this and take control of the contract?

shahthepro: An attacker can use this to make himself an owner of the contract and change the `Marketplace` address to his own contract that transfers the tokens (deposited for creating listing after the address change) to his account.

shahthepro: Ignore the previous comment. I thought that `_marketplace` is global variable. Didn't see that it was a function parameter. Anyway I'm still skeptical about the reassignment of owner address.

packages/contracts/build/contracts/ProxyFactory_solc.json
franckc: Got it. Thanks for explaining.
Probably worth adding a README.md to explain since I'm sure we'll forgot why it is the case and ask again... Same thing for the .s files.

packages/contracts/contracts/identity/IdentityProxy.s
franckc: Sounds good.

packages/contracts/contracts/identity/IdentityProxy.s
nick: We're using the ERC725 v2 standard for the account... though that could change later

nick:
@franckc @shahthepro ready for final review now... it's all wired off so we should be OK to merge to master

Issue #1047, opened 2018-12-05
Messaging Notifications


micahalcorn:
~Browser n~Notifications are currently only triggered by offer-related events. The notifications server should also be aware of messaging-related events and push these to subscribing recipients.

micahalcorn:
The status quo:

- [Here](https://github.com/OriginProtocol/origin/blob/master/origin-messaging/src/verify.js#L69-L76) the messaging server notifies the linking server of a new message (for mobile push notifications).
- [Here](https://github.com/OriginProtocol/origin/blob/master/origin-messaging/src/verify.js#L69-L76) the notifications server listens for offer-related events (from the [discovery listener](https://github.com/OriginProtocol/origin/blob/master/origin-discovery/src/listener/webhooks.js#L85-L113)).
- [Here](https://github.com/OriginProtocol/origin/blob/master/origin-notifications/src/app.js#L173-L180) the notifications server notifies the linking server of a new offer-related event (for mobile push notifications).

Perhaps the notifications server should be responsible for all posts to the linking server regardless of source. I'm not sure whether or not the linking server correctly handles those [here](https://github.com/OriginProtocol/origin/blob/master/origin-linking/src/logic/linker.js#L485-L503).

cc @Natasha08 & @tomlinton (in case the are tunneling restrictions that will need to evolve)

micahalcorn:
To Do (somewhat related): refine all of the copy for the actual message strings that are pushed.

micahalcorn:
Bumping this up to 🅿️1️⃣ in accordance with sprint planning and shifting the scope from browser push to email.

wanderingstan:
Gonna call this done with https://github.com/OriginProtocol/origin/pull/2257

micahalcorn:
Done ✅

PR #2257, opened 2019-05-09
No dupes for multiple messages

wanderingstan:
- Per discussion in yesterday's ENG call, this will ensure that every message will trigger an email and not get caught by the dupe checker. Basically it adds a hash of the message data to the message/email object, which is then hashed by the dupe checker when looking for duplicates. 
- A little UI fix in template putting space between text and the button. 
- Include eth addresses in error messages


wanderingstan:
@wanderingstan mark this issue done when merged: https://github.com/OriginProtocol/origin/issues/1047

PR #2258, opened 2019-05-09
handle the case where blocked account tries to enroll

sparrowDom:
### Description:
Handle the case so that un-enrolled account gets directed to blocked page in case that account is blocked. 

### Checklist:

### DIFF
--- a/infra/growth/src/apollo/resolvers.js
+++ b/infra/growth/src/apollo/resolvers.js
@@ -169,10 +169,20 @@ const resolvers = {
           )
         }
 
-        return { authToken }
+        return {
+          authToken,
+          isBanned: false
+        }
       } catch (e) {
-        logger.warn('User authentication failed: ', e.message, e.stack)
-        throw new AuthenticationError('Growth authentication failure')
+        if (e.message === 'This user is banned') {
+          return {
+            authToken: '',
+            isBanned: true
+          }
+        } else {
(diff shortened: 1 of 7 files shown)


infra/growth/src/apollo/resolvers.js
```diff
@@ -169,10 +169,20 @@ const resolvers = {
...
 
-        return { authToken }
+        return {
+          authToken,
+          isBanned: false
+        }
       } catch (e) {
-        logger.warn('User authentication failed: ', e.message, e.stack)
-        throw new AuthenticationError('Growth authentication failure')
+        if (e.message === 'This user is banned') {
```
franckc: I'm slightly worried that in the future someone may change the error message emitted and it will break.
To make this less breakage prone, what do you think of defining a new exception type and relying on e.name ?

Code sample:
```
class BannedUserError extends Error {
  constructor(msg) {
    super(msg)
    this.name = this.constructor.name
   }
 }

try {
  throw new BannedUserError(`User ${ethAddress} is banned`)
} catch (e) {
  if (e.name === 'BannedUserError') {
    console.log('Banned user error !', e.message)
  }
}
```

sparrowDom: Yeah that is a good point - your way is much more resilient... making the correction

sparrowDom:
@franckc corrected the error, PTAL

PR #2270, opened 2019-05-13
Revamp search scoring

DanielVF:
This should be a big improvement to current search scoring.

No scientific method here yet - just played with each improvement until it seemed good.

# Results

image

# Changes

### Centralized Scoring

I've centralized all scoring in a single script. This makes it way easier to figure out what is going, rather than the alternative crazy nesting of json scoring functions. This script is compiled once by the server, and reused for subsequent requests. 

I've used multiplication for all score modification, rather than addition. This way better textual match listings get positively moved up in results more than worse matches.

### Downrank listings that are not active

If a listing is not active, a generic search shouldn't rank it highly. If we were to filter a search to look for only sold listings, then this downranking will not affect the sold search ranking at all, since all sold listings would equally affected.

### Downrank cheap listings to reduce low quality listings in results

There are a lot of low quality wash listings on the site. I set an arbitrary cutoff of around $1.50 USD/DAI and it's current ETH equivalent. This feels like it catches 80% of them. Much better results.

### Redo OGN boosts to work off multiplication, not adding

By using multiplication, a boosted listing needs to be at least somewhat relevant in order to go to the front of the results. Otherwise for search results without strong real matches, irrelevant boosted listings would overwhelm the results. Boosted listings are still a little over powered. When relevant, they the go straight to the front of the line, but even just a tenuous connection,  they can still move up to near the font of results. Probably better to error on the side of power for now.

### Temporary boost for recently posted listings

Arbitrarily selected 15 day linear boost that adds up to 50% to the result score.

# Production

- You must rebuild the Elasticsearch listing index before use.


infra/discovery/src/lib/search.js
```diff
@@ -220,20 +220,70 @@ class Listing {
...
+              // - Script itself stays the same, and never needs to be recompiled
+              now: new Date().getTime()
+            },
+            source: `double score = _score;
+
+            // Downrank "cheap" listings to hide wash/spam transactions.
+            // We might want to make this a smoother transition later.
+            // ETH value may need to become a dynamic parameter later.
+            if (doc['price.amount'] != null && doc['price.currency.id'] != null) {
+              if(doc['price.currency.id'].value == "fiat-USD"){
```
sparrowDom: Another way to express this in a more "config" like way: 
```
const cheapListingsPenalty = {
 'fiat-USD' : {
    'minimumAmount': 1.5,
    'scoreReduction': 0.2
  },
   ...
}

if (cheapListingsPenalty[doc['price.currency.id'].value]) {
  const cheapListingPenalty = cheapListingsPenalty[doc['price.currency.id'].value]
  if(doc['price.amount'].value < cheapListingPenalty.minimumAmount) {
    score *= scoreReduction;
  }
}
```

matthewliu:
Excited to see this in action! I think this will go a long way in showing new users that we don't just have a bunch of crap on the platform.

infra/discovery/src/lib/search.js
franckc: How about the other currencies we currently support ? They are not super popular (see stats below) but might be worth adding them for future ?

origin=> select data->'price'->'currency'->'id', count(*) from listing group by data->'price'->'currency'->'id' order by count(*) desc;
  ?column?   | count 
-------------+-------
 "token-ETH" |  4730
 "fiat-USD"  |  4272
 "fiat-EUR"  |    84
 "fiat-KRW"  |    10
 "fiat-GBP"  |     6
 "fiat-JPY"  |     4
 "fiat-CNY"  |     4

infra/discovery/src/lib/search.js
```diff
@@ -220,20 +220,70 @@ class Listing {
...
+              if(commission > 100) {
+                commission = 100;
+              }
+              score *= (1.0 + doc['commissionPerUnit'].value * 0.000000000000000000025);
+            }
+
+            // Temporary boost for recently created listings
+            if (doc['createdEvent.timestamp'] != null) {
+              double recentBoostAmount = 0.5;
+              long boostPeriod = 1555200; // 15 days, linear drop off
```
franckc: Is that in seconds ? If yes, 1555200 is 18 days.
Perhaps express as long boostPeriod = 15 * 24 * 60 * 60 to make it easier in future to adjust the period.

infra/discovery/src/lib/search.js
DanielVF: I've added Euros. Looks like this takes care of 99% of the listings.

infra/discovery/src/lib/search.js
DanielVF: Thanks! Changed it to work in code.

infra/discovery/src/lib/search.js
DanielVF: Changed to a more data-driven approach. Thanks!

DanielVF:
Should be final.

Issue #2281, opened 2019-05-14
GraphQL Service & Performance Mode


mikeshultz:
Looking to add a "Performance Mode" to the dapp, especially for use with Mobile.  This will allow clients to make GraphQL requests against a server we have setup that will do better caching and ideally perform better in general when making the needed background requests.

There's a few pieces to this project:

- Create a GraphQL service
- Make necessary alterations to client to make network requests when configured to do so
- Setup "Performance Mode" UI for the dapp

### Notes & Considerations

- For the time being at least, the search logic will stay in discovery
- Not all Resolvers are available for server-side.  Perhaps an Apollo "link" needs to be created to make sure not to send these exceptions via network? 
- Look into including [Apollo metrics](https://www.apollographql.com/docs/apollo-server/features/metrics) in service

PR #2289, opened 2019-05-15
Performance mode

mikeshultz:
### Description:

This is the PR for __PERFORMANCE MODE__.  Includes:

- `context.config.performanceMode` bool value is now a thing
- Client updated to "route" queries to remote if the types/schemas are available and performance mode is enabled
- Service wired up, pretty much stock from the original `npm run start --prefix packages/graphql` with the exception of the addition of service port configuration by the `GRAPHQL_SERVER_PORT` env var and addition of Apollo metrics support. Also added a check and block to prevent mutations from running
- Dapp integration (initial, have not yet received mockups and verbiage)

### TODO

- I need someone to add the proper `GRAPHQL_SERVER_PORT` (`4002`) to envkey, and verify the chart templates are configured for it correctly
- I need someone to add an Apollo metrics API key as the env var `APOLLO_METRICS_API_KEY`(or `ENGINE_API_KEY` but I prefer the former for clarity).  An [API key can be had on their site](https://engine.apollographql.com/).  I assume we only want this for prod.
- Do all the remote types included in `src/typeDefs/remote.js` make sense?  This may reveal itself during testing, either way. 
- Mockup if necessary, and verbiage including tooltip (waiting on Matt)
- I still need to run `npm run translate`!

### Checklist:

Ref: #2281


mikeshultz:
CORS might need to be configured, it wasn't clear from the apollo docs if default CORS settings were permissive or not.  Will need to test in dev real quick and get a fix in if it fails.

packages/graphql/src/server.js
```diff
@@ -15,7 +15,52 @@ const schema = makeExecutableSchema({
...
+        console.warn(
+          `Mutations not allowed.  Operation: ${operation.operationName}`
+        )
+        throw new Error('Mutations not allowed')
+      }
+    }
+  }
+}
+
+const API_KEY = process.env.ENGINE_API_KEY || process.env.APOLLO_METRICS_API_KEY
```
franckc: Why using 2 env vars ?

franckc:
Added @tomlinton as a reviewer to double check the service configs.

devops/dockerfiles/origin-graphql
```diff
@@ -0,0 +1,22 @@
+FROM node:10 as build
+
+WORKDIR /app
+
+ENV NODE_ENV=production
+ENV GRAPHQL_SERVER_PORT=4002 # eh?
```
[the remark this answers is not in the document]
mikeshultz: Not sure why I even had that there to begin with, thanks.

devops/kubernetes/values/origin/values-dev.yaml
```diff
@@ -60,3 +60,7 @@ adminImageTag: '09d63fe'
 # origin-relayer
 relayerImage: origin-relayer
 relayerImageTag: 'latest'
+
+# origin-graphql
+adminImage: origin-graphql
```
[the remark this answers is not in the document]
mikeshultz: Good catch!

packages/graphql/src/link.js
```diff
@@ -1,17 +1,27 @@
...
-const link = ApolloLink.split(
-  operation => {
-    const definition = getMainDefinition(operation.query)
-    return definition.operation === 'subscription'
-  },
-  new SubscriptionsLink(),
-  SchemaLink
-)
+/**
+ * Making decisions on whwat to send the query to.  Either the subscription
```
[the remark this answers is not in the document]
mikeshultz: whwat?

packages/graphql/src/server.js
mikeshultz: `ENGINE_API_KEY` is supported by ApolloServer and is what's in their docs.  I preferred `APOLLO_METRICS_API_KEY` because it's purpose is clearer.  I thought it best to support the one in the docs in case someone uses it.  If I didn't add support for `ENGINE_API_KEY` and someone used it, it might cause some unexpected behavior that I thought best to avoid.

packages/graphql/src/link.js
```diff
@@ -1,17 +1,27 @@
+import Observable from 'zen-observable'
```
franckc: Does zen-observable need to be added to package.json ?

mikeshultz: Shouldn't be necessary unless you want to be explicit.  It's already a dependency of apollo.

devops/dockerfiles/origin-graphql
```diff
@@ -0,0 +1,22 @@
+FROM node:10 as build
+
+WORKDIR /app
+
+ENV NODE_ENV=production
+ENV GRAPHQL_SERVER_PORT=4002
```
franckc: How about reading this from EnvKey rather than hardcoding here ?

packages/graphql/src/link.js
franckc: sounds good.

packages/graphql/src/server.js
franckc: Ok. Maybe add a comment.

micahalcorn:
Just dropping by to say that we haven't forgotten the settings page design ⚙️ It neeeds another revision or two but should be ready tomorrow or Wednesday.

mikeshultz:
Out of curiosity, do you expect major changes, @micahalcorn?

micahalcorn:
@mikeshultz mostly copy and getting a better understanding of how it works. I'm still trying to understand what all can be configured. @franckc, please edit this with corrections:

- ⚙️ Configurable
- 🗿 Fixed
- ☠️ Not Supported
- `default`

|Variable|Performance|Decentralized|
|---|:-:|:-:|
|Discovery Server|🗿 `discovery.originprotocol.com`|☠️|
|Attestation Server|⚙️ `bridge.originprotocol.com`|⚙️ `bridge.originprotocol.com`|
|IPFS Gateway|⚙️ `ipfs.originprotocol.com`|⚙️ `ipfs.originprotocol.com`|
|Ethereum Node|☠️|⚙️ `mainnet.infura.io`|
|Meta Tx Relayer|⚙️ `?.originprotocol.com`|☠️|

franckc:
@nick : could you take a look at the graphql changes ?

@tomlinton : could you double check the configs ? (i'll add the secrets in a separate PR)

devops/kubernetes/charts/origin/templates/graphql.deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+  template:
+    metadata:
+      labels:
+        app: {{ template "graphql.fullname" . }}
+        release: "{{ .Release.Name }}"
+    spec:
+      containers:
+      - name: origin-graphql
+        image: "{{ .Values.containerRegistry }}/{{ .Release.Namespace }}/{{ .Values.graphqlImage }}:{{ .Values.graphqlImageTag }}"
+        imagePullPolicy: Always
```
tomlinton: Might need a definition of the port here, i.e.:

`ports:
    - containerPort: 4001
`

It'll still work without it but its good practice.

tomlinton:
Kubernetes part looks good, just one minor thing I commented on! Good work @mikeshultz!

dapps/marketplace/src/mutations/SetNetwork.js
```diff
@@ -2,6 +2,6 @@ import gql from 'graphql-tag'
 
 export default gql`
   mutation SetNetwork($network: String!, $customConfig: ConfigInput) {
-    setNetwork(network: $network, customConfig: $customConfig)
```
nick: Hrm should this have been removed?

mikeshultz: No... not at all.  And I'm quite curious how I didn't notice this in testing since my changes depend on this.  I even added support for the `customConfig` param in the backend in this PR...

Thank you for catching this!

mikeshultz:
Thank you all for your thorough reviews!  I think all feedback has been addressed. Look forward to testing on dev.

micahalcorn:
**Update** @mikeshultz @nick @tomlinton @sparrowDom @shahthepro 

After going through the exercise above with the table, we realized that there's not a clear mapping of configuration options to a binary choice of performance mode vs decentralized mode. As @joshfraser says, the main question is whether the user "trusts Origin" or not.

The conclusion that we reached is that we should have three high-level states:

1️⃣ Standard/default/performance mode: you can see all of the values for the configuration options but not edit them. They all point to Origin endpoints or those that we trust (e.g., Infura).
2️⃣ Custom/advanced/developer mode: you can edit any or all of the configuration options and it will probably, mostly work. Attestations, for example, are a bit complicated because a user might hypothetically want to trust us for attestations that have been published but run their own server for signing new ones. But this could still be considered a "performance mode" if a developer is running her own discovery server.
3️⃣ Decentralized mode: you click a button and a new window opens running the DApp directly on a public IPFS URL displaying raw data. Discovery, meta transactions, etc are not possible. We should work incrementally to support this.

@matthewliu and @auregimon and I will revisit the settings page design to match this new line of thinking. @franckc will file an issue outlining what we would need to get a minimal decentralized mode working.

mikeshultz:
@micahalcorn to be clear, does any of that impact this PR, or can this move forward and work to complete the rest follows?

Also, I'm not sure your table is accurate.  You should still be able to set a discovery server for performance mode.  This PR introduces a new service at `graphql.originprotocol.com` that's provided specifically for this centralized value-added service.  The discovery/search service may be integrated with this(or the other way aorund) at some point but will require further work.

franckc:
I'd vote for landing what we have now so that we can start testing and implementing changes to settings page in a separate PR. @micahalcorn cool by you ?

micahalcorn:
@mikeshultz yeah, the table is now no longer relevant. I'm totally fine with merging this whenever, but give us a day or two to redesign the settings page. Then that front-end work can be separate PR.

Issue #2136, opened 2019-04-26
Force/recommend updates on mobile


tomlinton:
We need the option in the mobile app to:

- Force an update, if we drop support for an older version for whatever reason
- Recommend an update, e.g. for stability fixes

I'll add a file into the mobile repository with version numbers for each of these. We'll need to add a component that blocks/nags accordingly. I think this is important to do soon because it is already becoming hard managing older versions.

@micahalcorn should we get mocks or I can just proceed with what I think looks OK.

micahalcorn:
@tomlinton just throw something together when you're ready and we can make it prettier when we get around to it. I'll put it on the list of things for @auregimon, but he has other priorities that are higher, I think.

tomlinton:
The App Store has this in the policy document:

`Apps must not force users to rate the app, review the app, download other apps, or other similar actions in order to access functionality, content, or use of the app.`

So I'm not sure from that whether that includes forcing the user to update the app. I think we should err on the safe side and still allow the user to use the app/dismiss the message but make the wording stronger for the "Force an update" part of this.

Any thoughts?

micahalcorn:
We still want the user to be able to "use the app" because she may want to export her key, for example. I think the main thing is just for it to be abundantly clear that the version is no longer supported and that things will not work as expected. If she wants to do anything meaningful (buy, sell, etc), she needs to update it.

tomlinton:
Closed #2291.

Issue #2132, opened 2019-04-26
Mobile Authentication


micahalcorn:
Encrypt the data in Origin Marketplace and require authentication for certain actions, such as accessing the recovery phrase. The authentication method would ideally involve a biometric (face/fingerprint) test or fall back to a six-digit pin if the user declines access to the device's native feature.

tomlinton:
Closed #2291.

Issue #2131, opened 2019-04-26
Recovery Phrase Setup


micahalcorn:
Prompt users to back up their recovery phrases when funds are detected.

https://app.zeplin.io/project/59fa2311bac7acbc8d953da9?seid=5c99591046989524fb7f0a1e

tomlinton:
Closed #2291.

Issue #1918, opened 2019-04-04
Improved Mobile Onboarding + Mnemonic-Based Wallet


micahalcorn:
Also known as "mobile 1.75", the version of our Android/iOS app following https://github.com/originprotocol/origin/issues/1917 should feature an improved oboarding experience and account management using seed phrases instead of private keys.

-----

At Origin, we use [ZenHub](https://www.zenhub.com/) to manage our engineering tasks and product development. [Download their browser extension](https://www.zenhub.com/extension) and check out our open workspace at [github.com/originprotocol/origin#zenhub](https://github.com/originprotocol/origin#zenhub).

micahalcorn:
Noting here that we probably want to continue to allow for importing and exporting private keys, but I think we would still want to require the involvement of a recovery phrase for other accounts 🔑 We may need to add an indicator to those that don't follow the derivation path (as MetaMask does).

tomlinton:
Closed #2291.

PR #2291, opened 2019-05-16
Mobile 1.75

### LINKED ISSUE #2130: Mobile Origin Account Onboarding Components
Build the UI components for the new onboarding for Origin Marketplace as shown [here](https://app.zeplin.io/project/59fa2311bac7acbc8d953da9?seid=5c996bf879d02b35ea3596a9). These designs are likely to change with the introduction of Meta Transactions and the "Origin Account" concept.

tomlinton:
This is everything from 1.75. Features:

- New onboarding flow
- Pin/Touch/Face authentication
- Backup prompts
- Multiple mnemonic/account management
- Update prompt screens

There'll probably be a few fixes coming in as we test the beta builds, but we can merge this now. Closes #2130, #2136, #2132, #1149, #1916, #2131, and #1918. A record maybe?


micahalcorn:
Great comments 👏

I think we need a way to navigate back (cancel) from the mnemonic import form.

micahalcorn:
@tomlinton is the backup prompt only enabled on Mainnet? I can't seem to figure out how to get it to show up.

Also, is there a way to test the update prompt with the TestFlight version? Maybe ship another one now that I have 0.8 (2) installed?

franckc:
@micahalcorn asked me to put my testing feedback on this PR...
Tested on an iPhone 6S. App worked well ! Nice work :)
Some feedback on the UX:
- Webview page: there is a small white area on top. It stays even when scrolling down. See screenshot below.

 - App navigation: This is totally subjective but the icons for webview and wallet were not self explanatory to me. I had to click on them to understand what these were.
 - Webview navigation: The hamburger menu and profile/settings menu is not sticky at the top. So as I scroll down on listing page for example, I have to scroll back up to access them which might be cumbersome/confusing for users.
 - If I don't have enough ETH and try to publish my identity or buy a listing the error message saying so is red but in small print. Should probably be bigger. Also if I click cancel on that page, I then get an error page "You decline to sign the transaction" which seems unnecessary.
 - It seems confusing to have Settings both in webview hamburger menu and the app and to point to different pages.
 - Not specific to mobile. If they have no ETH, it is frustrating to let user go thru the listing creation flow, enter all the data and pick pictured , to then get told in last step (publish) they can't publish...

micahalcorn:
Thanks @franckc 👍 Here are some responses to some of your comments for posterity:

- webview header: This white space is a consequence of having to accommodate the "safe area" on newer iPhones. It shouldn't be there on older ones, so we'll investigate.
- footer icons: We decided to remove the labels because there is no real consequence to clicking the tabs and learning what they do very quickly.
- navigation: sounds like a vote for native 😁
- error: noted 🗒 
- funding: should be much less of a concern with meta transactions in place 🤞 We'll also escalate funding in the UX if we can come up with a satisfactory way to actually help the user instead of shrugging.

I don't think that any of these are specific to this PR, so I'll open separate issues where appropriate.

tomlinton:
@micahalcorn:

- Sounds good, will disable it.
- I think I'll do a similar thing and disable the button so it is clearer. Unless you want something more obvious.
- Will do
- Theres a few reasons why I don't want to do that:
  - Those are alerts so the UI is a bit strange if you display a screen requesting auth then display an alert
  - Its more code to block those individual features rather than the whole account screen
  - Switching the active account and renaming accounts are important enough actions to warrant the extra authentication step
  - It'll be fairly common to do a "Show Recovery Phrase" (to verify it is what you want) followed by a "Copy Recovery Phrase" which will result in two authentication prompts. We could make it so that if either is authed then the second prompt doesn't happen but that is adding quite a lot of complexity.
- Will do
- Will do
- Easy change if we want it. Just going back to the account list is a bit softer, in my mind the user will look through the list and find the account they want (even though it was already there).

The backup prompt displays under the following circumstances:
- The account was created (i.e. from the "Create a wallet" on the first onboarding screen, which is currently the only way we have to get a random account)
- The account gets a balance

So you'll probably have to create a random account, and transfer some Eth to it on Rinkeby for it to display. Or I can push a test build that'll always display it. I can do the same for the update prompt or we can temporarily commit a new `mobile/.version` file with higher versions.

tomlinton:
Covered off in the list except the two I commented on above. Fixed that inset issue @franck (haven't pushed a new test build yet). I think we can merge this and I'll push out new test builds so we can test the update prompt (this has to be merged for it to work)?

PR #2302, opened 2019-05-17
fixes clean-package-locks.sh for new package structure

mikeshultz:
I assume this changed with the package restructuring.  Tested and works as expected.


franckc:
Thanks for cleaning this up @mikeshultz 

Should we take this opportunity to delete all package-lock.json files that are checked in and also add package-lock.json to .gitignore ? @tomlinton what do ya thinka ?

PR #2310, opened 2019-05-20
Remove deprecated avatar writes from DAPP

DanielVF:
This stops the DAPP from writing the old style, base64 `avatar` fields to the user profiles. We're already only reading the new `avatarURL` fields from the dapp, so this change should be non-noticeable.

...getting near the end of issue #1795.


dapps/marketplace/src/pages/user/_EditModal.js
```diff
@@ -52,13 +51,15 @@ class EditProfileModal extends Component {
           <div className="row">
             <div className="col-6">
               <ImageCropper
-                onChange={async avatar => {
+                onChange={async avatarBase64 => {
                   const { ipfsRPC } = this.props.config
-                  const uploadedImages = await uploadImages(ipfsRPC, [avatar])
+                  const uploadedImages = await uploadImages(ipfsRPC, [
+                    avatarBase64
```
franckc: Do you want to add a comment explaining why we upload base64 encoded image ?
This is just because of legacy.

Do you actually think it would be worthwhile uploading as binary moving forward ?
On the plus side, it would reduce a bit size and computation needed on client to decode. But may not be worth the extra complexity in the code to handle legacy base64 format ?

DanielVF:
Ready for merge to master

PR #2313, opened 2019-05-20
Makes the avatar backfill process parallel.

DanielVF:
About a 6x speedup at this low level of concurrency.

Of course, in theory now that I have run it today, we won’t every need to run it again. But, still nice to have around incase things ever get unpinned for some reason.


packages/graphql/src/scripts/201904_backfill_avatar.js
```diff
@@ -39,67 +44,89 @@ async function backfillAvatarToIpfs() {
...
-        console.error(calculatedUrl, actualUrl)
+      if (ONLY_ACCOUNT !== undefined && ONLY_ACCOUNT != account) {
         continue
       }
-      showStatus('SUCCESS: Avatar uploaded to IPFS')
+      tasks.push({ account, identityIpfsHash })
     }
   }
+
+  for (const chunklet of chunk(tasks, CONCURRENCY)) {
```
franckc: This totally works. An alternative that might be simpler FYI is to use the p-limit package.
See https://github.com/OriginProtocol/origin/blob/master/infra/discovery/src/listener/listener.js#L173

PR #2317, opened 2019-05-21
CircleCI

nick:
- Replace Travis with CircleCI
- Use Yarn instead of NPM
- Use Yarn Workspaces instead of lerna hoist
- Result: much faster builds 🙂- about 6 mins from start to finish compared to 13 mins for Travis
- Faster node_modules installation - now takes 90 seconds for a fresh clone.


tomlinton:
Looks good! There is probably a bunch of documentation that needs updating (stuff that references `npm install`), or was the plan to support both npm and yarn?

.circleci/config.yml
```diff
@@ -0,0 +1,160 @@
...
+    steps:
+      - attach_workspace:
+          at: .
+      - run: yarn run test --prefix packages/contracts
+
+  Discovery:
+    <<: *db
+    environment:
+      DATABASE_URL: postgres://postgres@localhost/travis_ci_test
+      AFFILIATE_ACCOUNT: 0x821aea9a577a9b44299b9c15c88cf3087f3b5544
```
franckc: These ACCOUNT definitions are from origin-js. Not needed anymore.

.circleci/config.yml
```diff
@@ -0,0 +1,160 @@
...
+    - image: circleci/node:10-browsers
+  working_directory: ~/origin
+
+db: &db
+  docker:
+    - image: circleci/node:10-browsers
+    - image: circleci/postgres:9.6-alpine-ram
+      environment:
+        POSTGRES_USER: postgres
+        POSTGRES_DB: travis_ci_test
```
franckc: super nit: would that make sense to call that db circle_ci_test ?

.circleci/config.yml
nick: Nice, have removed these now

.circleci/config.yml
nick: Updated 🙂

nick:
Assuming this PR doesn't cause issues for Docker / Mobile / Deployments I think we can merge it

tomlinton:
Updated some more things, I didn't test all the builds but it'll take too long locally, we can just merge it and let the Google builds tell us if anything is wrong. :grin:

PR #2327, opened 2019-05-22
MetaTX wip

nick:
Updates to relayer

### DIFF
--- a/infra/relayer/src/relayer.js
+++ b/infra/relayer/src/relayer.js
@@ -48,28 +48,57 @@ class Relayer {
    */
   constructor(networkId) {
     this.networkId = networkId
+    let privateKey
 
     // Load contract addresses based on network id.
     switch (networkId) {
       // Mainnet
       case 1:
         this.addresses = require('@origin/contracts/build/contracts_mainnet.json')
+        this.web3 = new Web3(
+          'https://mainnet.infura.io/v3/98df57f0748e455e871c48b96f2095b2'
+        )
+        privateKey = process.env.FORWARDER_PRIVATE_KEY_1
         break
       // Rinkeby
       case 4:
         this.addresses = require('@origin/contracts/build/contracts_rinkeby.json')
+        this.web3 = new Web3(
+          'https://eth-rinkeby.alchemyapi.io/jsonrpc/D0SsolVDcXCw6K6j2LWqcpW49QIukUkI'
+        )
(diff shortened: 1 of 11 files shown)


infra/relayer/src/relayer.js
```diff
@@ -48,28 +48,57 @@ class Relayer {
...
     this.networkId = networkId
+    let privateKey
 
     // Load contract addresses based on network id.
     switch (networkId) {
       // Mainnet
       case 1:
         this.addresses = require('@origin/contracts/build/contracts_mainnet.json')
+        this.web3 = new Web3(
+          'https://mainnet.infura.io/v3/98df57f0748e455e871c48b96f2095b2'
```
franckc: How about we use env var PROVIDER_URL ? I can set it up in ENVKEY

infra/relayer/src/relayer.js
```diff
@@ -48,28 +48,57 @@ class Relayer {
...
+        )
+        privateKey = process.env.FORWARDER_PRIVATE_KEY_1
```
franckc: Should this just be FORWARDER_PK without network and we can set different value for network 1,4,2222 in EnvKey ?

infra/relayer/src/relayer.js
nick: we could... I always find env vars a pain to manage when you're developing locally as you have to manually switch them when you want to change environments though

franckc: If you want we can hardcode for 999 but use envkey for others ?
Having values in envkey in prod makes it lot easier to view what is being used rather than having to grep the code...

nick: Sure can do

nick: So to work with rinkeby or mainnet locally, I have to manually go into .env and update the provider?

franckc: To work on rinkeby/mainnet locally you would set PROVIDER_URL. No need to use .env / envkey if you don't want to.

Some alternatives:
1. hardcode default values. like const providerUrl = process.env.PROVIDER_URL || 'https://mainnet.infura.io/v3/98df57f0748e455e871c48b96f2095b2'
2. move the config that defines the provider url out of graphql, to a separate package. perhaps packages/config. and could have default values as in 1

Personally I like 2 the best for the long-term. But will take some work. So in short-term I'm fine with whatever works to unblock the testing of the relayer :)

PR #2328, opened 2019-05-22
Mobile currency exchange fix

tomlinton:
This is the fix that went into the latest release. I added detection for the DApp currency setting in mobile so that when it changes the transaction notifications update accordingly. Also removed previously hard coded exchange rates which I somehow missed. 🙈


micahalcorn:
Approved, I guess 👺

tomlinton:
You abandoned me for a cooler project, this is what you get 👹

micahalcorn:
I got heavily peer-pressured. 👬

PR #2331, opened 2019-05-23
Mobile localisation first pass

tomlinton:
This is most of the mobile localisation using fbt and crowdin. It's not complete yet, but we can merge it now and test the crowdin integration and get things moving. I've set up the integration. Things left to do:

- Detect best available language for the device
- Add a language option in settings?

Thanks @micahalcorn for your work on wrapping strings.


mobile/src/components/update-prompt.js
```diff
@@ -106,15 +117,15 @@ class UpdatePrompt extends React.Component {
               type="primary"
               style={styles.button}
               textStyle={{ fontSize: 18, fontWeight: '900' }}
-              title={'Update'}
+              title={fbt('Update', 'UpdateScreen.updateButton')}
```
micahalcorn: @tomlinton so we don't need to wrap this in `String` if it is being passed as a prop to a component?

tomlinton: No, it just has to be wrapped for certain things like react-native <Alert />. I think basically its fine everywhere we pass it as a prop.

PR #2345, opened 2019-05-27
[Listener] Exit process when exception caught in main loop

franckc:
### Description:

Force listener process to restart when catching an exception in its main loop.

### Checklist:

### DIFF
--- a/infra/discovery/src/listener/listener.js
+++ b/infra/discovery/src/listener/listener.js
@@ -233,6 +233,7 @@ setNetwork(config.network)
 try {
   main()
 } catch (err) {
-  logger.error('Error occurred in listener main() process')
-  logger.error(err)
+  logger.error('Error occurred in listener main() process:', err)
+  logger.error('Exiting')
+  process.exit(-1)
 }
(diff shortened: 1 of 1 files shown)


infra/discovery/src/listener/listener.js
```diff
@@ -230,9 +230,8 @@ logger.info(
...
-try {
-  main()
-} catch (err) {
-  logger.error('Error occurred in listener main() process')
-  logger.error(err)
-}
+main().catch(err => {
+  logger.error('Error occurred in listener main() process:', err)
+  logger.error('Exiting')
+  process.exit(-1)
```
mikeshultz: Should probably not use a negative number here.  Can cause unexpected behavior on *nix systems.  Use a number `> 0` to signal an error.

franckc: good point. done.

mikeshultz:
LGTM, but I don't see an option to approve now?

franckc:
I re-requested your review @mikeshultz - does that work now ?

PR #2349, opened 2019-05-27
Consistent AvatarUrls

### LINKED ISSUE #2322: [GraphQL] avatarUrl format is inconsistent
I ran a listener backfill in prod (e.g. mainnet) to populate the DB table column identity.avatarUrl
Once that completed, when I looked at the data I noticed some entries have an IPFS URL while some others have just an IPFS hash.

The listener is storing the [avatarUrl field](https://github.com/OriginProtocol/origin/blob/master/infra/discovery/src/listener/handler_identity.js#L219) and not the avatarUrlExpanded field so I was expecting only IPFS hashes ?

Here is a sample of the data: https://gist.github.com/franckc/90c6f4251c8f4a30fbf369dde72efc7b

DanielVF:
I had overlooked that automatically "upgraded" `avatar` to `avatarUrl` fields were being set from GraphQL without an IPFS protocol in the URI (eg they looked like 'QmS4...', rather than 'ipfs://QmS4...'.) This is fixed in this PR and the correct format is now returned.

However, we also have 149 cases where a user profile was edited with an old style avatar data field and no change was made to the avatar image. If this happened, then the "ipfs://"-less url would be saved into the profile JSON. I've added a check to GraphQL that will fix data up, so that everything coming from out of the GraphQL avatarUrl field will be correct and consistent.

Fixes #2322

### DIFF
--- a/packages/graphql/src/resolvers/IdentityEvents.js
+++ b/packages/graphql/src/resolvers/IdentityEvents.js
@@ -149,7 +149,7 @@ export function identity({ id, ipfsHash }) {
     // We have 149 identity.avatarUrls missing the ipfs:// protocol.
     // Prepend ipfs:// if needed.
     if (
-      identity.avatarUrl.length == 46 &&
+      identity.avatarUrl.length === 46 &&
       identity.avatarUrl.indexOf('://') === -1
     ) {
       identity.avatarUrl = 'ipfs://' + identity.avatarUrl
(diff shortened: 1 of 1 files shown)


packages/graphql/src/resolvers/IdentityEvents.js
```diff
@@ -137,14 +137,24 @@ export function identity({ id, ipfsHash }) {
...
         // If we can't translate an old avatar for any reason, don't worry about it.
         // We've already tested the backfill script, and not seen a problem
         // for all valid avatar images.
       }
     }
 
+    // We have 149 identity.avatarUrls missing the ipfs:// protocol.
+    // Prepend ipfs:// if needed.
+    if (
+      identity.avatarUrl.length == 46 &&
```
franckc: super nit: use === operator

PR #2350, opened 2019-05-27
Disables performance mode for docker network and listener, with some docker config fixes

mikeshultz:
### Description:

Disables performanceMode for the listener, and also the docker network.  Thanks, @DanielVF  for pointing this out.  Also fixes docker-compose and docker network configuration so it references the backend services instead of 'localhost'.

### Checklist:


infra/discovery/src/listener/listener.js
```diff
@@ -236,7 +236,7 @@ logger.info(
   )}`
 )
 
-setNetwork(config.network)
+setNetwork(config.network, { performanceMode: false })
```
franckc: Good catch! So that was causing the listener to talk to graphql.originprotocol.com in prod ?

packages/graphql/src/server.js
```diff
@@ -13,7 +13,7 @@ try {
   console.warn('EnvKey not configured')
 }
 
-setNetwork(process.env.NETWORK || 'test')
+setNetwork(process.env.NETWORK || 'test', { performanceMode: false })
```
franckc: Is the performanceMode config used in any ways by the graphQL server ?
Or is this just defensive coding for future ?

infra/discovery/src/listener/listener.js
mikeshultz: I think so.  It isn't guaranteed to cause problems, but I thought it best to make this change to prevent any weirdness, or to run into those same caching issues we had before.

packages/graphql/src/server.js
mikeshultz: This was completely defensive, and a little unnecessary.  No, it should never actually come into play.  I could probably remove this if you want, but it made me feel better.

DanielVF: I'm happy to have it here!

franckc: Sounds good - thanks for explaining.

PR #2351, opened 2019-05-28
Reduce backups

tomlinton:
Restrict the backups to prod and things in the Origin cluster. Previously we backed up everything (including Ethereum nodes which is pretty wasteful). This is to reduce our Google Cloud bill in anticipation of our credits expiring. 

I have built/pushed this container and update the backup cronjob. The excess backups we have will expire after 14 days so there is no need to manually remove.


franckc:
Pls merge when you are ready...

PR #2363, opened 2019-05-29
[Discovery] Scoring to javascript

### LINKED ISSUE #1015: [multi-units listing] Use per-unit commission for search ranking
For multi-units listing, search ranking should use the per-unit commission value rather than the total commission. So [this code](https://github.com/OriginProtocol/origin/blob/master/origin-discovery/src/lib/search.js#L187) should be updated to take into account the new perUnitCommission field being added to the data stored in IPFS for unit listings as part of #977

### LINKED ISSUE #2284: Boosted Listings broken in DApp/GraphQL
Boosted listings are currently broken in the DApp / GraphQL. It seems that the data is neither being saved correctly, nor read correctly.

@franckc created listing https://dapp.originprotocol.com/#/listing/1-000-2768, boosting it with 50 OGN. 

## Confusing results

GraphQL returns

```
{
  marketplace {
    listing(id:"1-000-2768") {
        __typename
      ... on UnitListing {
        commission,
        commissionPerUnit,
        deposit
      }
    }
  }
}
```

->

```
{
  "data": {
    "marketplace": {
      "listing": {
        "__typename": "UnitListing",
        "commission": "0",
        "commissionPerUnit": "0",
        "deposit": "50000000000000000000"
      }
    }
  }
}
```

It has zeros for both commission and commissionPerUnit.

IPFS data shows a 0 commission, 50 OGN commissionPerUnit.

https://ipfs.originprotocol.com/ipfs/QmUwXf41STd8aVf749tyRLHvAhrKdEpumSSdhsM4pc4orr

```
{
  ... snip ...
  "commission": {
    "currency": "OGN",
    "amount": "0"
  },
  "commissionPerUnit": {
    "currency": "OGN",
    "amount": "50"
  }
}
```

## Fields (on GraphQL)

- deposit: The total amount of OGN that the user has placed on deposit for the listing
- commission: Essentially a duplicate of the Deposit field. 
- commissionPerUnit: Amount of commission that can be paid out per unit sold.

## Changes needed

- [ ] Update the DApp to save the correct commission?
- [ ] Update GraphQL to save the correct commission
- [ ] Update GraphQL to correctly read in the commission
- [ ] Add a new commissionNextSale field to GraphQL that computes what a marketplace would receive for a sale that happened at this moment  `min(commission, deposit -  sum_of_pending_ commissions)`
- [ ] Unit test the GraphQL side of boosted listings for each listing type.
- [ ] Update discovery to use the new commissionNextSale

### See also

- #1015
- #1167
- https://github.com/OriginProtocol/origin/pull/1296/files

DanielVF:
(Need to verify actual search behavior against production listings before merging)

An incremental step that moves almost all listing scoring out of Elastic Search and into Javascript. 

### Search improvements

- Downrank listings with no images
- Fix boosting of boosted listings

### Also

- Fix broken commission calculating code in GraphQL
- Add a command to “in-place” reindex an index in es-cli. This allows easy local modification of an index mapping.
- Sweet tests for the scoring

Will close #1015 and will close #2284

### Deployment

- Will require a rebuild of the ElasticSearch index, since something is added to the mapping
- Will require a rerun of the listener to populate the scores.


packages/graphql/test/index.js
```diff
@@ -235,7 +235,7 @@ describe('Marketplace', function() {
           },
           category: 'Test category',
           subCategory: 'Test sub-category',
-          commission: '1.5'
+          commissionPerUnit: '1.5'
```
DanielVF: DAPP is currently only setting commissionPerUnit, not commission.

DanielVF:
Okay @franck, this one should be ready for a merge.

infra/discovery/devops/es-cli.js
```diff
@@ -164,7 +166,8 @@ function printInteractiveUsage() {
 3 create alias
 4 delete alias
 5 delete index
-6 reindex
+6 reindex (copy to new index)
+7 rebuild index (copy to temp index, and copy back)
```
franckc: Do we actually need this ?

We use alias in prod to point the search queries to a particular index. 
Assuming we have
 - index "listing1"
 - alias "listing" pointing to listing1

We can with the existing es-cli functionality do:
 - re-index from "listing1" into new index "listing2"
 - update alias "listing" to point to "listing2"

infra/discovery/src/lib/scoring.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+  if (
+    listing.price &&
+    listing.price.amount &&
+    listing.price.currency &&
+    listing.price.currency.id
+  ) {
+    const cheapThreshold = cheapListingThresholds[listing.price.currency.id]
+    if (cheapThreshold) {
+      if (listing.price.amount <= cheapThreshold) {
```
franckc: I think listing.price.amount is returned as a string.
Does this comparison between a string and a float work in JS ?
Perhaps let's explicitly cast to float ?

infra/discovery/src/lib/scoring.js
```diff
@@ -0,0 +1,56 @@
+async function scoreListing(listing) {
```
franckc: Does this need to be async ? I only see synchronous calls in the implementation below ?

infra/discovery/test/scoring.test.js
```diff
@@ -0,0 +1,87 @@
+const chai = require('chai')
```
franckc: Thanks for writing those unit tests. You are a champion.

infra/discovery/src/lib/scoring.js
DanielVF: In the future this may make a DB call, or check some kind of external fraud service, etc...

infra/discovery/devops/es-cli.js
DanielVF: We don't need it in production, but it's handy for me when working locally with docker.

It is a near duplication, it just speeds up my particular use.

infra/discovery/src/lib/scoring.js
DanielVF: Yeah, this works, but an explicit cast will be better.

PR #2377, opened 2019-05-30
Update pre-populated gift card images

### LINKED ISSUE #2368: Gift card types and images
Popular gift cards like eBay and Steam are not supported by our custom gift card listing type.

Also, many of the images, which were taken from Cardpool.com, are blurry and pixelated. Some are missing (Netflix)

joshfraser:
This PR updates a handful of the default images that we use to prepopulate the gift cards category for some of the top retailers. These images still aren't great, but they're better than what we had. I also added in some popular cards like eBay, Steam, Lowes, Home Depot, United and Delta that were missing. I want to get this PR in so @franckc has the latest and greatest for the June Rewards campaign. Would appreciate if @wanderingstan can make sure I did this right. I pointed the `upload2Ipfs.sh` script at our various gateways and ran it so all the relevant IPFS files should be already pinned if I'm understanding this correctly.

Closes #2368


wanderingstan:
Looks good to me. I had to fix a few lint/prettier things is all. 

Added the new images to the spreadsheet, and there we can see all the images loading fine:
https://docs.google.com/spreadsheets/d/1EWfHUrIkOra4RcptlMRzRz86QFUgfWyfSr4y4LLUueg/edit#
gid=1041625112

PR #2383, opened 2019-05-31
Metrics provider and some other minor stuff

mikeshultz:
### Description:

The meat:

- configures mainnet to use alchemy
- adds a Web3 provider to track RPC call metrics(outputs to stdout)
- prevents unnecessary eth_blockNumber calls for block number polling (incidental)
- allows customConfig for all networks(as discussed with @nick before, and thought was done already)
- Additon to `DEVELOPMENT.md` about adding packages with lerna.
- Additon of server env vars to `@origin/graphql`'s README

Ref #2346

### Checklist:


mikeshultz:
Seems like this may have broken something with Marketplace tests.  Maybe hold off on merging this.

mikeshultz:
Merged in master and things are passing again.  g2g

franckc:
In a meeting... will review shortly

packages/graphql/src/contracts.js
```diff
@@ -62,16 +63,23 @@ export function newBlock(blockHeaders) {
...
       web3.eth.getBlockNumber().then(block => {
         if (block > lastBlock) {
           web3.eth.getBlock(block).then(newBlock)
         }
+        inProgress = false
       })
     }, 5000)
   } catch (error) {
     console.log(`Polling for new blocks failed: ${error}`)
+    inProgress = false
```
franckc: Should we throw an exception here ? Since it basically means new blocks won't get polled at all.
Seems a critical error. Luckily should never happen since would only occur if for some reason setInterval fails ?

packages/graphql/src/contracts.js
```diff
@@ -62,16 +63,23 @@ export function newBlock(blockHeaders) {
...
 
 function pollForBlocks() {
+  let inProgress = false
   try {
     blockInterval = setInterval(() => {
+      if (inProgress) {
+        return
+      }
+      inProgress = true
       web3.eth.getBlockNumber().then(block => {
```
franckc: Should we have a .catch statement hat logs an error and sets inProgress to false ?
To handle the case where a call to web3.eth.getBlockNumber fails.

packages/graphql/src/contracts.js
```diff
@@ -127,6 +133,14 @@ export function setNetwork(net, customConfig) {
   clearInterval(blockInterval)
 
   web3 = applyWeb3Hack(new Web3(config.provider))
+
+  if (config.useMetricsProvider) {
+    addMetricsProvider(web3, {
+      echoEvery: 100,
```
franckc: nit: add units as a comment. Is that 100sec ? Same for breakdonwEvery.

packages/graphql/src/utils/metricsProvider.js
```diff
@@ -0,0 +1,142 @@
...
+ * Convert a standard provider to a web3-provider-engine subprovider
+ * @param provider {object} - A Web3 provider
+ * @returns {object} - A web3-provider-engine subprovider
+ */
+function convertWeb3Provider(provider) {
+  if (!provider) {
+    throw new Error('provider not provided')
+  }
+
+  if (provider.host && provider.host.indexOf('http') === 0) {
```
franckc: nit: could user startsWith rather than indexOf

packages/graphql/src/contracts.js
mikeshultz: hm, I hadn't considered this when I was looking at it.  Does `setInterval()` die forever on an error?  What if we moved the try/catch into the setInterval function argument?

packages/graphql/src/contracts.js
mikeshultz: Figured this was covered inside the MetricsProvider source, but can't hurt to add clarifications where they're set.  Will do.

PR #2384, opened 2019-05-31
Adds WEB3_PROVIDER_URL env var

mikeshultz:
### Description:

Adds the ability to specify an explicit web3 JSON-RPC provider URL with `WEB3_PROVIDER_URL`

This an okay location to do it?  Wanted to keep the config files clean.  Could probably use `customConfig` for `setNetwork()` as well and do it just for graphql server?

### Checklist:

### DIFF
--- a/packages/graphql/src/contracts.js
+++ b/packages/graphql/src/contracts.js
@@ -138,7 +138,10 @@ export function setNetwork(net, customConfig) {
   }
   clearInterval(blockInterval)
 
-  web3 = applyWeb3Hack(new Web3(config.provider))
+  const provider = process.env.WEB3_PROVIDER_URL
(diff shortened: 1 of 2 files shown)


packages/graphql/src/contracts.js
```diff
@@ -138,7 +138,10 @@ export function setNetwork(net, customConfig) {
   }
   clearInterval(blockInterval)
 
-  web3 = applyWeb3Hack(new Web3(config.provider))
+  const provider = process.env.WEB3_PROVIDER_URL
```
franckc: I think we use PROVIDER_URL for other services. So let's be consistent. plus it is what got added in README above :)

PR #2386, opened 2019-06-03
Normalize phone number before storing to DB

### LINKED ISSUE #2243: Phone numbers should be normalized before they are stored
We're currently storing international phone numbers in a variety of different formats. It would be better if we normalized these before saving them to the database. Right now, the same number can be saved 3 different ways:

- 1 5558675309
- 1 15558675309
- 1 +15558675309

Screen Shot 2019-05-08 at 12 05 23 PM

shahthepro:
Closes #2243 

At present, I'm parsing and converting the number just before generating the attestation. I have skipped doing that before generating a code. Should that be done?

Also, tried the [awesome-phonenumber](https://www.npmjs.com/package/awesome-phonenumber) package. But it seems heavier than the [libphonenumber-js](https://www.npmjs.com/package/libphonenumber-js) package.


nick:
Looks good. Just needs a few tests fixing...

infra/bridge/src/controllers/phone.js
```diff
@@ -130,7 +131,15 @@ router.post('/verify', phoneVerifyCode, async (req, res) => {
...
   }
-  const attestationValue = `${req.body.country_calling_code} ${req.body.phone}`
+  let attestationValue
+
+  try {
+    attestationValue = parsePhoneNumberFromString(
+      `+${req.body.country_calling_code} ${req.body.phone}`
+    ).number
+  } catch {
+    attestationValue = `${req.body.country_calling_code} ${req.body.phone}`
```
franckc: nit: log an error here so that we notice if the parsing fails.

shahthepro: Done.

PR #2387, opened 2019-06-03
Run tests with performance mode

nick:
Run puppeteer tests with Relayer, Proxies and Performance Mode


packages/graphql/src/configs/test.js
```diff
@@ -14,8 +14,8 @@ const config = {
   ipfsGateway: `http://${HOST}:8080`,
   ipfsRPC: `http://${HOST}:5002`,
   bridge: 'https://bridge.dev.originprotocol.com',
-  performanceMode: false,
-  graphql: `http://${HOST}:4002`,
+  performanceMode: localStorageHas('performanceMode'),
+  graphql: `http://${HOST}:4007`,
```
franckc: Why changing from 4002 to 4007 ? Should graphql/src/config/localhost.js be updated as well to keep the port consistent everywhere ?

nick: Something else was running on 4002 though couldn't work out what. Possibly IPFS. Will change for localhost too...

PR #2389, opened 2019-06-03
Fallback if discovery server returns error

nick:
Fallback if discovery server returns error

### DIFF
--- a/packages/graphql/src/resolvers/marketplace/listings.js
+++ b/packages/graphql/src/resolvers/marketplace/listings.js
@@ -142,13 +142,19 @@ export default async function listings(
   }
 
   let ids = [],
-    totalCount = 0
+    totalCount = 0,
+    discoveryError = false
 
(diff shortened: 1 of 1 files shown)


packages/graphql/src/resolvers/marketplace/listings.js
```diff
@@ -142,13 +142,19 @@ export default async function listings(
...
-    const discoveryResult = await searchIds(search, filters)
-    ids = discoveryResult.ids
-    totalCount = ids.length
-  } else {
+    try {
+      const discoveryResult = await searchIds(search, filters)
+      ids = discoveryResult.ids
+      totalCount = ids.length
+    } catch (e) {
+      discoveryError = true
```
franckc: Should we log an error in the console so that we know it failed ?

nick: Sure... just added that in

PR #2406, opened 2019-06-04
Add new attestations

shahthepro:
Adds new attestations

### Checklist:
- [x] Add KaKao Attestation
- [x] Add WeChat Attestation
- [x] Add GitHub Attestation
- [x] Add LinkedIn Attestation
- [x] Add Icons for the all attestations


franckc:
You are a machine, @shahthepro ! Cranking on so many features. That's awesome.

Quick question regarding those attestations, if the data is available (it may not be), should we consider recording the id of the user ? It's been really helpful for fraud purposes to do that for some of the old attestations... CC @micahalcorn for product input.

shahthepro:
@franckc It’s better if we record the unique id of the user instead of changeable fields like email and username.

micahalcorn:
I think the most important thing from the product perspective is that, where possible, we record an attribute that will allow us to link to a public profile. That has been problematic with LinkedIn for the Chai product. The next best thing would be to record a recognizable name that we can display. But as you say, we should definitely record the primary key that would allow us to identify a user despite a name change or something.

shahthepro:
> for those new attestations, as part of the OAuth flow, do we get an ID ? If yes, let's consider recording it in the DB (similarly to what we currently do for Airbnb and Twitter) rather than just the verified boolean

Yes, we do get an unique ID as well as basic profile details. Will update the code to store that ID too.

> I think the most important thing from the product perspective is that, where possible, we record an attribute that will allow us to link to a public profile. That has been problematic with LinkedIn for the Chai product. The next best thing would be to record a recognizable name that we can display.

To be clear, we need to store the profile URL (or a paramter to construct the profile URL), unique id and a public display name/handle of the account. Right?

dapps/marketplace/src/queries/Identity.js
```diff
@@ -26,6 +26,10 @@ export default gql`
           phoneVerified
           emailVerified
           websiteVerified
+          kakaoVerified
+          githubVerified
+          linkedinVerified
+          wechatVerified
```
nick: perhaps we should have a single field 'verified' with a parameter for type?

dapps/marketplace/src/queries/WechatAuthUrl.js
```diff
@@ -0,0 +1,9 @@
+import gql from 'graphql-tag'
+
+export default gql`
```
nick: Should we have a generalized AuthUrl mutation with a param for type?

nick: maybe call it attestationAuthUrl?

dapps/marketplace/src/pages/identity/OAuthAttestation.js
```diff
@@ -213,7 +247,29 @@ class GoogleAttestation extends Component {
   }
 }
 
-export default withRouter(GoogleAttestation)
+export default withIsMobile(withRouter(OAuthAttestation))
 
-require('react-styl')(`
-`)
+// require('react-styl')(`
```
nick: We don't need this anymore? Can just delete it if not

dapps/marketplace/src/queries/Identity.js
shahthepro: What if we want to get all/multiple attestation status?

dapps/marketplace/src/queries/WechatAuthUrl.js
shahthepro: This is file is not in use anymore. I missed to delete it. We have a `getAuthUrl` query with a `provider` param

dapps/marketplace/src/pages/identity/OAuthAttestation.js
shahthepro: Removed it.

dapps/marketplace/src/queries/Identity.js
nick: We could use an [alias](https://graphql.org/learn/queries/#aliases), something like:
```
kakaoVerified: providerVerified(provider: "kakao")
githubVerified: providerVerified(provider: "github")
```

Just means we don't have to add more fields to the resolver every time

dapps/marketplace/src/queries/WechatAuthUrl.js
shahthepro: https://github.com/OriginProtocol/origin/blob/135441f649204f176dff32398c8ef8c413552005/dapps/marketplace/src/queries/GetAuthUrl.js#L3-L9

dapps/marketplace/src/queries/Identity.js
nick: Or we could have a new type `verifiedProviders` that returns an array of IDs

nick: I think the latter would be better as we could return other info about the attestation in the same result:

```
{
  verifiedProviders {
    id
    properties {
      type
      value
    }
  }
}
```

returns:

```
{
  verifiedProviders: [
    {
      id: 'facebook',
      properties: [
        { type: 'created', value: 'Jan 1, 2016' },
        { type: 'friends', value: '240' },
      ]
    },
    {
      id: 'twitter',
      properties: [
        { type: 'created', value: 'Feb 4, 2017' },
        { type: 'followers', value: '600' },
      ]
    },
    {
      id: 'website',
      properties: [
        { type: 'url', value: 'https://poulden.com' },
        { type: 'created', value: 'Mar 7, 2011' },
      ]
    }
  ]
}
```

Maybe we can do this in a separate PR though.

shahthepro: Sure. Will open another PR for that.

PR #2407, opened 2019-06-04
Meta tx fixes

nick:
- Attestation valid on either proxy or owner. 
- GraphQL server config override for proxies.

### DIFF
--- a/packages/graphql/src/resolvers/IdentityEvents.js
+++ b/packages/graphql/src/resolvers/IdentityEvents.js
@@ -6,6 +6,7 @@ import get from 'lodash/get'
 import contracts from '../contracts'
 import { getIdsForPage, getConnection } from './_pagination'
 import validateAttestation from '../utils/validateAttestation'
+import { proxyOwner } from '../utils/proxy'
 
 const websiteAttestationEnabled =
   process.env.ENABLE_WEBSITE_ATTESTATION === 'true'
@@ -79,9 +80,15 @@ export function identity({ id, ipfsHash }) {
     if (!contracts.identityEvents.options.address || !id) {
       return null
     }
+    let accounts = id
     if (!ipfsHash) {
+      const owner = await proxyOwner(id)
+      if (owner) {
+        accounts = [id, owner]
+      }
+
       const events = await contracts.identityEvents.eventCache.getEvents({
-        account: id
+        account: accounts
       })
       events.forEach(event => {
         if (blockNumber < event.blockNumber) {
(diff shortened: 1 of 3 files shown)


packages/graphql/src/server.js
```diff
@@ -15,7 +15,8 @@ try {
 
 setNetwork(process.env.NETWORK || 'test', {
   performanceMode: false,
-  useMetricsProvider: process.env.USE_METRICS_PROVIDER === 'true'
+  useMetricsProvider: process.env.USE_METRICS_PROVIDER === 'true',
+  proxyAccountsEnabled: process.env.PROXY_ACCOUNTS_ENABLED === 'true'
```
franckc: I don't see that used anywhere so I assume it is for future ?

packages/graphql/src/resolvers/IdentityEvents.js
```diff
@@ -79,9 +80,15 @@ export function identity({ id, ipfsHash }) {
...
+    let accounts = id
     if (!ipfsHash) {
+      const owner = await proxyOwner(id)
+      if (owner) {
+        accounts = [id, owner]
+      }
+
       const events = await contracts.identityEvents.eventCache.getEvents({
-        account: id
+        account: accounts
```
franckc: I haven't paid close attention to the EventCache interface. I assume it expects addresses checksummed and both id and owner are indeed checksummed ?

packages/graphql/src/resolvers/IdentityEvents.js
```diff
@@ -114,7 +121,7 @@ export function identity({ id, ipfsHash }) {
         'avatarUrl',
         'description'
       ]),
-      ...getAttestations(id, data.attestations || []),
+      ...getAttestations(accounts, data.attestations || []),
```
franckc: Does getAttestation need to be updated to de-dupe attestations that are identical between proxy and owner ?

packages/graphql/src/server.js
nick: Right now proxies are enabled with a localStorage var but we need the GraphQL server to always have it on for performance mode

packages/graphql/src/resolvers/IdentityEvents.js
nick: Yep they are

packages/graphql/src/resolvers/IdentityEvents.js
nick: It should only look at the latest IPFS hash which shouldn't contain any duplicates

packages/graphql/src/server.js
franckc: Got it. Sg.

packages/graphql/src/resolvers/IdentityEvents.js
franckc: ah right. perfect.

packages/graphql/src/resolvers/IdentityEvents.js
franckc: sweet

Issue #2183, opened 2019-05-01
Listing Image Gallery Navigation & Expansion


micahalcorn:
Moved from #1727 🚚

The following changes should be made to the image viewing experience on the listing detail page:

- [ ] Navigation arrows on the main image (shown [here](https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5c799f8c5c1db5628cf25150))
- [ ] ~Closeable lightbox expansion of main image on click (shown [here](https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5c799f877890375205bfc49c))~    **moved to https://github.com/OriginProtocol/origin/issues/2412**

ldu2:
I would like to take a look at this issue. But since I am not invited on zeplin, is there any other way I can see the demo? Or if it's possible, can I get invited on zeplin?

micahalcorn:
@ldu2 sure, are you on [our Discord channel](https://www.originprotocol.com/discord)? If so, pm me your email address or you can post it here and I'll delete it.

ldu2:
Message sent on discord, but the email is the same one as the one you see in my profile here.

ldu2:
So I added the arrows in and they work fine. But I am still kind of confused about the second part "Closeable lightbox expansion". Is it like if you click on certain section and it zooms in on the picture? Or is it like a magnifier lens which shows a zoom in section while moving cursor on the image?

micahalcorn:
Hi @ldu2 😬 Are you still working on this, by chance? We're going to start working on a redesign, and it will involve prioritizing a swipe/navigation feature similar to the one described here. Let us know if you have anything that you want to commit before we move forward.

ldu2:
I finished the first feature in the list and that's the part I want to commit. I prefer creating a pull request since it helps me to keep track of different repos in just one account. I will create that right away.

micahalcorn:
Done per #2416 🙁

PR #2416, opened 2019-06-05
Ldu2/listing style change

ldu2:
Regarding ticket Listing Image Gallery Navigation & Expansion #2183,

Navigation arrows on the main image feature:
- Added left and right arrow images at dapps/marketplace/public/images/
- Add button functions and corresponding styles to switch between preview pictures
image


micahalcorn:
Thanks @ldu2 🙂 Can you put a condition in there to hide the arrows if they aren't needed? So we shouldn't have any arrows if there is only one picture. And probably no left arrow if the first picture is the main one, for example.

ldu2:
Thanks. That's absolutely right. I have fixed that. But I am not sure what I did wrong about the 
"ci/circleci: Setup — Your tests failed on CircleCI" ? Is there any addition set up needed?

micahalcorn:
Hey @ldu2 sorry that this got away from me. 💔 Our move to the redesigned DApp is happening faster than I expected, and we're already rolling out a different UX for this. I would ask for you to submit a branch directly instead of forking, but it probably wouldn't get merged at this point anyway. Let me know if there's something else that you would like to work on. Otherwise, thanks again for tackling this one even though it won't end up getting shipped.

PR #2417, opened 2019-06-05
Proxy and Factory on prod.

nick:
- Deploy Proxy implementation and Factory to mainnet
- Add Relayer and Proxy toggles to Settings page


dapps/marketplace/src/pages/settings/Settings.js
```diff
@@ -304,6 +307,72 @@ class Settings extends Component {
                           />
                         </div>
                       </div>
+                      <div className="form-group row d-flex">
+                        <div className="col-sm">
+                          <label htmlFor="indexing">
+                            <fbt desc="settings.proxyAccountsLabel">
```
franckc: Please run npm translate before landing so that those new strings get picked up by crowdin.

PR #2427, opened 2019-06-06
[Listener] Process ProxyCreation events

franckc:
### Description:
 - Modified listener to watch for ProxyCreation event emitted by the ProxyFactory contract
 - Upon receiving an event, the listener looks up the owner (e.g. wallet) of the proxy and stores the mapping proxy address <-> wallet address in the DB
 
This proxy table can then be used for analytics purposes, but also for Origin Rewards, in order to credit to the wallet action recorded under the proxy address. For example, a listing sold or identity publish action performed via the proxy will get attributed to the wallet that owns that proxy address.

### Checklist:


infra/discovery/src/listener/listener.js
```diff
@@ -115,8 +117,9 @@ async function main() {
   const context = await new Context().init(config, errorCounter)
 
   const contracts = {
-    IdentityEvents: contractsContext.identityEvents,
-    Marketplace: contractsContext.marketplace
+    //IdentityEvents: contractsContext.identityEvents,
```
nick: Need to uncomment these again?

packages/graphql/src/contracts.js
```diff
@@ -182,6 +182,8 @@ export function setNetwork(net, customConfig) {
 
   setMarketplace(config.V00_Marketplace, config.V00_Marketplace_Epoch)
   setIdentityEvents(config.IdentityEvents, config.IdentityEvents_Epoch)
+  setProxyFactory(config.ProxyFactory, config.ProxyFactory_Epoch)
+  // TBD: do we need those 2 lines ?
```
nick: it's used in utils/proxy.js but we could refer the context.proxyFactory there instead now we're setting it here

franckc: Cleaned up. Sorry I had no pushed my latest changes.

infra/discovery/src/listener/listener.js
franckc: Done.

PR #2430, opened 2019-06-06
[Growth] Query events from owner and proxies

franckc:
### Description:

With the launch of meta-txn, the listener back-end will be recording GrowthEvents (e.g. Identity updates, Listing Sold/Purchased, etc...) under the address of either the wallet (in case user paid for gas) or the proxy (in case relayer paid for gas).

This PR updates the rules logic so that it loads events recorder under the wallet and any proxies owned by it.

The DApp always queries the growth back-end using the wallet address regardless of the wallet having aproxy or not. As a result, the user should see rewards for actions performed by either wallet or proxies.

Note: the fraud and payout pipeline needs to be updated to take into account proxies. I'll handle that as a separate PR.

### Checklist:

### DIFF
--- a/infra/growth/src/resources/rules.js
+++ b/infra/growth/src/resources/rules.js
@@ -1,6 +1,9 @@
 const Sequelize = require('sequelize')
 
-const db = require('../models')
+const _growthModels = require('../models')
+const _discoveryModels = require('@origin/discovery/src/models')
+const db = { ..._growthModels, ..._discoveryModels }
+
 const {
   GrowthEventTypes,
   GrowthEventStatuses,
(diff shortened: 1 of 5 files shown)


franckc:
CC @nick as a FYI

infra/growth/src/resources/rules.js
```diff
@@ -85,9 +90,14 @@ class CampaignRules {
...
-    const whereClause = {
-      ethAddress: ethAddress.toLowerCase()
-    }
+
+    // Load any proxy associated with the wallet address.
+    const ownerAddress = ethAddress.toLowerCase()
+    const proxies = await db.Proxy.findAll({ where: { ownerAddress } })
+
+    // Query events from wallet and proxy(ies).
+    const addresses = [ownerAddress, ...proxies.map(proxy => proxy.address)]
```
sparrowDom: Should we call `toLowerCase()` on proxy addresses as well?

franckc: The proxy table already stores addresses in lower case so that should not be necessary.

franckc:
Thanks @sparrowDom for the thorough review and quick turnaround ! Going to merge this to master and will deploy today...

PR #2431, opened 2019-06-06
JSON-RPC Throttling and retries

mikeshultz:
### Description:

addition of ThrottleRPCProvider to replace RPCProvider which handles rate limiting and retries.

This kind of suppresses a lot of errors that normally get tracked by MetricsProvider, since it adds retries.  At some point it might be good to refactor to allow handled errors to be tracked as well so we get a full picture of how things are working with the provider(e.g. Alchemy).

There's some minor test changes in here to deal with web3-provider-engine's idiosyncrasies.

NOTE: This is currently not enabled for browsers.

### Checklist:


packages/graphql/src/mutations/_txHelper.js
```diff
@@ -55,6 +55,22 @@ function useRelayer({ mutation, value }) {
...
+  if (typeof web3Inst.currentProvider.sendAsync !== 'undefined') {
+    return await web3Inst.currentProvider.sendAsync(
+      { method: 'evm_mine' },
+      () => {}
+    )
+  } else {
+    return web3Inst.currentProvider.send({ method: 'evm_mine' }, () => {})
+  }
+}
+
```
nick: Do we need the async / await here? Or could we do something like this:
```js
function mineBlock(web3Inst) {
  const hasAsync = typeof web3Inst.currentProvider.sendAsync !== 'undefined'
  const sendMethod = hasAsync ? 'sendAsync' : 'send'
  return web3Inst.currentProvider[sendMethod]({ method: 'evm_mine' }, () => {})
}
```

mikeshultz: yeah, that should work as well.  Doesn't look like anything is dependent on it returning anything.  Will have that change up a little later this morning.

nick:
Would like @franckc to also take a look before we merge it

packages/graphql/src/providers/ThrottleRPCProvider.js
```diff
@@ -0,0 +1,140 @@
...
+  if (error) {
+    throw new JsonRpcError.InternalError(error)
+  }
+
+  // check for error code
+  switch (response.status) {
+    case 405:
+      throw new JsonRpcError.MethodNotFound()
+    case 504: // Gateway timeout
+      return (function() {
```
franckc: My JS fu is terribly weak. Why do we need to use a function here ?

packages/graphql/src/providers/ThrottleRPCProvider.js
```diff
@@ -0,0 +1,140 @@
...
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify(payload)
+    })
+  } catch (err) {
+    error = err
+  }
+
+  if (error) {
+    throw new JsonRpcError.InternalError(error)
```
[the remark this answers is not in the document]
mikeshultz: Good point.  It's an artifact from the original RPCProvider code.  Will fix.

packages/graphql/src/providers/ThrottleRPCProvider.js
```diff
@@ -0,0 +1,140 @@
...
+  try {
+    data = await response.json()
+  } catch (err) {
+    console.log('errror parser')
+    console.error(err.stack)
+    throw new JsonRpcError.InternalError(err)
+  }
+  if (data.error) {
+    console.log('data.error')
+    throw data.error
```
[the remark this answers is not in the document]
mikeshultz: Also an artifact.  Good catch, this isn't an instance of `Error`.

packages/graphql/src/providers/ThrottleRPCProvider.js
```diff
@@ -0,0 +1,140 @@
...
+        throw new JsonRpcError.InternalError(response.text)
+      }
+  }
+
+  // parse response
+  let data
+  try {
+    data = await response.json()
+  } catch (err) {
+    console.log('errror parser')
```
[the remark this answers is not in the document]
mikeshultz: Agreed, it would be good to standardize in the future.  Right now, `@origin/graphql` doesn't have any logging packages, which is why `console` was used.

packages/graphql/src/providers/ThrottleRPCProvider.js
mikeshultz: More artifacts from the original source...  It was originally returning with a call to `end()` with this error, but we want to throw here in this new configuration.  I'll rip this stuff out.  Thanks

franckc:
Let's get this in...

PR #2447, opened 2019-06-09
Relayer account abstraction

mikeshultz:
### Description:

Account abstraction for the relayer.

This adds a component I named Purse because I'm not good at naming things.  This component will create HD derived accounts from a master account created from the `FORWARDER_MNEMONIC`. `FORWARDER_PK` has been depreciated.  It can create an unlimited amount of accounts, but it's currently fixed at either `FORWARDER_ACCOUNTS` or `3` by default.

I would love to add some level of auto-scaling to this, but @franckc seemed a little hesitant to that idea.  Perhaps that's a feature to add down the road.

Purse can(and I highly recommend) auto-fund the children accounts with funds deposited in the master.  It includes functionality that can drain the children back to the master as well, though it's not used by anything(could create a script in minutes, though).

Nonce tracking is done internally, persisted to redis, and will fallback to the chain if necessary.

This includes a bunch of tests, including some basic tests for Relayer.

Ideally the master account should be funded before prod deploy, however that's not necessary.  The service will idle, and probably timeout requests until the master account is funded.

I would specifically like to point out extra scrutiny:

- The weird false-thread I use for the process loop in the constructor.  I've used them successfully in the past, but would be open to alternatives
-  What should happen if the gas price rises to an unacceptable level?  What is that level?
- `autofundChildren` is enabled for the relayer.  You cool with that, @franckc?  It works fiarly well, and the ranges can adjusted if we'd rather have smaller funding transactions, or less gas usage.
- ~~I could probably use a pointer on how to setup the redis connection for the clusters.  Is there a proxy similar to the Postgres configurations?  Are there other services I can use for reference?  @tomlinton?~~

#### Env Var Changes

- `REDIS_URL` needs to be set for the relayer service for deploy
- `FORWARDER_PK` has been depreciated
- `FORWARDER_MNEMONIC` should be set and ideally funded
- `FORWARDER_ACCOUNTS` should be set IF the default of `3` isn't ideal

### Checklist:


dapps/marketplace/test/setup.js
```diff
@@ -3,20 +3,58 @@
...
   webpackProcess = shutdownServices.extrasResult.webpackProcess
+
+  /**
+   * Need to fund the master account for the Relayer Purse
+   */
+  const web3 = new Web3('http://localhost:8545')
+  const netId = await web3.eth.net.getId()
+
+  if (parseInt(netId) === 999) {
+    if (!process.env.FORWARDER_MNEMONIC) {
```
nick: Can we default this if `process.env.NODE_ENV === 'test'`? Then we can skip on the extra env var in the package.json scripts

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,682 @@
...
+
+    /**
+     * Finally, check with the JSON-RPC provider.  Chance of some pendings screwing this up if we
+     * get this far.  I'm at least sure that eth_pendingTransactions, is completely  not supported
+     * by Alchemy.  Doubtful from Infura as well.  And it's a poor method anyway with no arguments.
+     *
+     * EDGE CASE WARNING
+     * -----------------
+     *
+     * The only time this should be an issue is if the service is magically restarted, Redis is
```
nick: Yea was going to ask what happens if the service is restarted, for example when we do a new deployment and there are still transactions pending

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,682 @@
...
+    this.children = []
+    this.accounts = {}
+    // Keep track of transactions that are waiting to be mined
+    this.pendingTransactions = {}
+    // Counter for tx rebroadcasts
+    this.rebroadcastCounters = {}
+    // Callbacks for when receipts show up
+    this.receiptCallbacks = {}
+
+    this.rclient = this._setupRedis(redisHost)
```
nick: Would it be a lot of work to have redis as optional so it's still possible to run the relayer locally without any other dependencies?

dapps/marketplace/test/setup.js
```diff
@@ -3,20 +3,58 @@
...
+      throw new Error('FORWARDER_MNEMONIC missing, setup failed!')
+    }
+
+    const masterAddress = await mnemonicToAddress(process.env.FORWARDER_MNEMONIC)
```
nick: Wondering if we can move this code to the relayer and perhaps have an startup option 'fund=true' that can be called from @origin/services (similar to ganache inMemory option)

infra/relayer/src/purse.js
nick: (just reading through... maybe it's already optional)

infra/relayer/src/purse.js
mikeshultz: To be clear, the edge case I mention here isn't triggered just by a service restart.  A regular old service restart shouldn't have any negative effect.

infra/relayer/src/purse.js
mikeshultz: Yeah, it's optional.  For example, the marketplace tests aren't configured use Redis.

dapps/marketplace/test/setup.js
mikeshultz: That might be a good idea.  How was the previous `FUNDER_PK` account funded before?  I didn't see anything.

nick: I think I just used one of the default node accounts

mikeshultz: I agree though that it would be useful outside of testing(like local dev envs) and should probably go somewhere else but I don't know if this is the best option.  Could be an env var consumed by `app.js`, but then it has to give Relayer the option, which then should probably pass it to Purse, which then could fun during `init()`... for something that would only work in an environment with unlocked default accounts(ganache).  

What about including it in @origin/services whenever Relayer is enabled?  Would have to add web3 as a dependency there, but web3 is basically everywhere anyway.

nick: How about if we just do it as part of the `populate` script?

mikeshultz: That's a good spot.  Moved the funding to `populate.js`, let me know what you think.

tomlinton:
This is really cool.

On that redis question, the integration is simple because redis is run in the cluster. PostgreSQL is a bit more complex because we use the managed Google Cloud offering, rather than running our own in the cluster. That is why we need to use that `cloudsql-proxy`. It basically just provides connectivity between Google Cloud Postgres and our Kubernetes cluster.

All you need to do to make it work is give it the correct URI, e.g. `redis://dev-redis-master.dev.svc.cluster.local:6379` for dev. Find/replace dev for staging and prod. 😀

mikeshultz:
Thanks, @tomlinton!  

Would you recommend just setting the redis connection info via an env var?  Also, should I use a different redis db number(e.g. `1`)?  I made sure to prefix all the keys for this to prevent conflicts, but perhaps for safety or `keys *` performance it might be better to use a db separate from whatever else uses it?

tomlinton:
Yep `REDIS_URL` is what we've used elsewhere. Good idea re: db. I like doing it that way as it often makes debugging easier if you are doing `keys:*` or similar, so you can quickly see if keys are getting created without all the other services keys turning up.

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,696 @@
...
+   * Send a transaction from an available sender account.
+   *
+   * NOTE: This does not wait for the transaction to be mined!
+   *
+   * @param tx {object} - The transaction object, sans `from` and `nonce`
+   * @param onReceipt {function} - An optional callback to call when a receipt is found
+   * @returns {string} The transaction hash of the sent transaction
+   */
+  async sendTx(tx, onReceipt) {
+    const address = await this.getAvailableAccount()
```
franckc: If the relayer receives N requests in parallel, I don't know much about the node event scheduler, but isn't it possible that all N requests would then resolve to the same address ?

I guess my underlying question is do we need to have some sort of locking mechanism and load balancing of accounts being picked in getAvailableAccount ?

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,696 @@
...
+      this.receiptCallbacks[txHash] = onReceipt
+    }
+
+    // In case it needs to be rebroadcast
+    this.pendingTransactions[txHash] = rawTx
+
+    // blast it
+    await sendRawTransaction(this.web3, rawTx)
+
+    await this.incrementTxCount(address)
```
franckc: I don't see anywhere that we decrement txCount. What happens if the transaction fails ? For example if we have a bug and submit an invalid tx ? Wouldn't that cause subsequent transactions to get stuck because there would be a hole in the nonce sequence ?

I wonder if it's worth subscribing to errors and decrementing the nonce... ?

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,696 @@
...
+async function tick(wait = 1000) {
+  return new Promise(resolve => setTimeout(() => resolve(true), wait))
+}
+
+class Account {
+  constructor({ txCount, pendingCount = 0, wallet, funded }) {
+    this.txCount = txCount
+    this.pendingCount = pendingCount // needed?
+    this.wallet = wallet
+    this.funded = funded
```
franckc: I'm not super clear on the role of the funded flag in the constructor here as well as in the Account objects. It does not seem to be used for any logic ? Shall it be removed then ?

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,696 @@
...
+     * wiped or unavailable, all while there are transactions still in the pool waiting to be
+     * mined.
+     *
+     * Also, tiny chance of transaction replacement here if the gas price goes up during the above
+     * edge case.  That would be... like a dropped transaction in the client.  Perhaps not a deal
+     * breaker, but maybe some thought should be put into that in the future as volume goes up. The
+     * bigger the load, the likelier this becomes.
+     */
+    const w3txCount = parseInt(await this.web3.eth.getTransactionCount(address))
+    if (txCount < w3txCount) txCount = w3txCount
```
franckc: I'd log a warning that w3count was used rather than Redis. This indicates something is fishy... like at some point perhaps Redis count was not updated...

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,696 @@
...
+    if (masterBalance.lt(maxFunding)) {
+      logger.warn(
+        `Master account ${masterAddress} does not have enough funding. Expected ${maxFunding}, has ${masterBalance}`
+      )
+      return
+    }
+
+    // If the equal split between all children is less than the base fund value
+    // use it instead of trying to overdraw
+    const eqSplit = masterBalance.div(numberToBN(this.children.length))
```
franckc: Don't we also need to account for cost of gas to pay for the funding transactions to children ?

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,696 @@
...
+      `Sending ${fundingValue.toString()} (total cost: ${txCost}) from ${masterAddress} (bal: ${masterBalance})`
+    )
+
+    const unsigned = {
+      nonce: await this.txCount(masterAddress),
+      from: masterAddress,
+      to: address,
+      value: fundingValue.toString(),
+      gas: 21000,
+      gasPrice: await this.gasPrice()
```
franckc: why calling this.gasPrice() again whereas it was already called on line 497 ?

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,696 @@
...
+    const fundingValue = value ? value : BASE_FUND_VALUE
+    const gasPrice = await this.gasPrice()
+    const txCost = fundingValue.add(gasPrice.mul(new BN(21000)))
+    const masterPrivkey = this.masterWallet.getPrivateKey()
+    const masterAddress = this.masterWallet.getChecksumAddressString()
+    const masterBalance = stringToBN(
+      await this.web3.eth.getBalance(masterAddress)
+    )
+
+    if (masterBalance.lt(BASE_FUND_VALUE)) {
```
franckc: Should this test on txCost rather than BASE_FUND_VALUE ?

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,696 @@
...
+      const pendingHashes = Object.keys(this.pendingTransactions)
+      for (const txHash of pendingHashes) {
+        const receipt = await this.web3.eth.getTransactionReceipt(txHash)
+
+        if (!receipt) continue
+
+        logger.debug(`Transaction ${txHash} has been mined.`)
+
+        // Remove from pending if it exists(it should)
+        if (!receipt.status) {
```
franckc: Should this also check blockNumber is set ? I believe when I ran the June rewards payment we ran into this issue that Alchemy returns a receipt with no blockNumber before the tx is mined.

infra/relayer/src/purse.js
mikeshultz: This was kind of one of my concerns(as noted in the file TODO at the top).  It wasn't super clear to me if express would be handling parallel requests, which would then send those requests to Purse.  Seems likely, but I'm really not sure.  Since node doesn't support threads and all, unless there's multiple workers being stood up in the clusters(which would introduce even further issues with Purse), I'm thinking these requests should still be sequential in that there probably isn't a race condition here.  But my confidence in that isn't very high.  This might warrant further research.

infra/relayer/src/purse.js
mikeshultz: Fairly certain a failed tx is still counted in the nonce as it's still included in a block.  Dropped transactions however, are a little bit of a concern here as I've noted in a couple places in the source.  The re-broadcasting should in the best of times make this a non-issue.  But if there's a pending tx, the service is restarted(lost from memory), and a tx is dropped, that will cause a problem.

I'm just not sure if that edge case needs to be dealt with now or later before volume increases.  I'd probably defer to your opinion on that.

infra/relayer/src/purse.js
mikeshultz: Appears to be an artifact.  I was using it at one point, but no longer.  Will remove, thanks for catching that.

infra/relayer/src/purse.js
mikeshultz: Yes!  Thank you for catching that.

infra/relayer/src/purse.js
mikeshultz: This isn't necessarily fishy.  This was defensive, and it will be triggered at a very minimum on the initial process startup.  I'm adding the warning under the condition of `txCount > 0`.

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,696 @@
...
+} = require('./util')
+const logger = require('./logger')
+
+const REDIS_RETRY_TIMEOUT = 30000
+const REDIS_RETRY_DELAY = 500
+const DEFAULT_CHILDREN = 5
+const DEFAULT_MAX_PENDING_PER_ACCOUNT = 3
+const ZERO = new BN('0', 10)
+const BASE_FUND_VALUE = new BN('50000000000000000', 10) // 0.05 Ether
+const MIN_CHILD_BALANCE = new BN('1000000000000000', 10) // 0.001 Ether
```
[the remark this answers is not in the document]
mikeshultz: I think you're right.  Here's the cost in Ether of deploying a proxy at 10gwei:

    >>> (2000000 * 10e9) / 1e18
    0.02

We also don't want to have funding transactions occurring every time a tx is fired.  What do you think are the appropriate levels here?  Could be like 1E/0.5E, but I don't really want to make business assumptions here.  

That also leads into one of my TODOs that I hope you can answer: what should be the gasPrice limit for this service?  What is too expensive for Origin where the service should error out and refuse to send them?

infra/relayer/src/purse.js
mikeshultz: I've read a handful of articles and docs on the NodeJS event loop and I still can't seem to get a straight answer on this.  However, I think the solution is relatively simple and low-effort that I'm just going to go ahead and implement a simple lock that is only relevant between the `getAvailableAccount()` and `incrementTxCount()` calls.  

The worst that could happen is I spend 15 minutes to make Purse future-proof if we make any improvements to this Express server(e.g. [cluster](https://nodejs.org/api/cluster.html)).  At the best, we save ourselves some head-aches later down the road.

Will implement this later tonight or in the morning.  Thanks for raising this concern.

mikeshultz: Added a simple locking mechanism, and some gc in the process loop just in case...  Take a look when you get a chance.

mikeshultz:
Went ahead and added pending transaction persistence.  Now, if there's pending transactions and the service is restarted, the new instance will pickup the previous pending transactions and manage them accordingly.

infra/relayer/src/purse.js
mikeshultz: My latest commented change should take care of this edge case I mentioned.  I think this can be resolved?

infra/relayer/src/purse.js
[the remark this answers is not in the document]
franckc: Thanks for the data. How about these numbers to start with and we can adjust if needed later:
 - BASE_FUND_VALUE: 0.5 ETH
 - MIN_CHILD_BALANCE: 0.1 ETH

infra/relayer/src/purse.js
franckc: Based on my understanding the node event loop may switch to executing a new callback anytime there is some blocking IO. So in our case anytime there is a disk or network operation. So yes I'm pretty sure that if express receives concurrent requests, it would not process them sequentially but would switch from one to the other when an IO operation occur.
Makes sense to implement locking.

infra/relayer/src/purse.js
```diff
@@ -0,0 +1,802 @@
...
+              await this._fundChild(child)
+            }
+          }
+        }
+      }
+
+      /**
+       * Handle incoming receipts and remove pending transactions and adjust pending counts.
+       */
+      const pendingHashes = Object.keys(this.pendingTransactions)
```
franckc: The relayer(and therefore the DApp) is blocking on the receipt being ready.

This is not optimal in terms of overall latency for a transaction since we process all the transaction hashes serially and also may fund child accounts balance.

It would be more performant to create a new individual Promise at the end of the _sendTx method that wait for the receipt for the tx that was sent and invokes the callback.

I don't think it's going to be noticeable though until we have some transaction volume. It is more of a possible future performance optimization. Perhaps add as a comment so we remember ?

PR #2459, opened 2019-06-12
[Discovery] Listing moderation tool

DanielVF:
## Purpose

We want to allow human moderation of listings results/search served by our discovery service.  This is to promote good listings, lower poor quality listings, remove spam, and remove illegal content. In the future this tagging could provide learning data for more automated scoring.

This does not affect data on the blockchain itself, and the Origin Protocol at that level remains entirely unmoderated.

## Features

A list of moderator ETH addresses can be set in the `MODERATOR_ADDRESSES` env variable

Moderators use the admin dapp. Each listing page has a moderation button. If the moderator is not logged in, this button appears as a "Login" button. Logging with a signature creates an auth token that lasts 36 hours. 

Once logged in, the moderator can see the listing's current tag, and can click to change the tag.

These changes take place "instantly" in the database and in ElasticSearch. Queries will immediately begin using the new scores.

These scoring tags are stored in the database as well as ElasticSearch, and future reindexes continue to use the tags for a listing. We keep an audit log in the database of each tag change event, along with the Eth address that made the change.

Non-moderators cannot read the tags or listing score over GraphQl.

## Local testing

You need to uncomment the discovery service line in `packages/graphql/src/configs/localhost.js` to allow the admin dapp to use local listings.

The first two metamask accounts are set as moderators in the local docker environment.

## Deployment

You will need to set the MODERATOR_ADDRESSES env variable to a comma seperated (no space) list of moderator ETH addresses.

This code should not require a listener reindex or a ElasticSearch reindex. It does change the ElasticSearch mapping, but that is for future correctness, and is not required.


packages/contracts/build/contracts/IdentityEvents.json
```diff
@@ -876,7 +876,11 @@
     "name": "solc",
     "version": "0.4.25+commit.59dbf8f1.Emscripten.clang"
   },
-  "networks": {},
+  "networks": {
+    "999": {
+      "address": "0xeec918d74c746167564401103096D45BbD494B74"
```
nick: I think we can revert the changes to the contract build files as these addresses won't get used anywhere

dapps/admin/src/utils/discovery.js
```diff
@@ -0,0 +1,104 @@
...
+  const authToken = activeAuthToken()
+  return new Promise(function(resolve, reject) {
+    const headers = {
+      'Content-Type': 'application/json',
+      Accept: 'application/json'
+    }
+    if (authToken) {
+      headers['x-discovery-auth-token'] = authToken
+    }
+    fetch(getDiscovery(), {
```
nick: Ideally we'd have the discovery GraphQL server merged in the same way as Growth is, but this works for now 😊

packages/contracts/build/contracts/IdentityEvents.json
DanielVF: Ugh. Yes, this should not have been committed.

infra/discovery/src/apollo/app.js
```diff
@@ -36,10 +36,14 @@ app.use(bundle)
 const server = new ApolloServer({
   resolvers,
   typeDefs,
-  context: async () => {
+  context: async context => {
+    const headers = context.req.headers
+    if (headers['x-discovery-auth-token']) {
```
franckc: Please add this header to the [ingress config for discovery.](https://github.com/OriginProtocol/origin/blob/master/devops/kubernetes/charts/origin/templates/discovery.ingress.yaml) Also the cors configuration so that admin.originprotocol.com can query discovery.originprotocol.com.

Here is an example: https://github.com/OriginProtocol/origin/blob/master/devops/kubernetes/charts/origin/templates/growth.ingress.yaml

This could be done as a separate PR.

infra/discovery/src/apollo/resolvers.js
```diff
@@ -46,7 +72,107 @@ const resolvers = {
...
+    async listingSetScoreTags(_, data, context) {
+      const auth = await authenticate(context.discoveryAuthToken)
+      if (!auth) {
+        throw 'You are not logged in'
+      }
+      const { ethAddress } = auth
+
+      const { id, scoreTags } = data
+      await db.DiscoveryTagAction.create({
+        ethAddress,
```
franckc: nit: we have decided on the convention to alway store eth addresses lowercases in the DB. update to:
`
ethAddress: ethAddress.toLowerCase()
`

infra/discovery/src/apollo/resolvers.js
```diff
@@ -46,7 +72,107 @@ const resolvers = {
         nodeEnv: process.env.NODE_ENV ? process.env.NODE_ENV : 'undefined'
       }
     }
+  },
+  Mutation: {
+    async listingSetScoreTags(_, data, context) {
+      const auth = await authenticate(context.discoveryAuthToken)
+      if (!auth) {
+        throw 'You are not logged in'
```
franckc: nit: we follow the convention in our codebase to throw an Error object. So like:
throw new Error('You are not logged in')

infra/discovery/src/apollo/resolvers.js
```diff
@@ -46,7 +72,107 @@ const resolvers = {
...
+      if (usedCount > 0) {
+        throw 'Nonces may only be used one time'
+      }
+
+      const authToken = Web3.utils.randomHex(16)
+      const secondsToExpiration = 36 * 60 * 60
+      const now = new Date().getTime()
+      const expires = new Date(now + secondsToExpiration * 1000)
+
+      db.DiscoveryAccessToken.create({
```
franckc: missing an await here.

infra/discovery/migrations/20190612151515-index-discovery-access-token.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+module.exports = {
+  up: queryInterface => {
+    return queryInterface.addIndex('discovery_access_token', {
```
franckc: I think we also need another index on auth_token column.

infra/discovery/src/lib/search.js
```diff
@@ -84,6 +84,55 @@ class Listing {
...
+    const result = await client.get({
+      index: LISTINGS_INDEX,
+      type: LISTINGS_TYPE,
+      id: id
+    })
+    const listing = result._source
+    listing.scoreTags = scoreTags
+    const { scoreMultiplier } = await scoring.scoreListing(result._source)
+    listing.scoreMultiplier = scoreMultiplier
+    const body = {
```
franckc: nit: add some comments here or at top of function to explain we use painless script to re-index the document with a new score calculated based on scoreTags.
Is painless actually necessary rather than just re-indexing the doc we got from result._source ?

infra/discovery/src/lib/search.js
```diff
@@ -125,6 +177,13 @@ class Listing {
       term: { valid: false }
     })
 
+    // Never return any listings moderated as hidden
+    esQuery.bool.must_not.push({
+      terms: {
+        scoreTags: ['Hide', 'hide', 'Super Hide', 'super hide']
```
franckc: why do we need to handle different casing here ? under which scenario would the tag get lower cased ?

dapps/admin/src/pages/listings/_EditScoring.js
```diff
@@ -0,0 +1,114 @@
...
+import { query } from '../../utils/discovery'
+
+const ALL_TAGS = [
+  { tag: 'Super Featured', icon: 'clean' },
+  { tag: 'Featured', icon: 'star' },
+  { tag: 'High Quality', icon: 'thumbs-up' },
+  { tag: 'Normal', icon: 'circle' },
+  { tag: 'Low Quality', icon: 'thumbs-down' },
+  { tag: 'Hide', icon: 'ban-circle' },
+  { tag: 'Super Hide', icon: 'ban-circle' }
```
franckc: What's the difference again between Hide and Super Hide ?
"Super Hide" is not super explicit :)... should it be renamed "Delete" ?

infra/discovery/src/lib/search.js
```diff
@@ -299,6 +365,14 @@ class Listing {
     ])
     const listings = []
     searchResponse.hits.hits.forEach(hit => {
+      console.log(
```
franckc: Remove this debug statement or replace with logger.debug

infra/discovery/src/lib/search.js
DanielVF: Previously I tested this, and the lower case version seemed to be required. I've tested now and that is at least now no longer the case.  I've removed the lowercase names.

dapps/admin/src/pages/listings/_EditScoring.js
DanielVF: I've renamed it to "Delete".

infra/discovery/migrations/20190610140811-discovery_tag_action.js
```diff
@@ -0,0 +1,40 @@
...
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true
+      },
+      listing_id: {
+        type: Sequelize.STRING(32),
+        allowNull: false
+      },
+      eth_address: {
+        type: Sequelize.CHAR(42),
```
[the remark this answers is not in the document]
DanielVF: Yes, I based these of an old migration. Changed to use STRING.

infra/discovery/src/lib/search.js
DanielVF: Comment added. Painless isn't necessary, but it seems cleaner and reduces the size of the race condition window when a doc could be being updated simultaneously by multiple systems.

PR #2466, opened 2019-06-17
Mobile 2.0

tomlinton:
This is most of the work for the new mobile experience.

- New onboarding with meta tx support
- API to the Marketplace GraphQL layer so GraphQL queries can be run from mobile leveraging the Marketplace code
- Restructured nav so the Marketplace is always loaded so it can respond to GraphQL queries
- Refactored wallet component to be a HOC with global web3

There'll probably be a few more fixes to come, but we should merge this now to get the new interface strings up for translation.


mobile/src/hoc/withWeb3Accounts.js
```diff
@@ -0,0 +1,279 @@
...
+        privateKey: account.privateKey
+      })
+      return account
+    }
+
+    /* Add a new account
+     */
+    addAccount = account => {
+      this.props.addAccount(account)
+      global.web3.eth.accounts.wallet.add(account)
```
nick: should this kind of thing be happening in the action? Not a big deal but think business logic should generally live in there

mobile/src/screens/marketplace.js
```diff
@@ -108,16 +101,56 @@ class MarketplaceScreen extends Component {
...
       this.handleBridgeResponse(msgData, response)
+    } else if (msgData.targetFunc === 'signPersonalMessage') {
+      // Personal sign is for handling meta transaction requests
+      const decodedData = JSON.parse(
+        global.web3.utils.hexToUtf8(msgData.data.data)
+      )
+      const decodedTransaction = decodeTransaction(decodedData.txData)
+      // If the transaction validate the sha3 hash and sign that for the relayer
+      if (this.isValidMetaTransaction(decodedTransaction)) {
+        const dataToSign = global.web3.utils.soliditySha3(
```
nick: Again seems like this kind of logic should happen in either an action or a reducer but no big deal

mobile/src/hoc/withWeb3Accounts.js
tomlinton: Thanks, yea, there should probably be some simplifying here. It followed a bit of a weird evolution to get here.

mobile/src/components/address.js
```diff
@@ -1,39 +1,36 @@
...
-              { text: String(fbt('OK', 'Address.alertOkButton')) }
-            ])
-          }
-        }}
-      >
-        <Text style={style}>{truncateAddress(address, chars)}</Text>
-      </TouchableOpacity>
-    )
-  }
+const Address = ({ address, chars, label, style, onPress }) => {
```
nick: Hmm I think we can remove the brace on the end?

mobile/src/components/currency.js
```diff
@@ -1,52 +1,49 @@
...
+const Currency = ({
+  abbreviation,
+  balance,
+  imageSource,
+  labelColor,
+  name,
+  precision = 5,
+  onPress
+}) => {
+  /* eslint-disable-next-line no-extra-semi */
```
nick: not sure how this would work with the brace on line 16?

mobile/src/components/learn-card.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
+
+import React, { Component } from 'react'
+import { StyleSheet, Text, View } from 'react-native'
+import { fbt } from 'fbt-runtime'
+
+import OriginButton from 'components/origin-button'
+
+class LearnCard extends Component {
```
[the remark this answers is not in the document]
tomlinton: Refactored a bunch of these!

mobile/src/hoc/withWeb3Accounts.js
tomlinton: Moved a lot of the logic into redux and out of this HOC

tomlinton:
Think this is good to go now. I'll work in smaller PRs after this 😬

PR #2470, opened 2019-06-17
Instrumentation!

mikeshultz:
### Description:

This PR is mostly to help diagnose #2346, and takes a couple shots in the dark at fixing it.  So far I've been unable to repro the error at all.

This includes:

- ~~A change to provide Sentry with the proper dapp URL~~
- Logging messages in places I think might help
- Some queries are now skipped without proper ID
- ~~Enables source maps for production, though I'm not sure if anything else needs to be done here for this to work in prod.  @nick?~~

### Checklist:

### DIFF
--- a/dapps/marketplace/src/components/Reviews.js
+++ b/dapps/marketplace/src/components/Reviews.js
@@ -62,6 +62,10 @@ export default class Reviews extends Component {
     const { id, after } = this.props
     const first = this.props.first || 3
 
+    if (!id) {
+      console.error('Error: Reviews: id missing!')
+    }
+
     return (
       <Query
         query={query}
--- a/dapps/marketplace/src/components/Reviews.js
+++ b/dapps/marketplace/src/components/Reviews.js
@@ -62,14 +62,11 @@ export default class Reviews extends Component {
     const { id, after } = this.props
     const first = this.props.first || 3
 
-    if (!id) {
-      console.error('Error: Reviews: id missing!')
-    }
-
     return (
       <Query
         query={query}
         variables={{ id, first, after }}
+        skip={!id}
         notifyOnNetworkStatusChange
       >
         {({ data, loading, error, fetchMore, networkStatus }) => {
--- a/dapps/marketplace/src/index.js
+++ b/dapps/marketplace/src/index.js
@@ -13,7 +13,6 @@ import ReactDOM from 'react-dom'
 import { ApolloProvider } from 'react-apollo'
 // import { persistCache } from 'apollo-cache-persist'
 import { HashRouter } from 'react-router-dom'
-import { createBrowserHistory } from 'history'
 import Styl from 'react-styl'
 import client from '@origin/graphql'
 import * as Sentry from '@sentry/browser'
@@ -23,8 +22,6 @@ import setLocale from 'utils/setLocale'
 import App from './pages/App'
 import Analytics from './components/Analytics'
 import './css/app.css'
-
-let browserHistory
 
 if (process.env.NODE_ENV === 'production') {
   try {
@@ -38,24 +35,6 @@ if (process.env.NODE_ENV === 'production') {
       release: `marketplace-dapp@${process.env.GIT_COMMIT_HASH}`,
       environment: process.env.NAMESPACE
     })
-    // Update the `url` on route change
-    try {
-      browserHistory = createBrowserHistory()
-      browserHistory.listen(loc => {
-        console.log('browserHistory loc: ', loc)
-        Sentry.configureScope(scope => {
-          const winLoc = window.location
-          const fullURL = `${winLoc.protocol}//${winLoc.host}${loc.pathname}${
-            loc.search
-          }${loc.hash}`
-          console.debug('changed path: ', fullURL)
-          scope.setTag('url', fullURL)
-        })
-      })
-    } catch (err) {
-      console.warn('Unable to configure browser history')
-      console.debug(err)
-    }
   }
 } else {
   try {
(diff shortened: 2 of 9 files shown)


dapps/marketplace/webpack.config.js
```diff
@@ -25,7 +25,7 @@ const config = {
   entry: {
     app: './src/index.js'
   },
-  devtool: isProduction ? false : 'cheap-module-source-map',
+  devtool: 'cheap-module-source-map',
```
nick: Does this include the source map in the main build? Need to verify it's in a separate file that doesn't get downloaded by clients by default

dapps/marketplace/src/pages/user/User.js
```diff
@@ -17,6 +17,11 @@ import { getProviderDisplayName } from 'utils/profileTools'
 const User = ({ match }) => {
   const id = match.params.id
   const vars = { id: match.params.id }
+
+  if (!id) {
```
nick: Instead of this we can just add `skip={!id}` to the `Query` component do the query doesn't even get run if id isn't present

dapps/marketplace/src/index.js
```diff
@@ -67,7 +89,7 @@ class AppWrapper extends Component {
     if (!ready) return null
     return (
       <ApolloProvider client={client}>
-        <HashRouter>
+        <HashRouter history={browserHistory}>
```
nick: Hrm not sure if this will cause unintended side effects... perhaps we can remove this for now and look at it again if the error still isn't fixed with the other changes

dapps/marketplace/src/components/Reviews.js
```diff
@@ -62,6 +62,10 @@ export default class Reviews extends Component {
     const { id, after } = this.props
     const first = this.props.first || 3
 
+    if (!id) {
+      console.error('Error: Reviews: id missing!')
+    }
+
     return (
       <Query
         query={query}
```
nick: lets try adding `skip={!id}` to the `Query` component and remove the console.error above. The query shouldn't be getting run at all if no id is set

packages/graphql/src/resolvers/Marketplace.js
```diff
@@ -58,5 +59,19 @@ export default {
     return events.length
   },
   users,
-  user: (_, args) => ({ id: args.id, account: { id: args.id } })
+  user: (_, args, context, info) => {
```
nick: Can we just return `null` if args.id isn't set?

packages/graphql/src/resolvers/User.js
```diff
@@ -145,6 +146,20 @@ async function reviews(user, { first = 10, after }) {
...
+      console.log('User.id', review.reviewer.id)
+      console.log('path: ', apolloPathToString(info.path))
+      console.log('returnType: ', info.returnType)
+      console.log('operation: ', info.operation.operation)
+      console.log(
+        '========================================================================='
+      )
+    }
+
     if (review.rating) {
```
nick: maybe lets add a check here that reviewer.id is also available before the node gets added?

dapps/marketplace/webpack.config.js
mikeshultz: webpack is seriously a strange black box to me.  I just did a test prod build and there's no `.map` files in the build dir.  I also can't seem to find any details about `cheap-module-source-map` [in the webpack docs](https://webpack.js.org/configuration/devtool/) beyond a mention of a setting.  Suggestions?

nick: hrm maybe let's just put it back to how it was for now and do this separately...

dapps/marketplace/src/pages/user/User.js
mikeshultz: is that ideal?  I'd kind of like to know if this is an issue.  Though I did take one other shot in the dark at fixing this as well.

nick: I've been doing it elsewhere in the code and think it's fine, yea

dapps/marketplace/src/index.js
mikeshultz: I still think it would be immensely valuable to have this so Sentry has the proper URL, even if it has no baring on this specific issue.  What're your concerns?  

It's building a history obj using the same dependency of react-router.  Even if that somehow fails and `browserHistory` is undefined, I don't should be an issue.

packages/graphql/src/resolvers/Marketplace.js
mikeshultz: Should id ever not be set?  I have the same concerns as the other suggestion, that it wouldn't help diagnose.  But, I'm game if you want.

dapps/marketplace/webpack.config.js
mikeshultz: Could be really useful for this... but aight.  Reverted.

packages/graphql/src/resolvers/Marketplace.js
nick: we can keep in the console.log but we should never return { id: null }. Perhaps just add return null after line 73

dapps/marketplace/src/index.js
nick: Happy to merge this in a separate PR after more testing. My only concern is it changes URLs from dapp.com/#/listing/123 to dapp.com/listing/123 which would break a lot of stuff, including deploying to non-root URLs like https://originprotocol.github.io/test-builds/

nick: also... does Sentry not support hash based URLs? I did a quick search and it said it did...

mikeshultz: What do you mean it changes URLs?  This is literally the same dependency that react-router already uses and I'm not making any changes to the router's internal, nor the browser's history.  It's only used to listen for location change events.

Right now, they're treated as fragments, not as a first-class URL as they are with the dapp.  But I don't really want to argue this forever, so I'll revert this...

PR #2472, opened 2019-06-18
Fix attestations on staging and production

shahthepro:
For some reasons the IdentityResolver.js file was loaded even before loading the environment variables. Most likely has to do something with the order of require/import during babel transpilation.

### DIFF
--- a/packages/graphql/env.js
+++ b/packages/graphql/env.js
@@ -1,0 +1,8 @@
+import dotenv from 'dotenv'
+dotenv.config()
+
+try {
+  require('envkey')
+} catch (error) {
+  console.warn('EnvKey not configured')
+}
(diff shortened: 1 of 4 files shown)


tomlinton:
I saw in Discord that you said that you tried moving the envkey require to the top of `src/server.js` but it didn't help. It seems to me like that should fix it. Perhaps it was the transpiled babel being cached? I think you can disable it with the `[BABEL_DISABLE_CACHE](https://babeljs.io/docs/en/babel-register)` env var.

shahthepro:
@tomlinton I tried disabling babel cache but still resolvers files are loaded before environment variables are loaded. Most likely, if envkey loads the environment variables asynchronously, that should explain this behaviour.

tomlinton:
Ahh it is because babel always moves imports above statements, so the imports end up ahead of the require for EnvKey. I changed the EnvKey to be an import, we can't catch the case where there is no ENVKEY env var set anymore but that probably doesn't matter.

tomlinton:
The tests needed the server so I had to do it a different way.

shahthepro:
Thanks @tomlinton

PR #2489, opened 2019-06-19
Fix profile strength visibility and Horizontal listing view style

shahthepro:
- [x] Fix bug that hides profile strength
- [x] Fix "Listings" (horizontal) section style on public profile.


nick:
Now part of #2490

PR #2502, opened 2019-06-21
Purse error handling improvements and gas limit raised

mikeshultz:
### Description:

Pretty much just better RPC error handling and raising the gas limit to 25gwei.  

Also includes a mechanism to prevent Purse from firing out and **unyielding assault** of funding transactions to the children if the transactions aren't confirmed quickly, which is probably important for mainnet deployment.

EDIT: To be clear, that was hyperbole.  But it will send out unnecessary funding transactions if they aren't mined quickly.

### Checklist:


infra/relayer/migrations/20190620211348-enum-add-gaslimit.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+const ENUM_NAME = 'enum_relayer_txn_status'
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.query(`ALTER TYPE ${ENUM_NAME} ADD VALUE 'GasLimit';`)
```
franckc: use ADD VALUE IF NOT EXISTS so that in case we undo the migration it can run again.

infra/relayer/migrations/20190620211348-enum-add-gaslimit.js
```diff
@@ -0,0 +1,13 @@
...
+  },
+
+  down: () => {
```
franckc: Don't we need to return a Promise ? Maybe return Promise.resolve()

infra/relayer/src/relayer.js
```diff
@@ -281,10 +281,20 @@ class Relayer {
...
+        let errMsg = 'Error sending transaction'
+
+        if (reason.message && reason.message.indexOf('gas prices') > -1) {
+          status = enums.RelayerTxnStatuses.GasLimit
+          errMsg = 'Network is too congested right now.  Try again later.'
+        }
+
         if (dbTx) {
-          await dbTx.update({ status: enums.RelayerTxnStatuses.Failed })
+          await dbTx.update({ status })
```
franckc: I wonder if we should store the reason as a field in the data JSONB column ? For ease of troubleshooting rather than having to look into the machine logs.

infra/relayer/migrations/20190620211348-enum-add-gaslimit.js
mikeshultz: Considering there's no `down` for this one, I think that's reasonable.

infra/relayer/migrations/20190620211348-enum-add-gaslimit.js
mikeshultz: Not sure what you mean?  This has been tested to work, btw.

infra/relayer/src/relayer.js
mikeshultz: What's the purpose of the `data` field, anyway?  Doesn't appear used. Perhaps this is a TODO item?

infra/relayer/migrations/20190620211348-enum-add-gaslimit.js
franckc: Did you try to run sequelize db:migrate:down ? If yes, I could be wrong - I thought I encountered an issue on similar code in a PR of mine...

infra/relayer/src/relayer.js
franckc: Data field was for future needs to store any interesting metadata. I think storing the details of the failure reason in there would be a good use for it.

infra/relayer/migrations/20190620211348-enum-add-gaslimit.js
mikeshultz: I did not realize you were talking about `down()` for some reason.  You are correct, I get this with an `undo`:

    ERROR: Migration 20190620211348-enum-add-gaslimit.js (or wrapper) didn't return a promise

Will get a PR up shortly

PR #2510, opened 2019-06-22
Child funding changes

mikeshultz:
### Description:

Initially this was setup so that individual child accounts could get funded when they need to.  This caused Purse to try and fund 3 accounts when the balance of the master could only afford to fund 2.  Not the end of the world, but it caused a tx to be forever-broadcast until I guess the master balance was satisfied enough to send it.  

In addition to this change, the values for baseFunding could be set differently if 0.5 Ether is too much, but I'm going by @franckc's suggested numbers here.  I lowered prod down to 2 children so things would run smoothly for now.

### Checklist:

### DIFF
--- a/infra/relayer/package.json
+++ b/infra/relayer/package.json
@@ -40,7 +40,7 @@
     "start": "per-env",
     "start:development": "nodemon src/app.js",
     "start:production": "node src/app.js",
-    "test": "NODE_ENV=test mocha --timeout 10000 --file test/setup --exit"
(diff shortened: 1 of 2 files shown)


mikeshultz:
It isn't really necessary to change the default.  Was configured on prod using the env var to bring it down to `2`.

franckc:
Ah yes, sorry I did not notice you had set FORWARDER_ACCOUNTS to 2 in Envkey... let me merge this to master.

PR #2515, opened 2019-06-24
Store userId and other profile details to DB and IPFS

### LINKED ISSUE #2443: Store user IDs from attestations
Currently, for most attestations, we only store in the DB and on the blockchain whether or not the user was verified. But we do not store the user ID.

Moving forward, for all attestations (except phone and email), we'd like to:
- record the user ID in the DB.
- record the user IDs in the IPFS data blob so that we are transparent with our users as to what data we store.

On the UX for all these attestations, we should make sure to be very explicit with the user as to what information they are about to make public.

Note that for facebook, while the user's global facebook ID is not available, we do get back from facebook  a user ID scoped to the Origin app. So we should store that.

@shahthepro In terms of priority, this is a P2 for this sprint. But likely a P1 for next sprint as we get closer to July during which we'll run an Origin Reward campaign for a lot of the new attestation you added...

@micahalcorn Please chime in with more thoughts/details about all this.

shahthepro:
Should close #2443 

A few notes:
- For LinkedIn, we can get the user name (`vanityName`) and public profile url only if we are enrolled in their [Partner Program](https://developer.linkedin.com/partner-programs/apply#). For now, I'm storing only unique user Id. 
- Have dropped the columns `kakaoVerified`, `githubVerified`, `linkedinVerified` and `wechatVerified` from `identity` table. However `googleVerified` and `facebookVerified` are still there. As of now, if the unique Id is not in DB for Google and Facebook (which will be case for attestations that happened before this gets merged), it falls back to these two boolean values.

### Checklist:
- [x] Update texts in DApp on what is stored in DB and IPFS


infra/bridge/src/controllers/google.js
```diff
@@ -98,7 +98,10 @@ router.post('/verify', googleVerify, async (req, res) => {
     site: {
       siteName: 'google.com',
       userId: {
-        verified: true
+        raw: String(userDataResponse.body.id)
+      },
+      username: {
+        raw: userDataResponse.body.email
```
micahalcorn: @nick we probably don't want to store this in IPFS, do we?

nick: yea, no email in IPFS

infra/bridge/src/controllers/phone.js
```diff
@@ -149,7 +149,9 @@ router.post('/verify', phoneVerifyCode, async (req, res) => {
   const attestation = await generateAttestation(
     AttestationTypes.PHONE,
     attestationBody,
-    attestationValue,
+    {
+      uniqueId: attestationValue
```
nick: what is the value here?

infra/bridge/src/controllers/website.js
```diff
@@ -64,7 +64,9 @@ router.post('/verify', websiteVerify, async (req, res) => {
     const attestation = await generateAttestation(
       AttestationTypes.WEBSITE,
       attestationBody,
-      remoteOrigin,
+      {
+        uniqueId: remoteOrigin
```
nick: Are there any guarantees this is unique? I guess multiple users could attest to the same URL?

infra/bridge/src/controllers/phone.js
shahthepro: It's normalized phone number. https://github.com/OriginProtocol/origin/blob/b2be87875ee7764e2b922fa21cc762bd9b5e3754/infra/bridge/src/controllers/phone.js#L140-L147

infra/bridge/src/controllers/website.js
shahthepro: No, we don't have a unique check for this. Multiple addresses can attest with the same domain, subdomain or on a different port. We don't have a unique/primary constraint in the database column either, so it should not throw in case of a duplicate domain.

But we can switch this with singed message. That should be unique for each address, even if the domain is same.

Issue #228, opened 2018-05-28
Price approximation failure


micahalcorn:
@inkredabull @DanielVF 👇 

On a seemingly random basis, I get listing cards with USD prices sticking at "Loading..." due to the following (in Safari):

`[Error] Cross-origin redirection to https://api.cryptonator.com/api/ticker/eth-usd denied by Cross-Origin Resource Sharing policy: Access-Control-Allow-Origin cannot contain more than one origin.`

I tried reaching out to Cryptonator to see if we could whitelist the (Origin) origin, but I'm not getting anywhere. Do we want to just hide the price approximation until it loads successfully or is there a better fallback?

_This happens in Chrome also and in various environments._

DanielVF:
We're currently hitting the API once for every listing - we might be tripping some kind of rate limiting. We can reuse the same conversion rate for multiple listings.

wanderingstan:
@DanielVF good point. We should definitely hit the API once and store it for dapp-wide useage.

inkredabull:
Wrapping up other responsibilities today before heading out on vacation tomorrow for a week. 
Have unassigned myself if it needs to get done sooner than the week of 6/18.

inkredabull:
https://github.com/OriginProtocol/origin-dapp/pull/290

micahalcorn:
Done via #290 ✅

PR #2519, opened 2019-06-24
Exchange Rate Caching

shahthepro:
It seems like the API call that is made to fetch exchange rate is rate-limited and fails some times. Instead of invoking that API every time, we can cache it in Redis/DB and poll it in fixed intervals. We can have all requests query a single endpoint on bridge that returns the cached exchange rate for that crypto-fiat pair.


shahthepro:
@tomlinton As of now, Polling only works for ETH/USD market. I just wanted to know what you think about this before writing a more generic version of the poller

tomlinton:
Thanks for doing that so quickly @shahthepro! It looks great. Only thing I add is maybe adjust a few of the logger calls to warnings or errors where appropriate as they are all `logger.log`. It might make it easier to read the logs down the track.

I'll just add a bit of context here for anyone wondering what this is about. We were seeing `cryptonator.com` not responding to requests for some reason (perhaps rate limiting, perhaps just issues at their end, more [discussion here](https://github.com/OriginProtocol/origin/issues/228)). If that happens and the Origin marketplace app can not retrieve any exchange rate it'll crash when trying to convert crypto to fiat values. @shahthepro and I discussed and decided it'd be good to add our own exchange rate endpoints to the bridge as its a lot easier to ensure our own server works, and we won't have to push new builds to the App/Play stores if we need to swap out the exchange rate API.

franckc:
I like the idea. Only one question: have you considered putting this in the graphql server rather than bridge ? Bridge currently specializes in handling attestation related methods. Currency conversion feels more like business logic... The drawback of putting this in graphql is that it would need to run only on server side and not in dapp when performance mode is disabled...
I'm just asking to make sure you considered that option - I don't feel super strongly. If it is much simpler to implement in bridge, totally fine to go that route.

tomlinton:
I don't really have any strong feelings either. I think of bridge as a bridge between any third party APIs which this fits nicely into. I'd prefer we just have a single source of truth for it though - as you noted it gets tricky if this is running in GraphQL client side and we start seeing the failures again.

infra/bridge/src/controllers/utils.js
```diff
@@ -34,4 +36,13 @@ router.post('/exists', async (req, res) => {
   res.end()
 })
 
+/**
+ * Returns excahnge rate of given market
```
tomlinton: Typo

shahthepro:
@franckc 

> have you considered putting this in the graphql server rather than bridge? 

No, I also thought that bridge was to be used with external API services. If we were to add this in GraphQL server instead, we can probably use Apollo's InMemory cache instead of Redis. But still we may run into issues when not in performance mode as you say.

PR #2522, opened 2019-06-25
Mobile 2.0 refinements

tomlinton:
A bunch of misc improvements following testing:

- Use allowsBackForwardNavigationGestures on iOS because it is nicer than the custom pan handler implementation. Still using the custom pan handler on Android.
- Better pull to refresh implementation, thanks @nick!
- Fix for an Android build issue (also removal of `react-native-test-flight` because it wasn't happy on Android).
- Autofocus on last name input when return is hit from the first name input in the onboarding name entry.
- Guess the country of the phone attestation from the users phone locale setting so they hopefully don't need to scroll through the long list.


mobile/src/screens/marketplace.js
```diff
@@ -323,9 +330,7 @@ class MarketplaceScreen extends Component {
   /* Handle the response from window.onScroll
    */
   handleScrollHandlerResponse = ({ scrollTop }) => {
-    if (scrollTop < -100) {
-      this.dappWebView.injectJavaScript(`document.location.reload()`)
-    }
+    this.setState({ enablePullToRefresh: scrollTop === 0 })
```
nick: When I was testing I think everything still worked without having to test for scrollTop = 0

tomlinton: I think because the RefreshControl triggers a refresh when scrollY is 0 in the ScrollView it needs to be here. The scrolling happens in the WebView so scrollY is always 0 in the ScrollView. On Android it'll refresh if you try to scroll up after scrolling down in the DApp. I'm not sure why it doesn't do the same thing on iOS though.

PR #2525, opened 2019-06-25
Adds unique id to E-mail notifications for messages

mikeshultz:
### Description:

This adds a unique ID to the E-mail notifications for messages, as well as to the DB.  This should be able to help troubleshoot some issues(e.g. #2495)

### Checklist:


infra/messaging/migrations/20190625030841-add-content-hash.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.addColumn('Message', 'contentHash', {
+      type: Sequelize.STRING(66)
```
franckc: I'd recommend using String rather String(66) in case we change hash function in future.
From DB perspective, there is no storage saving by using String)66).

infra/messaging/migrations/20190625030841-add-content-hash.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.addColumn('Message', 'contentHash', {
```
franckc: Isn't the table name msg_message rather than Message ?

infra/notifications/src/emailSend.js
```diff
@@ -38,6 +38,11 @@ async function messageEmailSend(receivers, sender, messageHash, config) {
     return r.toLowerCase()
   })
 
+  let shortHash = ''
+  if (messageHash && messageHash.startsWith('0x') && messageHash.length > 9) {
+    shortHash = messageHash.slice(2, 7)
```
franckc: Any reason to use a shortHash rather than full hash to embed in the email ?
It will be hidden content to the user anyway and a few bytes more or less don't matter ?

infra/messaging/migrations/20190625030841-add-content-hash.js
mikeshultz: I know this, but I always forget how Postgres does underlying sizing with flexible columns.  Will make this change in a bit.

infra/messaging/migrations/20190625030841-add-content-hash.js
mikeshultz: I haven't fully tested this out.  Was using the name in the model.  You're probably right, and I'll make sure to test the migration before bringing this PR out of draft.

infra/notifications/src/emailSend.js
mikeshultz: Wasn't initially expecting it to be fully hidden, but you have a point.

infra/messaging/migrations/20190625030841-add-content-hash.js
mikeshultz: So... apparently Sequelize defaults to varchar(255) if you don't specify a length and I haven't found a way to use a standard postgres style `varchar`.  I'd rather live with a fixed `66` length than a fixed `255` in this case.

franckc: Oh interesting. I actually did not realize that behavior of sequelize...

PR #2539, opened 2019-06-26
Fix android keyboard issue

tomlinton:
This should fix keyboard overlapping the modal and not allowing input.


shahthepro:
@tomlinton This fixes the scroll on Android however breaks it on iOS.

tomlinton:
I've changed the fix. Seems to work OK on iOS too now. I'll merge this because even if the iOS issue persists it was just a glitchy scroll rather than being totally broken like it is currently on Android. You can still complete the onboarding. I

Issue #2398, opened 2019-06-04
DApp Redesign


micahalcorn:
This tracks the work on implementing the new styles and related UX from [our mobile-first DApp designs](https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/dashboard?seid=5ce48c3ee30c3a1e6861bfb5).

aka 𝓓𝓐𝓟𝓟𝟛

-----

At Origin, we use [ZenHub](https://www.zenhub.com/) to manage our engineering tasks and product development. [Download their browser extension](https://www.zenhub.com/extension) and check out our open workspace at [github.com/originprotocol/origin#zenhub](https://github.com/originprotocol/origin#zenhub).

micahalcorn:
Let's keep a running list here of things to fix in 𝓓𝓐𝓟𝓟𝟛 and then later move some to separate issues.

- [ ] edit
- [ ] me
- [ ] ☝️

PR #2548, opened 2019-06-27
𝓓𝓐𝓟𝓟𝟛 Tweaks

micahalcorn:
### Description:

Simple fixes noted in #2398 🙂 

### Checklist:

- [ ] ~Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)~


dapps/marketplace/src/pages/listings/Listings.js
```diff
@@ -81,6 +81,8 @@ class Listings extends Component {
             <fbt:param name="category">
               {this.state.search.subCategory.type === 'clothingAccessories'
```
shahthepro: Just a suggestion. If we are going to override more category names here, We can probably do it with a `switch` statement. IMO, It will be more readable than nesting conditional statements.

micahalcorn: I don't disagree, but I remembered reading this 😁 https://medium.com/javascript-scene/nested-ternaries-are-great-361bddd0f340 Also, I was being lazy.

PR #2550, opened 2019-06-27
Mobile currency injection

### LINKED ISSUE #2537: Location-Dependent Display Currency
We should be detecting the user's location and defaulting to the appropriate display currency if we support it. I believe that we are already doing the some for language, and we want to do our best to save the user from having to find the settings page.

tomlinton:
Makes the DApp in our mobile app use the best available currency depending on the mobile phones region.

Required adding of an onCurrency method similar to the existing onLocale so we can do `window.appComponent.onCurrency('fiat-XXX')`, which is nicer than just throwing it into local storage because it updates everything without a refresh. It wasn't much of a chance but involved moving the currency state up a level.

Closes #2537.


tomlinton:
@nick would you mind checking the DApp part of this when you get a chance?

PR #2551, opened 2019-06-27
Mobile native growth onboarding

tomlinton:
- Adds native growth onboarding UI and GraphQL/queries mutations
- Injects growth auth token into DApp local storage
- Fixed a crash if identity request failed

@franck and @sparrowDom I'm not too sure about the fingerprint data for the growth fraud piece here. I'm using a library which can retrieve a unique device id that is persistent across reinstalls of the app. I was hoping it'd be a good replacement for the UUID from desktop. The JSON structure is of course different to the desktop fingerprint data, but I thought we'd just adapt the fraud code. Let me know what you'd like!


franckc:
@tomlinton Your library for mobile seems nice. I can update the fraud script to handle that new mobile uuid. Perhaps we can save it in the growth_participant.data as "uuid_mobile" ?

tomlinton:
I can definitely rename it, right now the key is `uniqueId`. I might change it to `mobile_id` if you prefer that? I think we should avoid `mobile_uuid` because it isn't a true uuid.

franckc:
Sure, mobile_id sounds good ! ty sir

PR #2555, opened 2019-06-27
Upgrade npm packages

nick:
Upgrade npm packages throughout the monorepo (except mobile).

I skipped out some packages from the upgrade such as ganache-core, ipfs and some others


infra/bridge/test/twitter.js
```diff
@@ -119,7 +119,7 @@ describe('twitter attestations', async () => {
       req.session = {}
       req.sessionStore = {
         get(sid) {
-          expect(sid).to.equal(123)
+          expect(sid).to.equal('123')
```
mikeshultz: Could this have wider range impacts?  Don't see any related changes.

nick: @shahthepro after upgrading modules I had to modify this test... does it look ok to you?

shahthepro: @nick It shouldn't be a problem. `req.sessionId` is always a string.

nick:
Also saw there's a GitHub app for doing this automatically: https://github.com/apps/renovate (free for OSS)

infra/bridge/test/twitter.js
shahthepro: Could also [change this line](https://github.com/OriginProtocol/origin/blob/94c96d8534564d35bd7c0e6b014076bfd73f12c8/infra/bridge/test/twitter.js#L74) to make it a string instead of number

PR #2562, opened 2019-06-28
integrates @origin/web3-provider into Purse

### LINKED ISSUE #2477: Purse needs better error handling
The JSON-RPC error handling in Purse is awful and it should be fixed before it reaches production use.  If a provider decides to fail for whatever reason, the relayer probably shouldn't crash.

mikeshultz:
### Description:

This should help with error handling from Alchemy a bit and close #2477 

Though now that I'm writing this PR, maybe this should be in Relayer where the Web3 object is originally inited?  Thoughts, @nick?

### Checklist:


mikeshultz:
Doesn't look like the relayer uses the Web3 instance to make any network calls, so I'll leave this as-is.

PR #2586, opened 2019-07-01
DApp3 Bug Fixes

shahthepro:
Should fix the following from #2398 
- [x] When I go back from the phone attestation form, it closes and then separately reopens the "add verifications" screen. It should act like a navigation stack where I go directly back from phone to all attestations.
- [x] If I cancel the blockchain transaction that results from "Sign to publish", the modal prompt ("...will now ask you to sign...") won't go away and I'm stuck.
- [ ] ~~Both nav drawers should slide open instead of just appearing~~
- [x] The "Earn OGN" button should cause the right drawer to close before (or while) navigating. Otherwise, it remains open when navigating backward.
- [x] The container for the "browse" categories should be full-bleed with no margin so that there isn't some whitespace on the right and left when an icon is cut off in overflow.
- [x] `DownloadApp` should never be shown when the user is in our app.
- [x] Search view with featured categories remains open after clicking a category icon. This causes an odd UX if I navigate backward. It seems to hide after navigating backward if I wait long enough but I don't think it should be there at all.
- [x] Modals dealing with in-progress blockchain transactions should not be dismissible, particularly by clicking the screen outside of the modal.


dapps/marketplace/src/components/Dropdown.js
```diff
@@ -49,10 +49,23 @@ class Dropdown extends Component {
         onMouseOut={() => (this.mouseOver = false)}
       >
         {this.props.children}
-        {this.props.content && this.props.open ? this.props.content : null}
```
nick: This is a fairly major change as now all dropdown content will be rendered and hidden with CSS regardless if the dropdown is open... in a lot of cases that will mean loading extra data that was not required before. Why was this change needed?

packages/graphql/src/configs/localhost.js
```diff
@@ -14,14 +14,15 @@ const config = {
   ipfsGateway: `http://${HOST}:8080`,
   ipfsRPC: `http://${HOST}:5002`,
   relayer: `http://${HOST}:5100`,
-  bridge: 'https://bridge.dev.originprotocol.com',
+  bridge: `http://${HOST}:5000`,
```
nick: Can we revert this, so a local bridge is not required for local dev?

shahthepro: Will revert it.

dapps/marketplace/src/components/Dropdown.js
shahthepro: I added CSS animation for slide in/out for `.dropdown-menu`. The animations didn't happen if the element was rendered instantly with the class `.show`. 

Didn't think about it this way. Will change it.

dapps/marketplace/src/components/Dropdown.js
```diff
@@ -49,10 +49,23 @@ class Dropdown extends Component {
...
+        {this.props.content && this.state.open ? this.props.content : null}
       </El>
     )
   }
 }
 
 export default Dropdown
+
```
nick: Should we remove this for now?

dapps/marketplace/src/pages/nav/Mobile.js
```diff
@@ -49,7 +49,7 @@ const MobileNav = ({ open, onClose, onOpen }) => {
       content={
         <>
           <div className="dropdown-menu-bg" onClick={() => onClose()} />
-          <div className="dropdown-menu show">
+          <div className={`dropdown-menu${open ? ' show' : ''}`}>
```
nick: should we revert changes in this file?

dapps/marketplace/src/pages/nav/Profile.js
```diff
@@ -173,7 +199,11 @@ const ProfileDropdownRaw = ({
   return (
     <>
       <div className="dropdown-menu-bg" onClick={onClose} />
-      <div className="dropdown-menu dropdown-menu-right show profile">
```
nick: revert?

packages/graphql/src/configs/localhost.js
```diff
@@ -21,7 +21,7 @@ const config = {
   performanceMode: localStorageHas('performanceMode'),
   graphql: `http://${HOST}:4007`,
   automine: 2000,
-  attestationIssuer: '0x5be37555816d258f5e316e0f84D59335DB2400B2',
+  attestationIssuer: '0x627306090abaB3A6e1400e9345bC60c78a8BEf57',
```
nick: revert?

dapps/marketplace/src/pages/nav/Profile.js
shahthepro: Done

PR #2589, opened 2019-07-02
Geth upgrade and byzantium fork

mikeshultz:
### Description:

This updates geth to v1.8.27, and makes an attempt at a byzantium fork.  

The fork is tricky as I can't find an authoritative source for how this goes down.  [This StackExchange answer](https://ethereum.stackexchange.com/a/29412/8492) might be the best idea.  I think if we bring up a 3rd miner node with that genesis block, maybe stop the other 2 miners and let this one do its thing, the other should fall in line.  Then maybe if we have to clear state on nodes 0 and 1 and re-sync(not clear if that's necessary).

Though there's also a chance of an incompatibility, and if nodes 0 and 1 come up and don't like the conflicting genesis hash, everything could come to a halt.  So, a bit of a gamble here.  Would be interested in your take on it, @tomlinton.

### Checklist:

### DIFF
--- a/devops/kubernetes/charts/ethereum-network/templates/geth-genesis.configmap.yaml
+++ b/devops/kubernetes/charts/ethereum-network/templates/geth-genesis.configmap.yaml
@@ -16,7 +16,8 @@ data:
             "homesteadBlock": 0,
             "eip150Block": 0,
             "eip155Block": 0,
-            "eip158Block": 0
+            "eip158Block": 0,
+            "byzantiumBlock": 1200000
         },
         "difficulty":  {{ .Values.gethGenesisDifficulty | quote }},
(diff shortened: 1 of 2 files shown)


mikeshultz:
Well, if it does nuke the chain, we'll probably have to clear the state of everything(DB, elasticsearch, redis, etc).  Unless you mean restore from checkpoin/backups or something.  Does kube version volumes somehow?

tomlinton:
Thats not a problem. It might give us a chance to do some additional useful things anyway. I've been wanting to create a one off Kubernetes job that can handle a full reindex, I think it'd be really useful. And at some point we are probably going to want to clear out staging too.

I think it'll be fine though, it is unlikely we'll get it into a state where we can't recover.

Issue #2422, opened 2019-06-06
Mobile-First Transaction Detail


micahalcorn:
Update the transaction detail screen to match these designs:

- mobile starting [here](https://zpl.io/adXRYl7)
- desktop starting [here](https://zpl.io/aMpLj3R)

In addition to the visual changes that are proposed here, we would like to decouple the finalization step from the rating/review step. The buyer's review should now be a separate blockchain transaction using the `addData` method as the seller's review currently does. I'm told by @nick that this will require changing the way that we query for reviews since future ratings/reviews will not be found in the finalization event.

Also, note that the "Learn More" link should lead to a mystery page yet to be designed. For now, it can just open a modal. I can come up with the content for that.

micahalcorn:
Considering this done ✅

sparrowDom:
Thanks @micahalcorn . I should update this stuff more. What is your usual process? When the ticket is in Done agile column is that when it should also be closed?

PR #2590, opened 2019-07-02
Sparrow dom/transaction detail

sparrowDom:
### Description:
Creates transaction detail according to specs: https://github.com/OriginProtocol/origin/issues/2422

### Checklist:


sparrowDom:
@nick when you have time feel free to do a pass. I still need to fix tests and minor text UI changes might be needed after @micahalcorn reviews different transaction detail states.

micahalcorn:
Some UI/UX feedback from the screenshots:

- [ ] Let's remove the tooltips from the Purchase Details and Escrow Details headings.
- [ ] Escrow Details should have a "Learn More >" link, which for now can open the contents that were previously in the tooltip.
- [ ] About the Seller should be vertically aligned to the top.

dapps/marketplace/src/pages/transaction/Transaction.js
```diff
@@ -150,10 +140,28 @@ const Transaction = props => {
             </>
           )
 
+          const VerticalSeparator = <div className="vertical-separator my-3" />
```
nick: Can we move these outside the render method and maybe make then components?

nick: Or just use a div with the className vertical-separator?

dapps/marketplace/src/pages/transaction/Transaction.js
```diff
@@ -168,23 +176,24 @@ const Transaction = props => {
...
-                    {Progress}
-                    {History}
-                    {Listing}
-                  </div>
-                  <div className="col-md-5 col-lg-4 side-bar">
-                    {Offer}
-                    {Escrow}
-                    {About}
+                  <div className="col-12">{Progress}</div>
+                  <div className="col-12 d-flex">
```
nick: maybe we can just use a `<div></div>` instead of `<div className="row"><div className="col-12'></div></div>`?

dapps/marketplace/src/pages/transaction/Transaction.js
```diff
@@ -207,25 +216,32 @@ require('react-styl')(`
...
     .help-icon
       margin-left: 0.5rem
       vertical-align: sub
+    .vertical-separator
+      height: 1px
+      background-color: #dde6ea
+    .horizontal-separator
+      flex: 0 0 13%
+      max-width: 13%
+      > div
```
nick: can this be an :after css class to avoid styling divs

dapps/marketplace/src/pages/transaction/_OfferAcceptedBuyer.js
```diff
@@ -0,0 +1,211 @@
...
+
+  confirmOffer(offerId) {
+    const offersConfirmed = JSON.parse(
+      localStorage.getItem(CONFIRMED_OFFERS_KEY) || '[]'
+    )
+    offersConfirmed.push(offerId)
+    this.setState({
+      offerConfirmed: true
+    })
+    localStorage.setItem(CONFIRMED_OFFERS_KEY, JSON.stringify(offersConfirmed))
```
nick: can we not use state for this?

dapps/marketplace/src/pages/transaction/_Progress.js
```diff
@@ -96,177 +112,133 @@ const AcceptOrReject = ({ offer, refetch, loading }) => (
...
 )
 
-const OfferWithdrawn = ({ offer, party, loading }) => (
+const OfferWithdrawn = ({ offer, viewedBy, loading }) => (
   <div className={`transaction-progress${loading ? ' loading' : ''}`}>
     <div className="top">
       <h4>
-        <fbt desc="Progress.offerWithdrawn">Offer Withdrawn</fbt>
+        {viewedBy === 'seller'
+          ? fbt(
```
nick: can we use a regular `<fbt>` component unless we really need the function version?

dapps/marketplace/src/pages/transaction/_ReviewAndFinalization.js
```diff
@@ -0,0 +1,200 @@
...
+      localStorage.getItem(RATED_OFFERS_KEY) || '[]'
+    )
+    offersRated.push(offerId)
+    this.setState({
+      offerRated: true
+    })
+    localStorage.setItem(RATED_OFFERS_KEY, JSON.stringify(offersRated))
+  }
+
+  renderRating({ offer, loading, isSeller, isBuyer }) {
```
nick: I think we should try to use functional components instead of render methods going forward so we can try to transition away from React classes

dapps/marketplace/src/pages/transaction/Transaction.js
sparrowDom: I would use a div with the class name, but didn't want to copy-paste the code 3 times. Horizontal separator also has a nested div. 

My thinking for moving that to variables was to make code less error prone. Since you can quickly change properties of one vertical/horizontal separator and not the other one. When defined with a variable that is not a problem. 

If you think necessary I can use the divs inline the rest of the jsx.

dapps/marketplace/src/pages/transaction/Transaction.js
sparrowDom: Not sure I follow... are you suggesting removing `row` and `col-12` classes.

dapps/marketplace/src/pages/transaction/Transaction.js
sparrowDom: Ah yes good point tnanks

dapps/marketplace/src/pages/transaction/_OfferAcceptedBuyer.js
sparrowDom: Leaving it solely to state will result in not surviving page refreshes. Ideally this would work cross device (persisted in the smart contract/ipfs). With the `localStorage` solution it works on a single device across page refreshes.

dapps/marketplace/src/pages/transaction/_ReviewAndFinalization.js
sparrowDom: Cool converting to functional component with `useState` and `useEffect`

sparrowDom:
Thanks for a thorough UX review @micahalcorn and code @nick . There are still some open discussions on the code review part...

dapps/marketplace/src/pages/transaction/Transaction.js
nick: Yea, a row + col-12 is the same as just a div, right?

dapps/marketplace/src/pages/transaction/_OfferAcceptedBuyer.js
nick: Can we not get if the offer was confirmed from GraphQL?

dapps/marketplace/src/pages/transaction/_Progress.js
sparrowDom: Yeah this bit is pretty weird. I used the function because the linter would complain about this code: 
```
<fbt dec="smth">You've ....</fbt>
```
with the following error: 
```
origin/dapps/marketplace/src/pages/transaction/_Progress.js
  198:18  error  `'` can be escaped with `&apos;`, `&lsquo;`, `&#39;`, `&rsquo;`  react/no-unescaped-entities
```

This way there were no linter problems. I kind of prefer calling a function over escaping the `'` character with `&apos;`

Do you know a better solution for this maybe?

dapps/marketplace/src/pages/transaction/Transaction.js
sparrowDom: Yup good point 👍

PR #2592, opened 2019-07-02
Mobile onboarding back arrows

tomlinton:
Adds back arrows where needed in the mobile onboarding process according to what is in Zeplin. Positioned them slightly differently (alongside titles) to try and save space.


micahalcorn:
⬅️

PR #2594, opened 2019-07-02
Bridge server now requires Redis.

DanielVF:
The bridge server now requires Redis, but the docker-compose.yml file wasn’t updated to configure the needed connection URL.

This fix lets the bridge server start again in our docker env.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -252,6 +252,7 @@ services:
     volumes: *volumes
     environment:
       - DATABASE_URL=postgres://origin:origin@postgres/origin
+      - REDIS_URL=redis://redis-master:6379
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -252,6 +252,7 @@ services:
     volumes: *volumes
     environment:
       - DATABASE_URL=postgres://origin:origin@postgres/origin
+      - REDIS_URL=redis://redis-master:6379
```
mikeshultz: Need a DB number here?  e.g. `redis://redis-master:6379/0`  If it defaults to 0, cool.

DanielVF: Looks like it doesn't need it - see the other REDIS_URL env in the docker-compose.yml

PR #2597, opened 2019-07-02
PostgreSQLBackend fixes for event-cache

mikeshultz:
### Description:

EventCache's PostgreSQLBackend needed a bunch of fixes to get right with the world.  This should maybe do it.

This has been configured to be enabled through env vars.  Would like to initially start testing on dev for a bit.

### Checklist:


mikeshultz:
@DanielVF, when you have time and if you're interested in trying to poke holes in this, I'd appreciate it.  This isn't really a high priority, so no rush.

franckc:
@tomlinton I'm pretty swamped probably for rest of week on launching social sharing and won't likely have time to do an in-depth review. Do you have any cycles to review so that @mikeshultz is not blocked waiting on this ? thanks...

tomlinton:
Yep will do!

franckc:
Thanks @tomlinton !

mikeshultz:
Merged, I'll probably start testing this on dev tomorrow.

mikeshultz:
Reverted, build kept failing for the dapp, looks like because of `pg-hstore` package.  Doing the docker build manually on my local **did not** show this problem.  Could be temporary, but might need to look into how imports are resolved, or moving deps to optional deps... not sure yet.  

Local test build:

    docker build --build-arg ENVKEY="abcd123" -t origin-dapp -f devops/dockerfiles/origin-dapp .

Have to test this with cloudbuild, I guess, or try and figure out what's different between my local build and cloudbuild.  Will find a slow night to test it out again.

Failing build example: https://console.cloud.google.com/cloud-build/builds/9b01c9fc-09a9-4d40-a72d-a18c4223e57b?project=origin-214503

@tomlinton, anything jump out at you from that build?

franckc:
Not sure this helps, but what do we use pg-hstore for ? Could we use JSONB storage and get rid of the pg-hstore package dependence ?

mikeshultz:
It's not our dep. Must be secondary.

tomlinton:
That is strange. We do have `pg-hstore` as a dependency for some other packages. Nothing obvious from the build log.

I'd be most suspicious of this:

```
if (typeof window === 'undefined') {
  PostgreSQLBackend = require('./backends').PostgreSQLBackend
}
```

mikeshultz:
Well, yeah, that's the postgres import which will eventually apparently lead to pg-hstore.  But still, why would it be evaluated during webpack build as being part of the bundle?  Why isn't that reproducible in my local docker build using the same docker container and source?

Gonna explore that a bit over the next couple days, I guess.  Would be best if I can repro this locally instead of having to throw commits at dev.

tomlinton:
I got a local repro using the docker build command. Also found this which seems to be the issue:

https://github.com/sequelize/sequelize/issues/7509

So we can probably get it to work by adding pg-hstore to the externals config in webpack, not sure what it is going to do to the bundle size though :grin:

tomlinton:
I set externals to:

```
externals: {
    'pg-hstore': 'pg-hstore',
    Web3: 'web3'
}
```
Which got it to build, but it added ~1mb to the bundle size unfortunately.

PR #2598, opened 2019-07-02
Proxy contract build

nick:
- Update proxy contract bytecode
- Update proxy contract deployed addresses


packages/contracts/README.md
```diff
@@ -182,12 +179,26 @@ start their own local blockchain instance.
 
 To run contract tests and automatically re-run when files change:
 
-```
+```js
```
DanielVF: Aren't these more properly `bash` than JS?

nick: ah yea 🙂

PR #2602, opened 2019-07-03
Fix escaped ›

tomlinton:
Fix > characters in CSS. The uglify/minify breaks the escaping so we end up with this:

Screen Shot 2019-07-03 at 4 03 59 PM


sparrowDom:
nice thanks @tomlinton

PR #2609, opened 2019-07-03
Listing Detail (DApp3) Changes

shahthepro:
Adds the following changes to DApp3
- [x] Fix "View my offer" link
- [x] Show NavBar and Search bar on the page
- [x] Add box-shadow to Gallery component on this page
- [x] Limit the width of Listing Detail, if no images are there (only Desktop)


infra/discovery/src/listener/handler_marketplace.js
```diff
@@ -32,14 +37,6 @@ function isOfferEvent(eventName) {
   return OFFER_EVENTS.includes(eventName)
 }
 
-function getOriginListingId(networkId, event) {
```
shahthepro: @franckc @mikeshultz I had a need to reuse these two functions in the DApp. I was not sure about adding `@origin/discovery` as a dependency for the DApp. So I moved these two functions to `@origin/graphql` package since both the DApp and `@origin/discovery` are dependent on that already.

Let me know if you have any other suggestions for this.

mikeshultz: I have no issues with moving it and nothing else comes to mind for a better place.  Is the `esmImport` necessary?

shahthepro: I don't think we can `require` it, since `@origin/graphql` package uses ES6 modules (`import`/`export`). We don't have babel in `@origin/discovery` package. Could have added that but every other file uses `esmImport`. So used that for consistency.

Issue #2543, opened 2019-06-26
IPFS Pinner revamp


DanielVF:
Our IPFS pinner is currently broken. 

When the we changed the listener to use the GraphQL backend, the format of the data the IPFS pinner receives changed. It's not been pinning anything since that change.

- [ ] Support new format
- [ ] Add support for storing avatarUrls from profiles

franckc:
Thanks for creating this issue. Would also be good to run a backfill to pin the old data. This could either be done with a separate script that we write specifically for this or might be simpler to re-run the listener and re-process old data (let me know and I can run that in prod).

PR #2615, opened 2019-07-03
Fix IPFS Pinner by redoing parsing to match new formats.

DanielVF:
This fixes the IPFS pinner to work with the new post-GraphQL data format.

Parsing tests are updated to check that it works.

I can't actually test the whole integration with google function, and the whole devops side. Let me know if there are any problems.


franckc:
Adding @mikeshultz as reviewer since he is an IPFS guru and probably wants to be in the loop... :)

DanielVF:
Closes #2543

tomlinton:
Deployed this to GCF, will check it out after its run for a bit!

Issue #2606, opened 2019-07-03
Add-A-Listing Card


micahalcorn:
Encourage users to create listings by Injecting a call-to-action in the listings gallery on the home page in the place where the sixth listing would otherwise appear. See https://zpl.io/VOdAzKl for a mockup.

PR #2630, opened 2019-07-04
Make Drawers swipable and add static CTA listing cards

### LINKED ISSUE #2423: Floating Footer
[This element](https://app.zeplin.io/project/59fa2311bac7acbc8d953da9/screen/5ce5c8b32eb4af5c6e30eb9e) should float in the bottom right corner of the screen and lead to [an expansion of the footer](https://zpl.io/bWjqwWj) with links to Settings and Feedback as well as language and currency selects.

### LINKED ISSUE #2605: Origin Rewards Card
Inject an Origin Rewards call-to-action in the listings gallery on the home page in the place where the third listing card would otherwise appear. See https://zpl.io/2GGeL94 for a mockup. If the user has signed up, this should point to the campaign dashboard. Otherwise, it should point to [/welcome](https://dapp.originprotocol.com/#/welcome).

### LINKED ISSUE #2635: DApp onboarding email validation regex is too restrictive
It requires TLDs of length 2-4 which prevents emails with a lot of the new TLDs. Unsure about other possible issues with it but it should be looked at.

shahthepro:
# Description
- Add the slide in/out animation for navigation drawers
- ~~Allow to swipe to close the drawers~~
- Add the static "Earn Tokens" and "Create Listing" card (Closes #2605 and #2606)
- Update email validation regex to be less restrictive
- Disallow dismissing `WaitForTransaction` modal
- Add Fallback for listings created before deploying a proxy.
- Exclude current listing from "Other listings by seller" section
- Add Floating footer (Closes #2423 )
- Supports email subdomains of two or more characters (Closes #2635)


nick:
@micahalcorn do we want to hide the CTAs from desktop as we already have the links in the navbar?

image

nick:
@shahthepro can we extract the swipeable menu stuff into a separate PR? From testing it seems the responsiveness of the swipe isn't ideal. There might be a better approach we can take here.

ezgif com-resize

shahthepro:
@nick Have reverted that change in this PR.

dapps/marketplace/src/hoc/withTokenBalance.js
```diff
@@ -11,7 +11,9 @@ function withTokenBalance(WrappedComponent) {
       variables={{ account: wallet, token }}
       skip={!wallet}
     >
-      {({ data }) => <WrappedComponent {...props} {...tokenInfo(data)} />}
+      {({ data }) => (
+        <WrappedComponent wallet={wallet} {...props} {...tokenInfo(data)} />
```
nick: Why do we need this extra prop? Seems unrelated to what this hoc is doing?

dapps/marketplace/src/pages/App.js
```diff
@@ -181,6 +181,10 @@ class App extends Component {
 export default withIsMobile(withWeb3(withCreatorConfig(withRouter(App))))
 
 require('react-styl')(`
+  html, body
```
nick: Do we need this? What's it for?

dapps/marketplace/src/pages/listings/ListingCards.js
```diff
@@ -21,59 +24,75 @@ const ListingCards = ({
   const [redirect, setRedirect] = useState()
   if (!listings) return null
 
+  const listingToRender = [
```
nick: Can we hide these on desktop for now?

dapps/marketplace/src/pages/listings/_EarnTokensCard.js
```diff
@@ -0,0 +1,91 @@
...
+                data && data.enrollmentStatus === 'Enrolled'
+                  ? '/campaigns'
+                  : '/welcome'
+              }
+            >
+              <div className="header">
+                <h5>
+                  <fbt desc="listingCard.earnTokens">
+                    Earn
+                    <span>ORIGIN TOKENS</span>
```
nick: Can we capitalize with CSS

dapps/marketplace/src/hoc/withTokenBalance.js
shahthepro: `ListingDetail` component uses `withWallet` HOC but the `wallet` prop is not passed down when this HOC is used since ES6 destructing is used to extract the `wallet` and `token` from other props.

I needed both `wallet` and `walletProxy` props inside `ListingDetail` component. In old listings created without a proxy, we also have to use `wallet` prop to check if the current user created the listing in addition to `walletProxy`.

dapps/marketplace/src/pages/App.js
shahthepro: Added that as fix to make scrolling/swiping a bit smooth on iOS webviews when using `position: fixed`. Since swipe to close is reverted, I can revert this too.

dapps/marketplace/src/pages/listings/ListingCards.js
shahthepro: I guess @micahalcorn wanted them to be on Desktop also.

@micahalcorn Any thoughts on this?

shahthepro: @nick @micahalcorn Have disabled this everywhere other than home page on Mobile for now. Let me know if this has to be enabled in other places.

micahalcorn:
Hmm good question @nick. I suppose we should take those links out of the desktop navbar. 😂

dapps/marketplace/src/hoc/withTokenBalance.js
nick: Ah makes sense

micahalcorn:
@shahthepro @nick Aure's preference is to leave both the promo cards and the nav links intact for desktop. 🖥

PR #2634, opened 2019-07-04
Mo' DApp Cleanup

micahalcorn:
- Removes the mobile app CTA from listing creation when in our app
- Uses the aspect ratio that we recommend to users (4:3) on desktop
- Makes appliances listing creation use the hourly fractional steps
- Causes the strenth meter to link to the profile screen
- Adds a link to the test builds buried in developer mode settings

[Deployed to test build](https://originprotocol.github.io/test-builds/micah/#/) ☑️

### DIFF
--- a/dapps/marketplace/src/components/DownloadApp.js
+++ b/dapps/marketplace/src/components/DownloadApp.js
@@ -1,13 +1,7 @@
 import React from 'react'
 import { fbt } from 'fbt-runtime'
 
-import withWallet from 'hoc/withWallet'
-
-const DownloadApp = ({ walletType }) => {
-  if (walletType === 'Origin Wallet') {
-    return null
-  }
-
+const DownloadApp = () => {
   return (
     <div className="onboard-help origin-wallet">
       <div className="app-image mb-3">
@@ -35,7 +29,7 @@ const DownloadApp = ({ walletType }) => {
   )
 }
 
-export default withWallet(DownloadApp)
+export default DownloadApp
 
 require('react-styl')(`
   .onboard-help.origin-wallet
--- a/dapps/marketplace/src/pages/create-listing/ChooseCategory.js
+++ b/dapps/marketplace/src/pages/create-listing/ChooseCategory.js
@@ -6,6 +6,7 @@ import Categories from '@origin/graphql/src/constants/Categories'
 import Link from 'components/Link'
 import Redirect from 'components/Redirect'
 import DownloadApp from 'components/DownloadApp'
+import withIsMobile from 'hoc/withIsMobile'
 
 const CategoriesEnum = require('Categories$FbtEnum') // Localized category names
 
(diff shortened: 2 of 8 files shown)


dapps/marketplace/src/components/DownloadApp.js
```diff
@@ -1,13 +1,7 @@
 import React from 'react'
 import { fbt } from 'fbt-runtime'
 
-import withWallet from 'hoc/withWallet'
-
-const DownloadApp = ({ walletType }) => {
-  if (walletType === 'Origin Wallet') {
-    return null
```
micahalcorn: This didn't seem to be working. 😕

dapps/marketplace/src/pages/create-listing/ChooseCategory.js
```diff
@@ -125,17 +133,19 @@ const ChooseListingCategory = ({ listing, prev, next, onChange }) => {
             </div>
           </div>
         </div>
-        <div className="col-md-4">
-          <div className="gray-box">
-            <DownloadApp />
+        {!isMobileApp && (
```
micahalcorn: I preferred this over checking inside of `DownloadApp` because we'll probably soon render something different when `isMobileApp`.

dapps/marketplace/src/pages/settings/Settings.js
```diff
@@ -384,6 +384,13 @@ class Settings extends Component {
                     </div>
                   </div>
                 </div>
+                {this.state.developerMode && (
+                  <div className="text-center test-builds">
+                    <a href="https://originprotocol.github.io/test-builds/">
+                      Test Builds
```
micahalcorn: This allows us to QA a branch inside of our mobile app. 🔵

dapps/marketplace/src/components/DownloadApp.js
shahthepro: Nice catch. Seems like it wasn't returning `Origin Wallet` but `Mobile` if mobile bridge is enabled.

dapps/marketplace/src/pages/create-listing/ChooseCategory.js
shahthepro: `isMobileApp` will be true for all mobile wallets that are built using React Native, right?

https://github.com/OriginProtocol/origin/blob/master/dapps/marketplace/src/hoc/withIsMobile.js#L36

In that case, we should use `walletType === 'Mobile' || walletType === 'Origin Wallet'` unless we don't want this in other mobile DApp browsers/wallets.

dapps/marketplace/src/pages/listings/ListingCards.js
```diff
@@ -133,7 +133,7 @@ require('react-styl')(`
     cursor: pointer
 
     .main-pic
-      padding-top: 66.6%
+      padding-top: 75%
```
shahthepro: I have to update #2630 to keep this same ratio for those CTA listing cards before it gets merged.

PR #2657, opened 2019-07-08
[Relayer] Store gas price and Reverted tx status in DB

franckc:
### Description:
Store a couple additional pieces of data in the relayer_txn table, which should be helpful for monitoring and analytics purposes:
 - New "Reverted" status in case the EVM reverted the transaction
 - New gas_price column

### Checklist:


mikeshultz:
@franckc you have a relevant test failing, it looks like.

franckc:
It was indeed a real bug. I love those unit tests !
Pls take another look when you get a chance @mikeshultz

franckc:
Thanks for the super quick review.

I like keeping them migrations separate, this way they can be rolled back/forward independently if necessary in the future... So if it's ok will leave as is...

Issue #2631, opened 2019-07-04
Transaction error investigation


mikeshultz:
Franck saw a failed transaction and figure it should be looked at.

https://etherscan.io/tx/0x7b90a510928fb545ab83b4905bca7c44c5ef184f05ffa87ab9d86b2abc5f842c

    0x474fa4940000000000000000000000008ac16c08105de55a02e2b7462b1eec6085fa4d860000000000000000000000000000000000000000000000000000000000000080000000000000000000000000ef104523f3829b9f0b0f6d2d52f49f76f8655a9100000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000041795b99f3472de47bc780b86866e02ac0fca413108148d1e88fef933dedd6524350e45ef13538ad242156c271ce5e7b21c0df073743c50f6911f784a437de4dc31b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000

It's a tx to a user's proxy that does appear to have existed at the time of the tx.  The call was to `forward(address to, bytes sign, address signer, bytes data)`.  Decoded, the call arguments are: 

    to: 0x8ac16c08105de55a02e2b7462b1eec6085fa4d86 (IdentityEvents)
    sign: 795b99f3472de47bc780b86866e02ac0fca413108148d1e88fef933dedd6524350e45ef13538ad242156c271ce5e7b21c0df073743c50f6911f784a437de4dc31b
    signer: 0xef104523f3829b9f0b0f6d2d52f49f76f8655a91
    data: a30749df6f96a21e0a377a3ea466e68f3b675051d8543619869e94eb55a01f8d53786eae

The forwarded call decodes to:

    emitIdentityUpdated(QmVrFBfyoKoy85KLz3FAfBYqNuGZz58DBmpadCvLHxM47P)

I haven't tried to verify the signature yet.  A failed signature is my first guess as to why this TX failed.  Perhaps someone is just poking at it to see what it does?

Also curious how @franckc came across this?

cc @nick

franckc:
Thanks @mikeshultz for looking into this.

I noticed the issue by looking at the transactions from one of the relayer hot wallet account on etherscan: https://etherscan.io/txs?a=0x9629f3998d08a7bc479451be450e897e3240b2b2

Eyeballing the data, it looks like about 5% of transactions have this issue.

mikeshultz:
cc @DanielVF

DanielVF:
[...]

PR #2665, opened 2019-07-08
Relayer: One tx per proxy

mikeshultz:
### Description:

This makes sure the relayer only accepts one pending transaction per proxy contract.  It's a short-term solution to prevent bad nonces for #2631.

I'm not really sure if this gels with the behavior of the dapp with profile updates. @nick?

### Checklist:

### DIFF
--- a/infra/relayer/src/purse.js
+++ b/infra/relayer/src/purse.js
@@ -476,7 +476,7 @@ class Purse {
    */
   async hasPendingTo(proxy) {
     proxy = this.web3.utils.toChecksumAddress(proxy)
-    for (const txHash of this.transactionObjects) {
+    for (const txHash of Object.keys(this.transactionObjects)) {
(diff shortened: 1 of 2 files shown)


nick:
It's true the dapp would then try to get the user to pay gas... though we can force a hard error in this case. The dapp should never be in a state where a tx hasn't been mined already before attempting another tx though.

mikeshultz:
Does the dapp send using proxies as well?  That's the only way there'd be a nonce conflict.  If it falls back to their normal account it should go through fine.  Though it could still have unintended consequences that maybe the dapp should handle if it gets this error.

nick:
The dapp sends using proxies too, yea

mikeshultz:
Also, would it beneficial to the dapp if this returned the pending tx hash, or is this something that should already be known?

PR #2680, opened 2019-07-09
Fractional Listing Enchancements

shahthepro:
[Link to test build](https://originprotocol.github.io/test-builds/shah/#/)

Updates the Calendar components UX:
- User has to select both a check in and check out date. Check out date should be at least one day after the checkin date
- When one date is selected, only possible checkout dates are enabled. The rest will not be clickable.


micahalcorn:
Great work 👏 Here are some requests (most of which we have discussed):

- [ ] Allow me to select a prior date after selecting a start date.
- [ ] Make the available dates a darker text color.
- [ ] Put a rounded border on the current day.

shahthepro:
@micahalcorn I have made the changes you have listed. Also updated the test build.

micahalcorn:
I think this is ready for code review @nick. 🙂

PR #2722, opened 2019-07-12
Rewards: Lock rule as long as precondition rule is not unlocked

sparrowDom:
### Description:
- Create a template for august campaign.
- Add a social charing boilerplate rule code.
- Add a functionality that enables you to lock a rule by conditioning it with completion of another rule

### Checklist:


infra/growth/src/util/arrayUtils.js
```diff
@@ -0,0 +1,9 @@
+function concat(x, y) {
+  return x.concat(y)
+}
+
+function flatMap(f, xs) {
+  return xs.map(f).reduce(concat, [])
```
shahthepro: Just a note. We can probably skip the `.map` here.
```
xs.reduce((a, d) => [...a, ...f(d)], [])
```
But your approach looks pretty cool.

infra/growth/src/resources/rules.js
```diff
@@ -465,13 +504,33 @@ class BaseRule {
...
    */
-  async getStatus(ethAddress, events, currentUserLevel) {
+  async getStatus(ethAddress, events, currentUserLevel, rules) {
     // If the user hasn't reached the level the rule is in, status is Inactive.
     if (currentUserLevel < this.levelId) {
       return GrowthActionStatus.Inactive
     }
+    // If the rule has additional unlock conditions
+    else if (this.config.additionalLockConditions) {
+      this._validateLockConditions(rules, this.config.additionalLockConditions)
```
franckc: As opposed to calling _validateLockConditions here, could we call it in the constructor of the rule ? This way we wouldn't wait for getStatus to get called to find configuration issues.

infra/growth/src/resources/rules.js
```diff
@@ -231,10 +232,18 @@ class CampaignRules {
...
     const data = []
     for (let i = 0; i < this.numLevels; i++) {
       data.push(
-        ...(await this.levels[i].export(adapter, ethAddress, events, level))
+        ...(await this.levels[i].export(
+          adapter,
+          ethAddress,
+          events,
+          level,
+          allRules
```
franckc: Do we really need to pass this additional parameters ? Both Level and BaseRule have a "crules" field that gives access to all the rules. So the rules should be already accessible without this extra parameter.

infra/growth/src/resources/rules.js
```diff
@@ -868,4 +928,36 @@ class ReferralRule extends BaseRule {
   }
 }
 
+/**
+ * A rule that requires the purchase of a specific listing.
```
franckc: nit: fix comment

infra/growth/src/util/arrayUtils.js
```diff
@@ -0,0 +1,9 @@
...
+}
+
+module.exports = { concat, flatMap }
```
franckc: nit: no need to export concat.

infra/growth/src/util/arrayUtils.js
sparrowDom: Thanks! Skipping the concat function like suggested. 

About the `map` I kind of prefer keeping my version. While your works nicely and uses less code I think with `map` the code is a bit more readable.

infra/growth/src/util/arrayUtils.js
sparrowDom: tnx, incorporating @shahthepro's change the concat is removed

infra/growth/src/resources/rules.js
sparrowDom: 👍

infra/growth/src/resources/rules.js
sparrowDom: good point! Time to un-mess the method signatures

infra/growth/src/resources/rules.js
sparrowDom: Yup good point thanks. Moved to the constructor

sparrowDom:
Thanks @franckc  & @shahthepro for the CR. I haven't added any tests for the SocialSharing rule, since it is just boilerplate code ATM. 

There are tests for invalid locking configuration, and for checking that the additional lock conditions functionality is working. Both of those are at the end of `apollo-adapter-august.test.js`

All the CR stuff has been addressed so merging this for now.

PR #2729, opened 2019-07-15
EthGasStationProvider

mikeshultz:
### Description:

Adds EthGasStationProvider subprovider to @origin/web3-provider that pulls current safeLow gas price from ethgasstation.info
 
### Checklist:


packages/web3-provider/src/subproviders/EthGasStationProvider.js
```diff
@@ -0,0 +1,59 @@
...
+            data,
+            time: now
+          }
+
+          return end(null, data)
+        })
+        .catch(err => {
+          console.error('Error fetching results from ethgastation.info')
+          console.error(err)
+          return next()
```
nick: Do we want to fall back if ethgasstation is down for some reason?

packages/web3-provider/src/subproviders/EthGasStationProvider.js
```diff
@@ -0,0 +1,59 @@
...
+
+  async fetchData() {
+    const res = await fetch(GAS_STATION_URL)
+    if (res.status !== 200) {
+      throw new Error(`Fetch returned code ${res.status}`)
+    }
+    const jason = await res.json()
+    if (typeof jason[GAS_PRICE_KEY] !== 'undefined') {
+      // values come from EGS as tenths of gwei
+      return String(jason[GAS_PRICE_KEY] * 1e8)
```
nick: These numbers won't ever be so big as to need BN?

packages/web3-provider/src/subproviders/EthGasStationProvider.js
mikeshultz: `next()` should allow the following subprovider(s) to handle it, essentially skipping this one.  In this case, it'll be ThrottleRPCProvider, which will send it along to the JSON-RPC provider.

packages/web3-provider/src/subproviders/EthGasStationProvider.js
mikeshultz: If gas price is ever at 9 Ether(max JS int) we have other issues.  Think this should throw in that case?  I really don't foresee a near future where that could be real.

nick: Ah yea no need to if it's 9 Eth

mikeshultz:
This should probably not be enabled on networks other than mainnet...

packages/web3-provider/src/subproviders/EthGasStationProvider.js
franckc: nit: I had the same question as nick. So maybe add a comment explaining next will cause default web3 gas price implementation to be used. It was not directly obvious to me neither.

PR #2752, opened 2019-07-18
[Mobile] Use object rather than string for growth enrollment data

franckc:
### Description:

When a user enrolls in growth, the fingerprint data was sent by the mobile app as a string rather than an object. This caused the server to save it as a string rather than JSON in the growth_participant.data JSONB column in the DB.

Fixed the mobile side and put a temporary fix on the server side.
I'll either backfill the legacy data or adapt the scripts using it to deal with it.

I haven't tested yet on mobile but will attempt to do so now that Tom showed all of us how to start a mobile dev environment...

### Checklist:


franckc:
Was able to setup my mobile development environment and test it works...

PR #2758, opened 2019-07-19
[T3] Email and OTP authentication

franckc:
### Description:
Implements T3 authentication on client and server side.
Flow is as follow:
 - user enters their email
 - we send a confirmation code to that email
 - user enters email verification code
 - if user has not yet setup OTP, we show a setup screen with a QR code they need to scan in Google Authenticator
 - user enters OTP code
=> Once user completes both email and OTP verification, we considered them as logged in and set data in the session to indicate so.

On subsequent requests, the server checks the data in the session to determine if the user is logged in.

I also added a new user table to store the otp key, since that needs to persist across sessions.

I tested on my local environment and it seems to be functional. The UI is barebone but that can be tackled later.

### Checklist:


micahalcorn:
@franckc so the user won't get an email with a link to click containing a token? She will need to manually enter a code like we do for email attestation?

franckc:
@micahalcorn In this first implementation the user has to enter the code we sent by email, same as bridge attestation. But should not be too hard to implement a link in a subsequent iteration if we think it is a must-have.

micahalcorn:
Thanks @franckc 👍 The link might be preferable so that the code is not confused with the OTP and since Coinbase, et al, use the email link pattern.

franckc:
@micahalcorn sounds good. let's first get the T3 main functionalities working and then we can iterate on making UI and UX better...

infra/token-transfer-server/migrations/20190713051638-create-user.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('t3_user', {
+      email: {
+        type: Sequelize.STRING,
+        primaryKey: true,
```
tomlinton: What do you think about having the primary key being an auto incrementing ID? I think email here is not immutable, and it might cause some problems with joins to the other tables (e.g. grants).

franckc: Yes, I'll do that.

franckc: Done

PR #2759, opened 2019-07-20
Rewards for Social Sharing (Backend)

shahthepro:
### Description:
Adds backend support for social sharing rewards.

### Checklist:
- [x] Move GrowthEvent to separate package
- [x] Create GrowthEvent on successful verification
- [x] Add validation wherever possible
- [x] Add tests
- [ ] ~~[Wrap any new text/strings with `fbt`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#wrapping-text) for translation~~
- [ ] ~~If there are any new/changed strings to translate, [`npm run translate`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#integrating-translations)~~


infra/bridge/src/controllers/promotions.js
```diff
@@ -0,0 +1,119 @@
...
+      clearInterval(interval)
+      return res.status(200).status({
+        success: false,
+        errors: [`Couldn't verify`]
+      })
+    }
+  }, 1000)
+})
+
+// // TODO: Add request body validation
```
shahthepro: @franckc I added this code before checking out Activity API. Commented it out for now and the migration file is still there. Should I remove this or leave it so that we can probably use some of this later for social networks that don't support webhooks?

infra/bridge/src/controllers/promotions.js
```diff
@@ -0,0 +1,119 @@
...
+      success: false,
+      errors: [`Attestation missing`]
+    })
+  }
+
+  const redisKey = `${socialNetwork.toLowerCase()}/${type.toLowerCase()}/${
+    attestation.value
+  }`
+  let tries = 0
+  const interval = setInterval(async () => {
```
shahthepro: @franckc Can probably wrap this whole `setInterval` thing into a `Promise` and `await` it. But not sure if that is the right way. Any thoughts on this?

infra/bridge/src/app.js
```diff
@@ -43,7 +43,7 @@ app.use(bodyParser.urlencoded({ extended: true }))
 
 app.use(require('./controllers'))
 
-app.listen(5000, () => {
+app.listen(5000, async () => {
```
franckc: why does this need to be async ?

infra/bridge/src/controllers/hooks/index.js
```diff
@@ -0,0 +1,184 @@
...
+ * Twitter posts events to this endpoint
+ * Should always return 200 with no response
+ */
+router.post('/twitter', (req, res) => {
+  if (req.body.follow_events) {
+    // Follow event(s)
+    const events = req.body.follow_events
+    events.forEach(event => {
+      if (
+        event.target.screen_name ===
```
franckc: should screen_name be converted to lower case in the same way TWITTER_ORIGINPROTOCOL_USERNAME is ?

infra/bridge/src/controllers/hooks/index.js
```diff
@@ -0,0 +1,184 @@
...
+        process.env.TWITTER_ORIGINPROTOCOL_USERNAME.toLocaleLowerCase()
+      ) {
+        // Store the follower in redis for 60 minutes
```
franckc: typo: 30 min, per line 153 below

same comment online 171

infra/bridge/src/controllers/hooks/index.js
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * Twitter posts events to this endpoint
+ * Should always return 200 with no response
+ */
+router.post('/twitter', (req, res) => {
+  if (req.body.follow_events) {
+    // Follow event(s)
+    const events = req.body.follow_events
+    events.forEach(event => {
```
franckc: Note: we could use a Promise.all to parallelize the request to redis rather than handling them in a serial way. But if we do that we should make sure to not go over the max number of concurrent redis connections and may need chunking. So perhaps just add a TODO. If we start noticing the webhook takes a long time, we can optimize. We'll get very tiny volume initially so it will be fast enough how you have it currently.

Same comment for forEach statement on line 170.

infra/bridge/src/controllers/hooks/index.js
```diff
@@ -0,0 +1,184 @@
...
+        // Note: Only the latest tweet will be in redis for 60 minutes
+        redisClient.set(
+          `twitter/share/${event.user.id}`,
+          JSON.stringify(event),
+          'EX',
+          60 * 60 * 30
+        )
+      })
+  }
+
```
franckc: Might be useful to add a logger.info statement here with number of shares/follows that we added to redis. Always useful to have some logs to know what the process is doing.

infra/bridge/src/controllers/promotions.js
```diff
@@ -0,0 +1,119 @@
...
+const { getAsync } = require('../utils/redis')
+
+const Attestation = require('./../models/attestation')
+
+// TODO: Add request body validation
+router.post('/verify', async (req, res) => {
+  const { type, socialNetwork, identity } = req.body
+
+  const attestation = await Attestation.findOne({
+    ethAddress: identity,
```
franckc: I think attestation table stores eth_address in lower case. Should identity get converted to lower case ?

infra/bridge/src/controllers/promotions.js
```diff
@@ -0,0 +1,119 @@
...
+    method: socialNetwork
+  })
+
+  if (!attestation) {
+    return res.status(400).send({
+      success: false,
+      errors: [`Attestation missing`]
+    })
+  }
+
```
franckc: nit: should we have some defensive check to make sure type of either SHARE or FOLLOW ?

infra/bridge/src/controllers/promotions.js
franckc: I'd leave the code with a note explaining this is currently not being used.
If in a couple months we determined we won't need it, we can remove it.
As to the migration file, I'd remove it since otherwise it will get executed.
But perhaps have the table schema as a comment below ?

infra/bridge/src/controllers/promotions.js
franckc: One alternative that might be simpler to than setInterval is to define a tick() function like [this](https://github.com/OriginProtocol/origin/blob/master/infra/relayer/src/purse.js#L59).

Then have a loop like
do {
  ...
} while (retries < 300 && await tick())

infra/bridge/src/controllers/promotions.js
```diff
@@ -0,0 +1,119 @@
...
+      // TODO: Process event and Push to growth events
+      return res.status(200).send({
+        success: true
+      })
+    } else {
+      tries++
+    }
+
+    if (tries >= 300) {
+      // Request will timeout after 1000ms * 300 === 300s or 5mins
```
franckc: 5min is a long time to wait ? if the DApp calls the verification after the user posted the share/follow, really it should not take more than 1min for us to get the data via webhook from twitter ?

infra/bridge/src/controllers/promotions.js
```diff
@@ -0,0 +1,119 @@
...
+      clearInterval(interval)
+      return res.status(200).status({
```
franckc: nit: I would log.error here. so that we can know for looking at logs if verifications are failing.
And I would log.info when verification is successful.

infra/bridge/src/hooks/twitter.js
```diff
@@ -0,0 +1,172 @@
...
+    : process.env.HOST
+
+const HOOK_ENV = process.env.TWITTER_WEBHOOK_ENV || 'dev'
+
+/**
+ * List all registered webhooks
+ * @param {String} oAuthToken
+ * @param {String} oAuthAccessTokenSecret
+ */
+function getWebhooks(oAuthToken, oAuthAccessTokenSecret) {
```
franckc: should this be an async function since it returns a Promise. Same question for function createWebhook below.

infra/bridge/src/controllers/promotions.js
shahthepro: It should only take a few seconds. But the DApp may not know if the user closed the popover or not. I was thinking of calling the mutation immediately on clicking "Share/Follow" button.

Can reduce this time however.

infra/bridge/src/app.js
shahthepro: It doesn't have to be. Will revert it

infra/bridge/src/controllers/hooks/index.js
shahthepro: Yes, Should be done.

infra/bridge/src/controllers/hooks/index.js
shahthepro: Yep, It's 30 minutes. Nice catch.

infra/bridge/src/controllers/promotions.js
shahthepro: We can do that in request body validation. Will do that.

infra/bridge/src/hooks/twitter.js
shahthepro: It doesn't have to be, unless we are using await inside this function. I have some inconsistencies in async/await/returning Promise though. Will clean it up.

infra/bridge/migrations/20190715082720-add-verification.js
```diff
@@ -0,0 +1,47 @@
...
+      },
+      status: {
+        // VERIFYING = Pushed to the queue and polling is in progress to verify
+        // SHARED = Public account and everything is OK.
+        // FAILED = Account is protected/private, failed to verify
+        // UNSHARED = Is a public account, but couldn't verify after 'n' tries
+        type: Sequelize.ENUM('VERIFYING', 'SHARED', 'FAILED', 'UNSHARED'),
+        defaultValue: 'VERIFYING'
+      },
+      created_at: {
```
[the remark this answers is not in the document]
shahthepro: We can set `timestamps: true` in the model file. But we have to define the columns anyway in migration files.

shahthepro:
@franckc I'm working on adding a `GrowthEvent` when a verification is successful. I can do this in two ways:
1. Add `@origin/growth` as a dependency of `@origin/bridge`. But it seems that it will create a cyclic dependency. The only place where `@origin/growth` uses `@origin/bridge` is [inside a backfill script](https://github.com/OriginProtocol/origin/blob/master/infra/growth/src/scripts/oneoff/backfillIdentityCountry.js), which I assume is safe to delete now.

2. Use a messaging queue as a means of communication between `@origin/growth` and `@origin/bridge`. Perhaps, redis PubSub? 

Any suggestions on this?

franckc:
@shahthepro Let's keep things simple. The backfill script backfillIdentityCountry was a one-off and can be removed. 
Now, given that I suspect in the future we'll need to insert those GrowthEvent from various packages in our codebase and do not want to have all those packages having a dependency on the heavy growth package we could also try this alternative: define a new @origin/growth/event package that has the strict minimum needed to insert a row in the growth-event table. Use that package in bridge and other places where we don't need the full growth package but just need to insert a GrowthEvent row (for ex. in the event-listener).
It could be cleaner but is a little bit more work...

infra/bridge/src/controllers/promotions.js
```diff
@@ -0,0 +1,274 @@
...
+      if (
+        type === 'FOLLOW' ||
+        (type === 'SHARE' && JSON.parse(event).text === content)
+      ) {
+        await GrowthEvent.insert(
+          logger,
+          1,
+          identity,
+          PromotionEventToGrowthEvent[socialNetwork][type],
+          null,
```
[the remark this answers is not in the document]
shahthepro: Done

PR #2768, opened 2019-07-23
[T3] Transaction processing

franckc:
### Description:
- Added a new table "Transfer". It is used for 2 purposes: a) allows users to enqueue a transfer request and for it to be processed later by an offline job. b) tracks the different stage of a transaction when it gets submitted to the network.
- Added new script "execute_transfer" that loads pending requests and submits them to the network. We can schedule this job as a cron to run every hour or so.
 - Refactored the @origin/token package so that Token.credit now returns a txHash and a new method waitForTxConfirmation can be called to get the receipt after N confirmations. This allows to record the txHash in the transfer table before the receipt is confirmed, minimizing the chances to lose track of a tx in case the process is interrupted or crashes.
 - Refactored @origin/token Token constructor to simply take a networkId. Updated all callsites accordingly.
 - Refactored @origin/token test setup to user @origin/services rather than its own custom setup.

### Checklist:


franckc:
@tomlinton This is now ready for review. Sorry this PR got bigger than I wanted to due to some refactoring I did to simplify the @origin/token package...

franckc:
@mikeshultz Added you as a reviewer in case you have any feedback on the transaction piece. As I mentioned, for this first iteration I kept things simple and processed transactions serially so that I don't have to deal with noce tracking. If throughput becomes an issue then we can look into integrating with Purse...

infra/faucet/src/ogn.js
```diff
@@ -28,8 +33,14 @@ class OgnDistributor {
...
       const value = this.token.toNaturalUnit(NUM_TOKENS)
       const contractAddress = this.token.contractAddress(networkId)
-      const receipt = await this.token.credit(networkId, wallet, value)
-      const txHash = receipt.transactionHash
+      const txHash = await this.token.credit(wallet, value)
+      const { status } = await this.token.waitForTxConfirmation(txHash, {
+        numBlocks: NumBlockConfirmation,
+        timeoutSec: ConfirmationTimeoutSec
+      })
+      if (status !== 'confirmed') {
```
mikeshultz: I'm not sure this needs to be dealt with now, but you might want to consider what happens in production if the tx takes longer than 2 minutes to get mined.  For instance, will this record the error, assume it won't get mined(but it does), then retry later creating multiple transactions?

infra/faucet/src/ogn.js
```diff
@@ -6,6 +6,11 @@ const logger = require('./logger')
 // Credit 100 tokens per request.
 const NUM_TOKENS = 100
 
+// Number of block confirmations required for a transfer to be consider completed.
+const NumBlockConfirmation = 1
+// Wait up to 2min for a transaction to get confirmed
+const ConfirmationTimeoutSec = 20 * 60
```
mikeshultz: This would be 20 minutes.

infra/token-transfer-server/src/enums.js
```diff
@@ -0,0 +1,31 @@
...
+  // Transfer requested by user.
+  'Enqueued',
+  // Transfer transaction sent to the network. Waiting for confirmation.
+  'WaitingConfirmation',
+  // Transfer successfully executed by the network.
+  'Success',
+  // Transfer transaction failed.
+  'Failed',
+  // Transfer cancelled by the user.
+  'Cancelled'
```
tomlinton: Do you think we need an additional one called `Paused` or something similar, for if the transfer was paused by us if we want to do some manual verification?

infra/faucet/src/ogn.js
franckc: Currently error handling is poor on the OGN faucet. If transaction does not get mined within the timeout, there is no retry or rebroadcast or anything like that...
I added TODO - not a high priority since OGN faucet usage is extremely low.

mikeshultz: To be clear, my concern was if it **does** get mined, but the timeout still occurs.

infra/token-transfer-server/src/enums.js
franckc: Yes, good idea - it could be handy... Done.

infra/faucet/src/ogn.js
franckc: doh ! fixed

infra/faucet/src/ogn.js
franckc: Yes, makes sense. I added TODo to handle return of either error or timeout from the call to token.waitForTxConfirmation. Really if it gets usage we'll need to improve the faucet. It does not handle neither concurrent transactions (will re-use same nonce)...

Issue #2720, opened 2019-07-12
Spinner > "Loading..."


micahalcorn:
Replace instances of "Loading..." in [the marketplace DApp](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/src) (where visually appropriate) with the [`LoadingSpinner`](https://github.com/OriginProtocol/origin/blob/db6060c48fc3ba44513873ab6e371d8827fddced/dapps/admin/src/components/LoadingSpinner.js). 😑

rtman:
I'll have a look at this one to start.

rtman:
@micahalcorn  tom said that the linked [LoadingSpinner ](https://github.com/OriginProtocol/origin/blob/db6060c48fc3ba44513873ab6e371d8827fddced/dapps/admin/src/components/LoadingSpinner.js) above is not in the dependencies for marketplace. Is this [LoadingSpinner ](https://github.com/OriginProtocol/origin/blob/db6060c48fc3ba44513873ab6e371d8827fddced/dapps/marketplace/src/components/LoadingSpinner.js) useable as a replacement?

micahalcorn:
Ah, sorry, I linked that recklessly. Try [this one](https://github.com/OriginProtocol/origin/blob/db6060c48fc3ba44513873ab6e371d8827fddced/dapps/marketplace/src/components/LoadingSpinner.js).

rtman:
thanks @micahalcorn I'll get it implemented

micahalcorn:
Resolved by #2769 ✅

PR #2769, opened 2019-07-24
Loading to spinner #2720

rtman:
### Description:

#2720 Updated lingering text based loading indicators. Except in the situation where the indicator is inside a button, nav component or otherwise needs to be constrained by a container. The current `LoadSpinner` component is a full screen component. I think the rest of those text based loading indicators could be swapped out for a similar component that isn't full screen but would fit to the parent component.

### Checklist:

- [✅] Test your work and double-check to confirm that you didn't break anything
- [✅] [Wrap any new text/strings with `fbt`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#wrapping-text) for translation
- [✅] If there are any new/changed strings to translate, [`npm run translate`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#integrating-translations)
- [✅] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)


micahalcorn:
Thanks @rtman 🙏 Will you upload this branch as a [test build](https://github.com/OriginProtocol/test-builds) so that I can run it without jumping through a bunch of hoops?

I was hoping to get those contained "Loading..."s swapped out too, but that can be a different PR if this one gets us some easy wins.

dapps/marketplace/src/components/Reviews.js
```diff
@@ -184,7 +186,7 @@ export default class Reviews extends Component {
                   }}
                 >
                   {loading ? (
-                    <fbt desc="reviews.loadingMore">Loading More...</fbt>
```
micahalcorn: It may be appropriate to run the translation commands to remove these now unused string instances. 🤔

dapps/marketplace/src/components/Reviews.js
```diff
@@ -13,6 +13,8 @@ import QueryError from 'components/QueryError'
 import query from 'queries/Reviews'
 import EthAddress from './EthAddress'
 
+import LoadingSpinner from 'components/LoadingSpinner'
```
micahalcorn: It may just be me, but I like to keep all the sibling `import`s grouped together, aka this one on line 11, perhaps. 😁

rtman: Definitely not just you, I'll fix that!

dapps/marketplace/src/components/Reviews.js
rtman: Ok I'll look into that.

rtman:
@micahalcorn I've uploaded a [test build](https://originprotocol.github.io/test-builds/loading-to-spinner) of the branch. Lmk if there's any issues.

And ya I think a separate PR for the remaining Loading indicators that this `LoadingSpinner` component won't work for is a good idea.  It'd have to not be `position: fixed` and be bound by the parent container.

dapps/marketplace/src/components/Reviews.js
rtman: I've updated the translations

rtman:
I notice that sometimes the loading spinner overlays the content and it doesn't look quite right. It seems to only be on the details page of a listing as far as I can tell. I'll have a look at why that is

micahalcorn:
@rtman can you also take care of this one and remove the massive, redundant heading on mobile? 🤦‍♂ Bonus points for fixing that black backgrounded text.

IMG_8033
IMG_8034

micahalcorn:
Review pagination is another case of where the spinner ends up on top.

IMG_8035

rtman:
Ok I'll have a look soon, one option for the overlaying issue is to put a solid background behind the spinner. So that at least the spinner doesn't appear to mix with the content beneath. But that might not work with all usages or the UI design.

rtman:
@micahalcorn I fixed up that notification loading indicator and the black background on the tx hashes. Also took a stab at some styling for the spinner. I added a raised white background to it, looks better imo but let me know what you think/want to do. Here's an updated [test build](https://originprotocol.github.io/test-builds/loading-to-spinner/#/).

micahalcorn:
Thanks @rtman. Does it work without the shadow? Most of our designs are flat, and the shadow unnecessarily complicates the predominate case where the space behind the spinner is empty. ⬜️

rtman:
@micahalcorn yup, I've removed the shadow and here is the updated [test build](https://originprotocol.github.io/test-builds/loading-to-spinner/#/)

micahalcorn:
@rtman one last thing 😇 Maybe some translucency? How about `opacity: 0.8;`? What do you think?

rtman:
@micahalcorn I think it looks pretty good with opacity. It's an easy fix if we decide to go back. Want me to commit this?
opacity_spinner

micahalcorn:
ship

rtman:
@micahalcorn  haha, how about 0.9 opacity? Looks a bit better to me, but up to you!
opacity_09

rtman:
@micahalcorn ok I've updated the opacity and padding on the spinner, [test build](https://originprotocol.github.io/test-builds/loading-to-spinner/#/) is updated if you want to have a look.

PR #2772, opened 2019-07-24
Configure dapp.originprotocol.com to redirect to shoporigin.com

tomlinton:
- Removes dapp.originprotocol.com from the dapp ingress and replaces it with shoporigin.com
- Adds another ingress for dapp.originprotocol.com with a permanent redirect to shoporigin.com

Deployed this on `prod` to make sure it works so it is already configured.


tomlinton:
Yea I saw that, it looks like the DNS was changed to `216.239.32.21`, not sure what that is, I thought it might be a 3rd party redirect service. @joshfraser?

tomlinton:
Forgot to note we can't configure it while it is pointed elsewhere so I won't handle it in this PR.

PR #2786, opened 2019-07-26
Use EthGasStationProvider on staging

mikeshultz:
### Description:

Use EthGasStationProvider on staging so it can be tested before hitting prod.

### Checklist:

### DIFF
--- a/packages/graphql/src/contracts.js
+++ b/packages/graphql/src/contracts.js
@@ -168,7 +168,7 @@ export function setNetwork(net, customConfig) {
       breakdownEvery,
       maxConcurrent,
       qps,
-      ethGasStation: net === 'mainnet' ? true : false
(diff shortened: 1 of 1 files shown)


packages/graphql/src/contracts.js
```diff
@@ -168,7 +168,7 @@ export function setNetwork(net, customConfig) {
       breakdownEvery,
       maxConcurrent,
       qps,
-      ethGasStation: net === 'mainnet' ? true : false
+      ethGasStation: ['mainnet', 'rinkeby'].indexOf(net) > -1 ? true : false
```
franckc: nit: a maybe slightly simpler syntax would be:
['mainnet', 'rinkeby'].includes(net)

mikeshultz: simplified!

PR #2794, opened 2019-07-29
Social Sharing Rewards (Frontend)

shahthepro:
Frontend changes of rewards for social sharing.

Also has a few minor backend changes. 

### Checklist:


dapps/marketplace/src/pages/growth/Promotions.js
```diff
@@ -0,0 +1,289 @@
...
+      return 'TWITTER'
+  }
+
+  return null
+}
+
+const getActionHash = action =>
+  web3.utils.sha3(`${action.type}/${action.content.post.text.default}`)
+
+class Promotions extends React.Component {
```
nick: Can we use a bunch of separate functional components here instead of a Component with lots of different renderX methods? Hoping we can move away from React.Component in new code

nick: If it's a lot of work to change then no need.. but bear in mind for the future

shahthepro: Thanks. Will do that.

shahthepro: @nick Does this look good? https://github.com/OriginProtocol/origin/commit/13bf837576f15dea4688f81fbfa4117538d1c892

sparrowDom:
- The styling of the back link seems to be a bit off. Caret not rotated correctly and the text not having enough margin above.
- Also is the "share this" button on desktop supposed to be this wide
Screenshot 2019-07-29 17 15 14

On mobile things look good though

sparrowDom:
Might be cool if we had a lock state for "Promote Origin" and "Follow Origin" when you are not able to share this on any of the channels:
Screenshot 2019-07-29 17 24 01
Screenshot 2019-07-29 17 24 06
Screenshot 2019-07-29 17 24 15

See how both of these look unlocked, but once you want to share the content all the social network channels (only Twitter for now) are locked

shahthepro:
@sparrowDom Thanks for CR. Will fix those styling in Desktop.

Regarding the lock state, I guess there is a lock state in Zeplin. Can add that. @micahalcorn Is this how it should be?

packages/graphql/src/configs/docker.js
```diff
@@ -25,6 +25,7 @@ const config = {
   performanceMode: false,
   graphql: `http://${HOST}:4002`,
   automine: 2000,
+  attestationIssuer: '0x5be37555816d258f5e316e0f84D59335DB2400B2',
```
sparrowDom: @shahthepro  I hope you don't mind me adding this commit to your branch

infra/bridge/src/utils/validation.js
```diff
@@ -19,6 +19,12 @@ const identityValidation = check('identity')
   .withMessage('Field identity must not be empty.')
   .trim()
 
+const identityProxyValidation = check('identityProxy')
```
franckc: Some legacy users may not have a proxy. Should that field be optional ?

infra/bridge/src/controllers/promotions.js
```diff
@@ -26,13 +29,21 @@ const waitFor = timeInMs =>
...
 
   const attestation = await Attestation.findOne({
     where: {
-      ethAddress: identity.toLowerCase(),
+      [Sequelize.Op.or]: [
+        {
+          ethAddress: identity.toLowerCase()
+        },
+        {
+          ethAddress: identityProxy.toLowerCase()
```
franckc: In some case there may not be a proxy ?
Update the query to handle that case. Maybe a IN clause could make it simpler. Something like this:
```
const addresses = [ identity.toLowerCase() ]
if (identityProxy) addresses.push(identityProxy.toLowerCase()
Attestation.findOne({ ethAddress: { [Sequelize.Op.in]: addresses } }, ...)
```

infra/bridge/src/utils/index.js
```diff
@@ -44,10 +44,32 @@ function getAbsoluteUrl(relativeUrl, params = {}) {
...
+  apos: "'",
+  lt: '<',
+  gt: '>',
+  quot: '"',
+  nbsp: '\xa0'
+}
+
+const htmlEntityPattern = /&([a-z]+);/gi
+
+function decodeHTML(content) {
```
franckc: nit: it's fairly trivial code but would be good to add a unit test... just in case :) Also helps with catching regressions in the future. I would use as input the tweet content from anna on [this doc](https://docs.google.com/spreadsheets/d/1akbpgFFy5vW-PszmcJNImIW2rjkiwDk4xVXlUO7fMzo/edit#gid=0).

infra/bridge/src/utils/validation.js
shahthepro: Will update it

infra/bridge/src/utils/index.js
shahthepro: Sounds good. Will add the tests.

PR #2796, opened 2019-07-29
Sparrow dom/add growth to config

sparrowDom:
### Description:

Adding growth server to **/#/settings** configuration page: 
Screenshot 2019-07-29 16 33 30

### Checklist:

- ~[ ] [Wrap any new text/strings with `fbt`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#wrapping-text) for translation~
- ~[ ] If there are any new/changed strings to translate, [`npm run translate`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#integrating-translations)~
- ~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)~

### DIFF
--- a/dapps/marketplace/src/pages/settings/Settings.js
+++ b/dapps/marketplace/src/pages/settings/Settings.js
@@ -19,6 +19,7 @@ const store = Store('sessionStorage')
 const configurableFields = [
   'bridge',
   'discovery',
+  'growth',
   'ipfsGateway',
   'ipfsRPC',
   'provider',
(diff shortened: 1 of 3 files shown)


dapps/marketplace/src/pages/settings/Settings.js
```diff
@@ -308,6 +309,22 @@ class Settings extends Component {
                           />
                         </div>
                       </div>
+                      <div className="form-group row">
+                        <div className="col-sm">
+                          <label htmlFor="indexing">
```
franckc: nit: should this label be changed to "growth" rather than indexing ?

shahthepro: Actually, there is no element with `id="indexing"` or `id="growth"` props as far as I can see. So this `htmlFor` prop has no effect

shahthepro: We can probably get rid of all `htmlFor` attributes or set an unique `id` for every input (and change its label's `htmlFor` attribute accordingly).

PR #2802, opened 2019-07-30
Puppeteer fix

nick:
Fix for puppeteer

### DIFF
--- a/dapps/marketplace/src/pages/create-listing/CreateListing.js
+++ b/dapps/marketplace/src/pages/create-listing/CreateListing.js
@@ -69,7 +69,7 @@ const CreateListing = props => {
   if (
     props.creatorConfigLoading ||
     props.walletLoading ||
-    !props.identityLoaded
+    props.identityLoading
   ) {
     return <LoadingSpinner />
   }


nick:
Most difficult bug of all time to find and fix 😬

cc @shahthepro

shahthepro:
@nick I added that `identityLoaded` prop everywhere. It relies on `networkStatus`. Wondering why it failed on tests randomly. Is it because the identity is injected (through localStorage??) during tests?

nick:
The issue was when the identity data was getting refreshed or reloaded, the networkStatus would briefly change from '7' (ready) to '2' (loading), then back to '7' again. This would cause the identityLoaded prop to switch briefly from false => true => false which would trigger the LoadingSpinner to be briefly rendered in `src/pages/create-listing/CreateListing.js`. That meant the CreateListing button would be un-rendered then rendered again which caused it to lose its state and not display its "listing successfully created" modal.

Pretty difficult to track down but got there in the end :-)

tomlinton:
Hah! Good find @nick. Not sure if you remember but I ran into the same thing with something else a while back. I could never figure out why it did that.

PR #2805, opened 2019-07-30
[SocialSharing] Populate identity with twitter profile data

franckc:
### Description:
Several changes:
 - bridge: store the twitter profile data in GrowthEvent.data rather than in Attestation.profileData at time of post/follow verification
 - rule engine: implement [Micah's formula](https://docs.google.com/spreadsheets/d/1fJ1BVNSDnWFqKh_L0zwxhhklGLalxLbGS1Ss3B1xfvg/edit?ts=5d3f7715#gid=0) based on account age, verified, num followers, last tweet age.
 - listener: populate Identity.data.twiterProfile based on data loaded from Attestation.profileData

### Checklist:


infra/bridge/src/controllers/promotions.js
```diff
@@ -68,57 +74,28 @@ const persistEvent = async ({
...
     await GrowthEvent.insert(
       logger,
-      1,
+      1, // insert a single entry
       identity,
       PromotionEventToGrowthEvent[socialNetwork][type],
-      contentHash,
-      event,
+      contentHash, // set customId to the content hash.
+      profileData, // store the profile data that contains the user's social stats in the GrowthEvent.data column.
```
shahthepro: Can we store both `event` and `profileData` here? Or persist the `event` somewhere else?

We do get stuffs like tweet/status id from the `event` that can be useful during verification before payout (We don't have search through the user's timeline again, we could just fetch that specific tweet and see it is not a 404). Also, it will be useful in debugging if anything goes wrong

infra/growth/src/resources/rules.js
```diff
@@ -793,15 +831,8 @@ class SocialShareRule extends SingleEventRule {
...
-    if (
-      !identity.data ||
-      !identity.data.twitterStats ||
-      typeof identity.data.twitterStats.numFollowers !== 'number' ||
-      typeof identity.data.twitterStats.accountAge !== 'number'
-    ) {
-      logger.error(
-        `Missing or invalid twitterStats in identity of user ${ethAddress}`
-      )
+    if (!identity.data || !identity.data.twitterProfile) {
```
shahthepro: Is `identity.data.twitterStats` renamed to `identity.data.twitterProfile` here? Or are both different things? 

If it is a rename, I still see [some references of `twitterStats` in this file](https://github.com/OriginProtocol/origin/blob/b8700a463d18179c51c93ccbdaa750d42e8d3a8f/infra/growth/src/resources/rules.js#L859-L867).

infra/bridge/src/controllers/promotions.js
franckc: Sounds good. It is duplicate data but could be useful for debugging so let's store it for now and we can always stop storing it in future if it takes too much space.

infra/growth/src/resources/rules.js
franckc: Yes indeed. Good catch. Fixed !

PR #2810, opened 2019-07-31
[SocialRewards] Fix tweet content bug

shahthepro:
- Twitter shortens and replaces all the text that it identifies as URLs. So, if we have a text like `ogn.dev/mobile`, it is shortened to `t.co/xxxxxx`. Twitter does give a mapping of shortened URL to original URL. But the original URL in the mapping prepends `http://`. So there will be a mismatch in content which will result in "Share" event not being resolved by `bridge`.

   The easiest way to overcome this will to manually prepend `http://` in growth campaign rules.

- Twitter also truncates the tweet content (if the content is longer than 140 chars, I guess) but also gives the full text along with a preview. Should prefer `event.extended_tweet.full_text` (if it exists) over `event.text`

### DIFF
--- a/infra/bridge/src/controllers/promotions.js
+++ b/infra/bridge/src/controllers/promotions.js
@@ -158,7 +158,7 @@ const isEventValid = ({ socialNetwork, type, event, content }) => {
   // It may result in a different content than expected, So always prepend URLs with `http://` in rule configs.
 
   const entities = (event.extended_tweet || event).entities
-  entities.entities.urls.forEach(entity => {
+  entities.urls.forEach(entity => {
     encodedContent = encodedContent.replace(entity.url, entity.expanded_url)
   })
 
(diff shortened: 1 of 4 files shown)


infra/bridge/src/controllers/promotions.js
```diff
@@ -144,11 +144,21 @@ const isEventValid = ({ socialNetwork, type, event, content }) => {
...
+    JSON.stringify(
+      event.extended_tweet ? event.extended_tweet.full_text : event.text
+    )
+  )
 
-  // Important: Twitter shortens and replaces URLs
+  // IMPORTANT: Twitter shortens and replaces URLs
   // we have revert that back to get the original content and to get the hash
-  event.entities.urls.forEach(entity => {
+  // IMPORTANT: Twitter prepends 'http://' if it idenitifies a text as URL
```
sparrowDom: Weird :) maybe it would be cool to add a test for this?

shahthepro: Yeah, but will do that in a separate PR

PR #2813, opened 2019-07-31
[SocialRewards] Use proxy to get identity

shahthepro:
This PR will the make the growth server to fetch the latest identity and also add support for proxy addresses

### DIFF
--- a/infra/growth/src/resources/rules.js
+++ b/infra/growth/src/resources/rules.js
@@ -827,9 +827,21 @@ class SocialShareRule extends SingleEventRule {
       // No user passed. Return zero.
       return reward
     }
+
+    // Load any proxy associated with the wallet address.
+    const ownerAddress = ethAddress.toLowerCase()
+    const proxies = await db.Proxy.findAll({ where: { ownerAddress } })
+
+    // Query events from wallet and proxy(ies).
+    const addresses = [ownerAddress, ...proxies.map(proxy => proxy.address)]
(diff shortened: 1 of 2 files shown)


infra/growth/src/resources/rules.js
```diff
@@ -827,9 +827,21 @@ class SocialShareRule extends SingleEventRule {
       // No user passed. Return zero.
       return reward
     }
+
+    // Load any proxy associated with the wallet address.
```
franckc: @shahthepro This should not hurt but theoretically it should not be needed since identities are stored in the DB by owner address. See the [listener code](https://github.com/OriginProtocol/origin/blob/master/infra/discovery/src/listener/handler_identity.js#L273).

And then the ethAddress passed to all rules methods should be the owner address as well since it comes from the auth token signed by the user, which has to be the wallet address (proxy can't sign).

Do you have an example of an address where identity was keyed by proxy and that required this fix ?

shahthepro: @franckc 

It happens in local for me. I updated the identity of `0xf17...` changing the name from "Stan James" to "Another World". It created another entry with the proxy of `0xf17...` (`0xba...`)

Screenshot 2019-07-31 at 9 21 52 PM

franckc:
@shahthepro hmm. this is a bug in the listener then... CC @mikeshultz who found a similar issue while running the consistency checker in prod. Let me open a separate issue to track this problem. In the meantime let's leave this workaround as is in the rules code.

PR #2815, opened 2019-07-31
Sparrow dom/growth social fixes3

sparrowDom:
### Description:
- fix a problem where all events were considered for share rewards and not only twitter share events
- Add August campaign fixture

### Checklist:

### DIFF
--- a/infra/growth/src/resources/rules.js
+++ b/infra/growth/src/resources/rules.js
@@ -812,6 +812,28 @@ class SocialShareRule extends SingleEventRule {
   }
 
   /**
+   * Loads the identity from the database either from main or proxy account
+   * @param {string} ethAddress
+   * @param {Object} identityForTest - For testing only.
+   * @returns {Promise<Object>}
+   */
+  async loadIdentity(ethAddress, identityForTest = null) {
+    // Load any proxy associated with the wallet address.
+    const ownerAddress = ethAddress.toLowerCase()
+    const proxies = await db.Proxy.findAll({ where: { ownerAddress } })
+
+    // Query events from wallet and proxy(ies).
+    const addresses = [ownerAddress, ...proxies.map(proxy => proxy.address)]
+    const whereClause = { ethAddress: { [Sequelize.Op.in]: addresses } }
+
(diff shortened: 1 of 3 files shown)


sparrowDom:
Merging this so it can be deployed on staging. We can fix possible problem in subsequent PRs

infra/growth/src/resources/rules.js
```diff
@@ -894,9 +894,12 @@ class SocialShareRule extends SingleEventRule {
    * @returns {Promise<Array<Reward>>}
    */
   async getEarnedRewards(ethAddress, events) {
+    const allowedTypes = ['SharedOnTwitter']
     const inScopeEvents = this._inScope(events)
     const numRewards = await this._numRewards(ethAddress, inScopeEvents)
-    const eventsForCalculation = events.slice(0, numRewards)
+    const eventsForCalculation = events
+      .filter(event => allowedTypes.includes(event.type))
```
franckc: Something is weird. This extra filtering should not be needed because _numRewards is supposed to do the right filtering.
It is supposed to call _tallyEvents which does have filtering based on eventTypes...
See [this code](https://github.com/OriginProtocol/origin/blob/master/infra/growth/src/resources/rules.js#L468)

Is it possible somehow this.eventTypes for SocialShare rule is either not set or not set to the right SharedOnTwitter value ?

sparrowDom: Thanks for pointing this out @franckc . Giving it a better look, I see my fix above is a bit bonkers.

The `_tallyEvents` works as it should and `_numRewards` has the correct value. The problem that causes the error is that eventsForCalculation get filtered of off all the events. Events in there are also of type `ProfileCreated`, `AirbnbAttestationPublished` .... And when we filter them we do get the right number of events, but the wrong type.

And the twitter reward calculation function `_getTwitterRewardsEarned` requires only the events of type SharedOnTwitter since those contain the twitter profile information.

I have a better fix for it here where filtering happens according to the event type defined in rules config:
https://github.com/OriginProtocol/origin/pull/2860/files

PR #2831, opened 2019-07-31
[SocialRewards] Revert workarounds and logs and update links in tweet contents

shahthepro:
- Revert workaround I did for demo and remove the logs that were added for debugging
- Update links in tweet contents to have the protocol


infra/bridge/src/controllers/promotions.js
```diff
@@ -176,8 +173,6 @@ const isEventValid = ({ socialNetwork, type, event, content }) => {
 router.post('/verify', verifyPromotions, async (req, res) => {
   const { type, socialNetwork, identity, identityProxy, content } = req.body
 
-  logger.info(`Will be polling ${type} event for ${identity} with "${content}"`)
```
franckc: I would leave all these logs on for now - might be helpful in the short-term for debugging ?
or alternatively switch them to logger.debug ?

PR #2852, opened 2019-08-02
consistency+cron=<3

mikeshultz:
### Description:

Integrating the consistency checker into cron, which will run once per day and E-mail the results.

#### New Env Vars

These will need to be defined when deployed with origin-cron

- `EMAILS_TO` - A comma-delimited list of E-mail addresses
- `SENDGRID_API_KEY`
- `SENDGRID_FROM_EMAIL`
- `NETWORK`

### Checklist:


infra/cron/src/scheduler.js
```diff
@@ -100,3 +101,13 @@ growthUpdateCampaignsQueue.add(
   { repeat: { cron: '30 20 * * *' } }
 )
 logger.info('Scheduled growthUpdateCampaigns job.')
+
+// Growth campaign update job. Runs daily at 20:30UTC (~12:30 PST).
```
franckc: Typo: it's the consistency checker, not campaign update job. Also time is not correct. should be 23:30UTC

infra/cron/src/jobs/consistency.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  const logFile = tmp.fileSync()
+  const network = process.env.NETWORK || 'mainnet'
+
+  setNetwork(network)
+  if (!contractsContext.web3) {
+    throw new Error('web3 not initialized')
+  }
+  console.log(`Logging to file ${logFile.name}`)
```
franckc: nit: use logger rather than console

infra/cron/src/jobs/consistency.js
```diff
@@ -0,0 +1,100 @@
...
+  sendgridMail.setApiKey(process.env.SENDGRID_API_KEY)
+
+  // Assemble the E-mail
+  const to = process.env.EMAILS_TO.split(',').map(s => s.trim())
+  const dateString = todaysDateString()
+  const email = {
+    to,
+    from: process.env.SENDGRID_FROM_EMAIL,
+    subject: `Daily data consistency checker results (${dateString}) (${network})`,
+    text: `Attached are the results of the ${network} consistency checker run on ${dateString}`,
```
franckc: Might be good to put in the text some super high level stats. because I bet at some point we won't look at the attachment. We could also change the subject in case some discrepancy found rather than none.
All these are nice-to-have improvements but absolutely not blockers for rolling this out.

infra/cron/src/jobs/consistency.js
mikeshultz: Weird, I remember explicitly doing that.  Thanks

infra/cron/src/jobs/consistency.js
mikeshultz: Agreed, but think it should be done after this PR.  Added #2861.

mikeshultz:
Deployed to dev.  It's not configured on staging and prod so it'll error, but should be safe to deploy anyway.

PR #2855, opened 2019-08-02
Add support for different client for Webhooks

### LINKED ISSUE #2851: [Social Sharing] Twitter share/follow broken on dev and staging environment
Our corporate Twitter account is on the free tier and only allows us to run a single environment with subscription webhook enabled. We are using it for the bridge prod server and therefore we can't use it in dev and staging.

The impact is that we can't test tweet post/share verification on dev and staging.

Possible options:
- Upgrade to the paid tier for $4k / year. That will give us a max of 2 environments. So we could have staging and prod working. Seems a bit expensive...
 - We currently store the subscription data in redis. we could try to find a way to tunnel redis lookups made from the bridge on dev/prod to the redis prod environment
 - Store the subscription data in a central store accessible by dev/staging/prod. Maybe some GCP storage like memstore or a redis instance that we run separately and make accessible to dev/staging/prod bridge server

@tomlinton @mikeshultz @shahthepro Any thoughts ?

shahthepro:
A different OAuth client can be used for Webhooks. Should close #2851


shahthepro:
@franckc @sparrowDom Merging this and deploying to staging to enable webhooks. Let me know if you have any comments, will open another PR for that.

infra/bridge/src/hooks/twitter.js
```diff
@@ -8,8 +8,9 @@ const logger = require('./../logger')
 const oauth = new OAuth(
   'https://api.twitter.com/oauth/request_token',
   'https://api.twitter.com/oauth/access_token',
-  process.env.TWITTER_CONSUMER_KEY,
-  process.env.TWITTER_CONSUMER_SECRET,
+  process.env.TWITTER_WEBHOOKS_CONSUMER_KEY || process.env.TWITTER_CONSUMER_KEY,
```
sparrowDom: nit: there are a couple of places this check is performed. It might be more safe if there would be a couple of util functions e.g. `getTwitterConsumer[Key/Secret]` where this check would be done. Quickly you can change something in one place and forget to do it in another.

shahthepro: Yeah, that sounds good

Issue #2844, opened 2019-08-01
[Social sharing] Sharing does not work on Android Origin Wallet


franckc:
PR fixing this: https://github.com/OriginProtocol/origin/pull/2857

@sparrowDom Reported that sharing a tweet fails on Android when using Origin Wallet.
![Android bug](https://user-images.githubusercontent.com/945910/62309654-cbb22d00-b43c-11e9-9312-7c4a07ab3658.jpg)

sparrowDom:
A fix for this is located here: https://github.com/OriginProtocol/origin/pull/2857 . The downside is that we need to publish another version of Origin App and this bug will still be present on existing android applications.

sparrowDom:
Done via: https://github.com/OriginProtocol/origin/pull/2857

PR #2857, opened 2019-08-02
Sparrow dom/origin wallet twitter workaround

sparrowDom:
### Description:
Twitter does not display the share dialog with the user agent that is default to react Native webView. For that reason we replace that user agent with a working one

**Solves issue:**https://github.com/OriginProtocol/origin/issues/2844

### Checklist:

### DIFF
--- a/mobile/src/screens/marketplace.js
+++ b/mobile/src/screens/marketplace.js
@@ -54,7 +54,8 @@ class MarketplaceScreen extends Component {
       enablePullToRefresh: true,
       modals: [],
       fiatCurrency: CURRENCIES.find(c => c[0] === 'fiat-USD'),
-      transactionCardLoading: false
+      transactionCardLoading: false,
+      currentDomain: ''
     }
     if (Platform.OS === 'android') {
       // Configure swipe handler for back forward navigation on Android because
@@ -550,6 +551,13 @@ class MarketplaceScreen extends Component {
     updateExchangeRate(this.state.fiatCurrency[1], 'DAI')
   }
 
+  onWebViewNavigationStateChange = (state) => {
+    const urlMatch = state.url.match(/^https?\:\/\/([^\/?#]+)(?:[\/?#]|$)/i)
+    if (urlMatch && urlMatch[1]) {
+      this.setState({ currentDomain: urlMatch[1] })
+    }
+  }
+
   onWebViewLoad = async () => {
     // Check if a growth invie code needs to be set
     this.clipboardInviteCodeCheck()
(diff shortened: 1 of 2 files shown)


mobile/src/screens/marketplace.js
```diff
@@ -550,6 +551,13 @@ class MarketplaceScreen extends Component {
     updateExchangeRate(this.state.fiatCurrency[1], 'DAI')
   }
 
+  onWebViewNavigationStateChange = state => {
+    const urlMatch = state.url.match(/^https?:\/\/([^/?#]+)(?:[/?#]|$)/i)
```
shahthepro: nit: We could probably use [URL Web API](https://developer.mozilla.org/en-US/docs/Web/API/URL).

Something like `new URL(state.url).hostname`.

sparrowDom: A lot cleaner your way thanks!

mobile/src/screens/marketplace.js
```diff
@@ -550,6 +551,10 @@ class MarketplaceScreen extends Component {
     updateExchangeRate(this.state.fiatCurrency[1], 'DAI')
   }
 
+  onWebViewNavigationStateChange = state => {
+    this.setState({ currentDomain: new URL(state.url).hostname })
```
shahthepro: It is good to wrap this in `try { ... } catch { ... }` block, unless we are sure that `state.url` will always be a valid URL. `URL` will throw if `state.url` is a malformed URL

sparrowDom: This state is reported from the browser inside webView, so it is probably always going to be a valid URL...

Adding try catch to be extra sure. Defensive programming FTW!

shahthepro: Thanks @sparrowDom. Everything else looks good to me. But Tom may want to take a pass.

sparrowDom: Yup I am sure he will thanks @shahthepro

sparrowDom:
@tomlinton I didn't want to bump the Gradle build versions. This is a P1 bug so we might want to publish the app relatively soon. I can do it, I just might need the google console credentials I think

tomlinton:
Any idea why it didn't like the user agent that function returned? The function was written to try and return an approximation of a standard browser user agent. If twitter doesn't like it it probably means there is some issue there.

The next Android release is going to require quite a bit of testing due to some changes (and 64 bit support for the first time). I'll put a test build up over the weekend and let you know, it'd be great if you have physical devices to test it on.

sparrowDom:
aha big release then nice @tomlinton 

I can't find any official Twitter documentation for this. I think it is a defence mechanism on Twitter's side when they detect something that is not to their liking the running environment. I think it is to prevent possible password password stealing or tweet content manipulation without user knowing.

PR #2859, opened 2019-08-02
Google Analytics Improvements

### LINKED ISSUE #2319: Track search queries  via Google Analytics
We should make sure we can collect statistics regarding search queries using Google Analytics.

It's possible I missed it but when I logged in Google Analytics, I could not get data about search terms. Ideally we should configure our DApp so that the Behavior > Site Search section of Google Analytics gets automatically populated. See screenshot below.

Screen Shot 2019-05-21 at 10 44 12 AM

### LINKED ISSUE #2832: Google Analytics should not track development / ci tests
Currently our Google Analytics setup is recording events from testing / dev / CI Dapps. We probably want this to just record from the production dapp.

DanielVF:
Closes #2319 - We now pass query strings on through to Google Analytics. This allows us to track the user's search queries in Google Analytics.

Closes #2832 - We now do not track page views from localhost. Before we were recording every automated CI run into our stats, and likely every developer action, making a very lopsided view of the actual activity on our Dapp.

For example, the "Create Listing" page is heavily biased to tests and developers.

image

Finally, a bonus bug fix - we were always recording a users first page view on the Dapp as "/". This meant we didn't actually know which page a user actually arrived to the Dapp on.

### DIFF
--- a/dapps/marketplace/public/template.html
+++ b/dapps/marketplace/public/template.html
@@ -38,7 +38,6 @@
     }
     if (hostname !== 'localhost' && hostname !== '0.0.0.0') {
       gtag('js', new Date());
-      gtag('config', 'UA-106384880-2');
     }
   </script>
   <% _.forEach(htmlWebpackPlugin.files.js, function(file) { %><script src="<%- file %>"></script><% }); %>
--- a/dapps/marketplace/src/components/Analytics.js
+++ b/dapps/marketplace/src/components/Analytics.js
@@ -1,6 +1,9 @@
 import React, { Component } from 'react'
 import { withRouter } from 'react-router-dom'
 
+/**
+ * Records routing path changes to Google Analytics.
+ */
 class Analytics extends Component {
   constructor(props) {
     super(props)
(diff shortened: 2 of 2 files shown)


dapps/marketplace/public/template.html
```diff
@@ -38,7 +38,6 @@
     }
     if (hostname !== 'localhost' && hostname !== '0.0.0.0') {
       gtag('js', new Date());
-      gtag('config', 'UA-106384880-2');
```
DanielVF: This actually records a pageview, but it's doing it incorrectly since it's not taking into account the path of the page as used by react router. Instead it's always recording "/"

dapps/marketplace/src/components/Analytics.js
```diff
@@ -11,23 +14,35 @@ class Analytics extends Component {
...
+    clearTimeout(this._afterTitleTimeout)
+    this._afterTitleTimeout = setTimeout(() => {
+      const hostname = window.location.hostname
+      if (hostname === 'localhost' || hostname === '0.0.0.0') {
+        return // We don't want to record automated testing or developers
+      }
+      const gaTrackingId = process.env.GA_TRACKING_ID || 'UA-106384880-2'
+      const title = document.title
+      const location = this.props.location
+      const path = location.pathname + (location.search || '')
```
DanielVF: Added the query string to the path that we record.

Issue #2414, opened 2019-06-05
Sort Listings


micahalcorn:
Allow users to sort search results. The simplest example is by price (asc or desc). We can probably also easily add a sort by creation date. It's unclear what other use cases might be, but [here is a simple design for it](https://zpl.io/aNmRrDk).

franckc:
CC @DanielVF and @sparrowDom as a FYI

Listings use a variety of currencies so we'll need to figure out how to normalize into a single currency for doing the sorting. Hopefully similarly to scoring we could write a function using the Elastic "painless" language that would get invoked when doing the sorting and that could convert whatever currency the listing uses into USD or whatever primary currency we would use for sorting ?

DanielVF:
What makes this fun is that the currency conversions are "dynamic". For now, with a small number of listings and a small number of currencies, it's probably simplest to just pass in a bunch of conversion rates into the script used for scoring the search and use that to normalize the prices.

rtman:
@micahalcorn @DanielVF I Just spoke with @franckc and I'm going to be looking into implementing this 🚀

rtman:
@micahalcorn are there any designs for how this should show on desktop?

micahalcorn:
@rtman [this](https://zpl.io/2yqprOp) or whatever is Good Enough For Desktop™️.

PR #2873, opened 2019-08-05
Sort listings backend

rtman:
### Description:

#2414 I am implementing Sort by Price functionality in to the dapp, this PR is the backend only.

I used cryptogecko api to get the exchange rates via bridge server. These rates are then stored to redis to be pulled later when a listings search is performed. I added `sort: { target: '', order: ''}` to the listings [state](https://github.com/OriginProtocol/origin/blob/d622f02059b113a6849e9b5d6e862d52a864549b/dapps/marketplace/src/pages/listings/Listings.js#L49-L50)/query and all the relevant graphql schemas/resolvers. If there is not a sort to perform, leave the `target` and `order` as empty strings. Otherwise `target: 'price.amount'` and `order: 'asc'` or `order: 'desc'` will perform sorting based on `price.amount`. It is designed to allow other future sorting options. But for now just price works, though I think we could add date pretty easily.

There are a few lingering questions that may be issues.

1.  token-DAI is not supported by coingecko at the moment, I chose this API because we could use just one API (fiat + tokens), it is free and it has no rate limiting. This DAI issue may not be a problem since DAI is pegged to 1 USD, could we not just use the USD exchange rate or hardcode a value of `'1'`? I realise it fluctuates a bit but it looks fairly insignificant, I could be wrong though.

2. In my testing of multiple currencies in the listings, I have edited the currencies in`packages/graphql/fixtures/_demoListings.js`. This worked to test sorting based on different currencies and I verified that the code is working in the dapp. But when docker-compose started acting up the dapp seemed to fallback to another source of listings. However the sorting was still be based on my multiple currencies source (verified in localhost:4000/graphql), resulting in some strange looking results in the dapp (the listings would come up as `currency: { id: fiat-USD}` but sort was still based on the currency I had added). I think this is just an issue with sync between the centralized and decentralized graphql servers caused by my editing _demoListings.js. It should still work but it's been difficult to figure this out. I think I'd have to speak to @nick or @DanielVF regarding this.

3. Should we move to a master list of currencies? I feel like there should be one that gets imported to wherever currencies are used. I notice there is `dapps/marketplace/src/constants/Currencies.js`but that seems dapp specific, and I don't think you guys want to import that into discovery and bridge?

4. I removed the `idsOnly` option in `infra/discovery/src/lib/search.js` as it didn't actually seemed used? Only situation I can see that is using listingIds is as fallback in the listing query resolver (`packages/graphql/src/resolvers/marketplace/listings.js`). And the ids are only ever explicitly passed from [here](https://github.com/OriginProtocol/origin/blob/d622f02059b113a6849e9b5d6e862d52a864549b/dapps/marketplace/src/pages/listings/Listings.js#L137-L142) as far as I can tell.

Looking forward to feedback!

### Checklist:


micahalcorn:
Nice PR @rtman 👌 Let me know when you're ready for me to look at a test build. I'll defer to the others on the code.

franckc:
Thanks @rtman for sending this PR ! I will review tomorrow morning...

rtman:
@micahalcorn will do!. @franckc sounds good

dapps/marketplace/src/pages/listings/Listings.js
```diff
@@ -40,7 +40,14 @@ class Listings extends Component {
...
       first: 12,
       search: getStateFromQuery(props),
-      sort: 'featured'
+      // currently sort is activated through hardcoding the sort state
+      // order = 'asc' or 'desc'
+      // I have made it possible to sort by other values but its not tested
+      // target: 'price.amount' is the current goal
+      // this will be updated to be user set when the front end is completed
+      // graphql is expecting string values so to disable use empty strings
+      sort: { target: '', order: '' }
```
nick: Could we just have a var for 'sort' and another for 'order' instead of nesting those properties inside a 'sort' object?

dapps/marketplace/src/queries/Listings.js
```diff
@@ -5,7 +5,7 @@ export default gql`
   query AllListings(
     $first: Int
     $after: String
-    $sort: String
+    $sort: [ListingSortInput]
```
nick: Again perhaps we can just have a 'sort' and 'order' var of type string?

infra/discovery/src/lib/search.js
```diff
@@ -298,13 +336,53 @@ class Listing {
...
+                  }
+                },
+                order: order
+              }
+            }
+          default:
+            // this default is for non script based sorting, should satisfy most cases
+            return [
+              {
+                [target]: {
```
DanielVF: I wonder what kind of evil fun we can have here, with the ability to sort on any user chosen field in our data. Extract moderation scores?

dapps/marketplace/src/pages/listings/Listings.js
rtman: yea that is better, it makes the front end simpler as well. I'll change it.

dapps/marketplace/src/queries/Listings.js
rtman: will do!

infra/discovery/src/lib/search.js
rtman: There's a lot of possibilities, we could get quite deep here. There is also multi value based sorting, such as an array of different fields, or [applying modes (min,max,sum,average,median)](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#_sort_mode_option). This code could be easily expanded upon.

rtman: I'll add a whitelist and error if a value is entered outside of that.

infra/bridge/src/utils/exchange-rate.js
```diff
@@ -5,49 +5,77 @@ const request = require('superagent')
...
+  // however DAI is pegged to USD so maybe we could just use 1
+  // or call a different API to get it.
+  // 'DAI-USD',
+  'KRW-USD',
+  'SGD-USD',
+  'GBP-USD',
+  'EUR-USD',
+  'JPY-USD',
+  'CNY-USD',
+  'USD-USD'
```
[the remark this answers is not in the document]
rtman: Yea it was just part of the data pulled automatically so I included it to keep in convention. I can skip this and hardcode in 1 for the value.

infra/discovery/src/lib/search.js
```diff
@@ -298,13 +336,53 @@ class Listing {
       numberOfItems = 1000
     }
 
+    // sorting
+    const getSortQuery = () => {
+      // sort  returns as [Object null prototype] {target:, order: } not sure why
+      // the following code converts it to a regular object
+      const resolvedSort = JSON.parse(JSON.stringify(sort))[0]
```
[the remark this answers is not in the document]
rtman: Was able to remove this since nick suggested to move from a sort object to separate string values for sort and order.

infra/discovery/src/lib/search.js
```diff
@@ -298,13 +336,53 @@ class Listing {
...
+      const { target, order } = resolvedSort
```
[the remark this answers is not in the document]
rtman: As per daniels comment, I've added a whitelist for both variables and am not checking it.

infra/bridge/src/utils/exchange-rate.js
```diff
@@ -5,49 +5,77 @@ const request = require('superagent')
...
 /**
- * Fetch excahnge rate from remote URL
+ * Fetch exchange rate from remote URL
+ * @ param markets - Array | String - Can be single or multiple markets
  */
-async function fetchExchangeRate(market) {
-  const exchangeURL = `https://api.cryptonator.com/api/ticker/${market}`
+async function fetchExchangeRate(markets) {
+  try {
+    // const exchangeURL = `https://api.cryptonator.com/api/ticker/${market}`
```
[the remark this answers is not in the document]
rtman: The format is different, I'll add the comment re cryptonator.

infra/discovery/src/lib/search.js
```diff
@@ -342,7 +420,13 @@ class Listing {
         subCategory: hit._source.subCategory,
         description: hit._source.description,
         priceAmount: (hit._source.price || {}).amount,
-        priceCurrency: (hit._source.price || {}).currency
+        priceCurrency: (hit._source.price || {}).currency,
```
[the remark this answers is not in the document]
rtman: I would think so, but that actually breaks it if I do that. Not totally sure why. This code was originally here before I started so I'm following convention.

infra/discovery/src/lib/search.js
```diff
@@ -353,13 +437,10 @@ class Listing {
...
-    if (idsOnly) {
-      const listingIds = listings.map(listing => listing.id)
-      return { listingIds, stats }
-    } else {
-      return { listings, stats }
-    }
+    // debug
+    // console.log('listings - ', listings)
+    const listingIds = listings.map(listing => listing.id)
+    return { listingIds, listings, stats }
```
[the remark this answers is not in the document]
rtman: This is a question I'd like @nick or @DanielVF to weigh in on. The original [code ](https://github.com/OriginProtocol/origin/pull/2873/files/d622f02059b113a6849e9b5d6e862d52a864549b#diff-e8a9317173ba9fe5906263f858dbf137L15-27)only returned `listingIds` (hardcoded ) so I added `listings` to get the price and currency information. But the subsequent code didn't complain when it received listings instead of listingIds. So I feel like the answer is yes, we could just return listings. I'll wait till they weigh in though.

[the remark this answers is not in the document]
DanielVF: I'm good with returning just listings internally here.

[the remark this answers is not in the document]
nick: Yea, ultimately the dapp only queries for listing IDs but no harm in returning listings here.

infra/discovery/src/lib/search.js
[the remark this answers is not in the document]
DanielVF: The data in ElasticSearch roughly comes from our GraphQL data schema. On a listing, "price" is a Price object. Here's some definitions from the typeDefs/Common.js

```
  type Price {
    currency: CurrencyResult
    amount: String
  }
...
  interface Currency {
    id: ID!
    name: String
    code: String
    priceInUSD: Float
  }
```

So Price is an object. I've got the `|| {})` in there to prevent crashing the request getting the amount if for some reason the price object isn't there.

infra/bridge/src/utils/exchange-rate.js
```diff
@@ -67,13 +95,14 @@ async function getExchangeRate(market) {
...
     price = await fetchExchangeRate(market)
 
-    if (Number.isNaN(Number(price))) {
-      // API is also down, send back the fallback values
-      // FALLBACK_EXCHANGE_RATE_ETH_USD and FALLBACK_EXCHANGE_RATE_DAI_USD
-      return (
-        process.env[`FALLBACK_EXCHANGE_RATE_${market.replace('-', '_')}`] || 310
-      )
-    }
+    // Disabled Fallback values until we decide on a strategy
```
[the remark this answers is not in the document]
rtman: @franckc but where does the constant come from? Also does it need to be periodically updated? Eventually it could possibly be super out of date if I just set it to a price from today.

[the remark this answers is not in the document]
rtman: ok I'll just set the constants to a value from todays price.

infra/discovery/src/lib/search.js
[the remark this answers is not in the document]
rtman: I'll remove the listingIds then.

franckc:
@rtman Regarding constant as fallback, agreed they could get outdated if we use price as of today, but this is a very edge case scenario. I think it is ok to assume that we'll detect and fix relatively quickly (in a matter of a day) continuous failures to fetch exchange rates from coingecko. We haven't added it but it is our plan to add monitoring and alerting to the discovery server using Prometheus/Grafana.

rtman:
@franckc @DanielVF @nick I believe I've addressed all your concerns in this new version. Please have a look again! 😄

rtman:
@micahalcorn I have to wait for this PR to be merged to submit a test build with the front end changes. Will let you know.

franckc:
Thanks @rtman ! I will review by EOD...

infra/bridge/src/utils/exchange-rate.js
```diff
@@ -5,49 +5,80 @@ const request = require('superagent')
...
+async function pollExchangeRate(markets) {
   setTimeout(() => {
-    fetchExchangeRate(market).then(() => pollExchangeRate(market))
+    fetchExchangeRate(markets).then(() => pollExchangeRate(markets))
   }, process.env.EXCHANGE_RATE_POLL_INTERVAL || 30000)
 }
 
 /**
- * Fetch excahnge rate from remote URL
+ * Fetch exchange rate from coingecko
```
franckc: I grepped our code base to find out what component was using the bridge /utils/exchange-rate endpoint. It looks like the mobile app is.
See here: https://github.com/OriginProtocol/origin/blob/master/mobile/src/screens/marketplace.js#L564
Note that it queries for the exchange rate for ETH and DAI against the user's preferred fiat currency (e.g. USD, EUR, GBP, etc...)

So we need to support DAI exchange rates otherwise it will break the mobile app...

Either you could add some code to calculate those DAI exchange rates assuming thet DAI-USD is 1 or we stick to using cryptonator ?

infra/bridge/test/exchange-rate.js
```diff
@@ -27,28 +26,33 @@ describe('exchange rate poller', () => {
...
+      .reply(200, {
+        rates: {
+          btc: { name: 'Bitcoin', unit: 'BTC', value: 1.0, type: 'crypto' },
+          usd: { name: 'US Dollar', unit: '$', value: 11874.946, type: 'fiat' }
+        }
+      })
 
     const response = await request(app)
-      .get('/utils/exchange-rate?market=YYY-USD')
+      .get('/utils/exchange-rate?market=BTC-USD')
```
franckc: Since the mobile app queries DAI and ETH exchange rate, I'd suggest testing both for 'DAI-USD' and 'ETH-USD' here, in order to catch any regression in the future.

infra/discovery/src/lib/search.js
```diff
@@ -140,14 +173,25 @@ class Listing {
...
-   * @param {boolean} idsOnly - only returns listing Ids vs listing object.
    * @throws Throws an error if the search operation failed.
    * @returns A list of listings (can be empty).
    */
-  static async search(query, filters, numberOfItems, offset, idsOnly) {
+  static async search(query, sort, order, filters, numberOfItems, offset) {
+    const currencies = [
+      'token-ETH',
+      'token-DAI',
+      'fiat-JPY',
```
franckc: What about CNY and SGD ?

Let's not do it in this PR, but it seems there are many places where we define more or less the same currency list. Ideally we should have a single constant that we import everywhere... :)

infra/discovery/src/lib/search.js
```diff
@@ -298,13 +342,68 @@ class Listing {
       numberOfItems = 1000
     }
 
+    // sorting
```
franckc: super nit: remove this extraneous comment

infra/discovery/src/lib/search.js
```diff
@@ -298,13 +342,68 @@ class Listing {
...
+    /**
+     * Creates sort query
+     * @returns {Object} - Sort query for elastic search
+     */
+    const getSortQuery = () => {
+      const sortWhiteList = ['price.amount']
+      const orderWhiteList = ['asc', 'desc']
+      // if sort and order are set, BS  return a sort
```
franckc: nit: remove "BS" ?

infra/bridge/src/utils/exchange-rate.js
tomlinton: Awesome catch sir @franckc 🎉 As Frank said this has to support all the same functionality that was here before.

infra/discovery/src/lib/search.js
rtman: You're right I should add those currencies here, this was a temporary array as I was waiting for feedback on moving to a single source of truth for the currrency list. Happy to add them now and we can defer the currency list decision to another PR.

infra/discovery/src/lib/search.js
rtman: hah weird! Will do.

infra/bridge/src/utils/exchange-rate.js
rtman: Ah yes good catch! I had implemented the hardcoding of DAI-USD in search.js but I can move it here and store DAI-USD as '1' in redis. I'd prefer to stick to one API as they all return different formats and it makes for a lot messier code. Unless there's any objections to pegging DAI-USD to '1'?

infra/bridge/test/exchange-rate.js
rtman: I can do that, the reason I chose BTC is because all the values in the api call are returned in BTC value and I have to convert to USD. But I can add DAI and ETH to this test as well.

rtman:
@franckc I've updated the PR with the requested changes, please have a look!

franckc:
@rtman Regarding function parseAndSetRateData, is it going to work for the exchange rates that mobile may request ?
See https://github.com/OriginProtocol/origin/blob/master/mobile/src/screens/marketplace.js#L563
I think mobile may request conversion of any fiat currency (EUR, GBP, KRW, ...) against both DAI and ETH.
Perhaps double check by having a unit test that queries for example DAI-CNY and ETH-CNY, DAI-GBP, ETH-GBP, etc...

rtman:
@franckc I've updated the exchange-rate.js to accept conversions to different currencies (not just USD), as well as the relevant tests.

infra/bridge/src/utils/exchange-rate.js
```diff
@@ -5,49 +5,82 @@ const request = require('superagent')
...
-      }
+  } catch (e) {
+    logger.error(`Error getting ${markets} exchange rate:`, e)
+  }
+}
 
-      return price
-    })
-    .catch(e => logger.error(`Error getting ${market} exchange rate:`, e))
+function parseAndSetRateData(market, rates) {
```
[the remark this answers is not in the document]
rtman: @franckc It looks like it will work with the mobile requests.

The [mobile request format](https://github.com/OriginProtocol/origin/blob/master/mobile/src/utils/price.js#L30) matches the format of the new code.  The [original fetchExchangeRate code](https://github.com/OriginProtocol/origin/commit/f825e30d9b97c4d33a0e67458c59327a6802b67e#diff-8dc11c87ee00a865639a36105ce00c22R37-R48) returned a price string and [so does my change](https://github.com/OriginProtocol/origin/blob/master/infra/bridge/src/utils/exchange-rate.js#L81-L91). Should be good! Unless I'm missing something.

infra/bridge/src/utils/exchange-rate.js
rtman: @franckc I ended up adding a statement to use the USD value, instead of 1, whenever DAI is involved. Since it is pegged to USD this should be a better approach.

[All the exchange rates are calculated based on values returned from coingecko.](https://github.com/OriginProtocol/origin/blob/f2f2c0b6240b3e5550d601a5953fa3a7002ac3bc/infra/bridge/src/utils/exchange-rate.js#L67-L85) so this shouldn't be an issue.

rtman: Just noticed an error in my comment above the rate calculation (see the link above), the value is no longer calculated against just USD, its calculated against whatever the exchangeToSymbol is set as.

franckc:
Deployed to staging. Bridge is at https://bridge.staging.originprotocol.com and Discovery is at https://discovery.staging.originprotocol.com

franckc:
After the code got deployed to the staging discovery server, I tried to issue a search query with sorting by price but the server returned an error :(

To reproduce: go to https://discovery.staging.originprotocol.com/

Issue the following query:
```
{
  listings(
    searchQuery: "house"
    sort: "price.amount"
    order: "asc"
    page: { offset: 0, numberOfItems: 10 }
  ) {
    nodes {
      id
      price {
        currency
        amount
      }
      description
    }
  }
}
```

Looks like Elastic is failing when executing the painless script... not sure what the root cause is...

rtman:
Ok I'll have a look in the am

PR #2905, opened 2019-08-08
Add Telegram Attestation and Webhook

shahthepro:
Should add both telegram attestation and webhook controllers

### Known Issues
- ~~For some reasons, once I have authorized one account with one client, I cannot use the same client to authorize another account. Couldn't figure out why it happens though. Might be just a misconfiguration from my side~~ This is no longer valid, since we switched to Telegram login widget

### Checklist:


dapps/marketplace/src/components/AttestationBadges.js
```diff
@@ -116,6 +116,8 @@ require('react-styl')(`
         background-image: url('images/growth/github-icon.svg')
       &.linkedin
         background-image: url('images/growth/linkedin-icon.svg')
+      &.telegram
+        background-image: url('images/growth/website-icon.svg')
```
franckc: Replace with telegram-icon.svg

dapps/marketplace/src/components/Attestations.js
```diff
@@ -89,6 +89,10 @@ require('react-styl')(`
       background-image: url(images/growth/kakao-icon.svg)
       background-size: 100%
       border: 0
+    &.telegram
+      border-color: #6331dd
+      background-color: rgba(#6331dd, 75%)
+      background-image: url(images/identity/website-icon-light.svg)
```
franckc: Should be telegram

packages/graphql/src/resolvers/IdentityEvents.js
```diff
@@ -26,7 +26,8 @@ const attestationProgressPct = {
   kakao: 10,
   github: 10,
   linkedin: 10,
-  wechat: 10
+  wechat: 10,
```
franckc: The sum of all those is now 120% :)
Make each of those 100/12

shahthepro: @franckc This shouldn't be a problem, since we cap the strength at 100% https://github.com/OriginProtocol/origin/blob/a8861d8c85f632dbc6f7d2393c5b2f4839b5951d/packages/graphql/src/resolvers/IdentityEvents.js#L300-L306

shahthepro:
@nick @sparrowDom When you are available, Can you do a frontend review for this PR?

dapps/marketplace/src/components/TelegramLoginButton.js
```diff
@@ -0,0 +1,50 @@
...
+const TelegramLoginButton = ({ redirectURL, buttonText, className }) => {
+  let rootEl
+
+  useEffect(() => {
+    const script = document.createElement('script')
+    script.async = true
+    script.setAttribute('src', 'https://telegram.org/js/telegram-widget.js?7')
+    script.setAttribute(
+      'data-telegram-login',
+      process.env.TELEGRAM_BOT_USERNAME
```
sparrowDom: Can we return an error if this is not set?

dapps/marketplace/src/components/TelegramLoginButton.js
```diff
@@ -0,0 +1,50 @@
+import React, { useEffect, memo } from 'react'
+
+const TelegramLoginButton = ({ redirectURL, buttonText, className }) => {
+  let rootEl
+
+  useEffect(() => {
+    const script = document.createElement('script')
+    script.async = true
+    script.setAttribute('src', 'https://telegram.org/js/telegram-widget.js?7')
```
sparrowDom: curious what is the `?7` for?

dapps/marketplace/src/components/TelegramLoginButton.js
```diff
@@ -0,0 +1,50 @@
...
+    script.setAttribute('data-auth-url', redirectURL)
+    script.setAttribute('data-userpic', false)
+    script.setAttribute('data-size', 'large')
+
+    rootEl.appendChild(script)
+  }, [redirectURL])
+
+  return (
+    <div ref={el => (rootEl = el)} className="telegram-login-button">
+      <button className={className}>{buttonText}</button>
```
sparrowDom: lets detect an `onClick` here since if that function gets called user has most probably miss configured the telegram bot. We can display some sort of a warning about that

shahthepro: The iframe still renders on top of the button though. So onClick will never be called.

dapps/marketplace/src/components/TelegramLoginButton.js
shahthepro: I'm not sure either. That URL is from here: https://core.telegram.org/widgets/login

dapps/marketplace/src/components/TelegramLoginButton.js
shahthepro: Can do that

shahthepro:
@sparrowDom Thanks for the review. As I mentioned on Discord, You might have misconfigured the bot. I have shared a doc on how to configure it properly. Let me know if you run into any other issues.

dapps/marketplace/src/components/TelegramLoginButton.js
sparrowDom: aha I see... nvm then :)

dapps/marketplace/src/components/TelegramLoginButton.js
```diff
@@ -0,0 +1,58 @@
...
+  useEffect(() => {
+    if (
+      !process.env.TELEGRAM_BOT_USERNAME ||
+      process.env.TELEGRAM_BOT_USERNAME === 'origin_protocol_test_bot'
+    ) {
+      // origin_protocol_test_bot is the default value set in dapps/marketplace/webpack.config.js
+      console.error(`Create a bot and set TELEGRAM_BOT_USERNAME env variable`)
+    }
+
+    const script = document.createElement('script')
```
nick: Should we verify this element doesn't already exist to prevent multiple scripts being attached if the component is re-rendered?

shahthepro: I guess it shouldn't be a problem, since we are using `useEffect`. We want to render a different script element with different attributes when there is a change in `redirectURL`. 

The script file loading multiple times, when the component has unmounted and then mounted, shouldn't be a problem either. Since the script just injects an iframe, which gets removed when the component is unmounted, along with script element.

dapps/marketplace/README.md
```diff
@@ -20,3 +20,39 @@ To observe the tests running in the browser:
...
+    ```
+6. Start the DApp with the following command
+    ```
+    HOST=<YOUR IP ADDRESS> PORT=80 yarn start
+    ```
+
+### Additional step for Telegram Webhooks
+When you want to run the Telegram Webhooks for group join event verification, you should do the following. 
+1. Create a web tunnel for bridge server. [Refer `@origin/bridge`'s README.md](https://github.com/OriginProtocol/origin/blob/master/infra/bridge/README.md#setup-tunnel-for-webhooks-optional)
+2. Add the bot you created to the group as an admin.
```
sparrowDom: Very cool readme 👏 

Can you also add that you need to make the group public. And there should probably be some configuration somewhere where you specify the name of the group that dapp needs to open.

PR #2906, opened 2019-08-08
Sparrow dom/growth facebook share

sparrowDom:
### Description:
- add Facebook like and Facebook share to Origin Rewards
- handle simple social event confirmation on origin growth. Where a client is trusted to confirm the execution of the action
- in order to be able to group different social network shareable actions in the dapp. The content hash is no longer used to identify the selected action rather a custom content id
- on a selected promotion back button is changed to "back to promotions" instead of "back to campaigns" on desktop 

**Keep in mind:**
- this requires `growth-event` migrations to be run
- active august rewards rules need to be updated

**Resolves issue:** https://github.com/OriginProtocol/origin/issues/2907

### Checklist:


infra/growth/src/apollo/resolvers.js
```diff
@@ -189,6 +192,49 @@ const resolvers = {
...
+
+      // track growth events only for supported platforms
+      logger.warn(
+        `Not supported actionType: ${args.actionType} for social share`
+      )
+      return false
+    },
+    async confirmSocialFollow(_, args, context) {
+      requireEnrolledUser(context)
+      console.log('RECEIVED THE FOLLOWING', args)
```
franckc: remove this debug statement

packages/graphql/src/typeDefs/Growth.js
```diff
@@ -221,5 +224,9 @@ module.exports = `
     log(event: JSON!): Boolean
     # Remind a user that his invitation is still pending
     inviteRemind(invitationId: Int!): Boolean
+    # Confirm executing social share
+    confirmSocialShare(contentId: String!, actionType: GrowthActionType!): Boolean
```
franckc: super nit on the naming: this is not so much a "confirm" but more a "log" ?

infra/growth/campaigns/august.js
```diff
@@ -264,6 +268,8 @@ const augustConfig = {
             eventType: 'SharedOnTwitter',
             additionalLockConditions: ['TwitterAttestation'],
             content: {
+              // different social network shares sharing same content should have the same id
```
franckc: It was not immediately obvious to me what the id was used for.
I think I understand it is used for a) grouping in the DApp and b) to log event in case of a non-verifiable sharing. Couple suggestions:
 - explain this in the comment
 - add a unit test for august rules that checks that for rules with same id, the content is the same. I'm sure you defined it correctly in august but may help us catch discrepancies in rules config in future campaigns :)

franckc: Another thought would be to take the content definition out of the rules and move it up as a separate section in the config. So that it is defined once and does not need to be duplicated in every rule. A rule would only need to have the content.id in its config.
It would look like:
```
const augustConfig = {
  numLevels: 3,
  levels: {
    0: {
      rules: [
      ....
   }
   content: [
   {
     id: "origin_app",
     titleKey: ...
   },
   ....
   ]
```

Though the downside is that in the future the content may vary depending on the social network ? Like tweets are short and include hashtags while facebook posts can be longer and have no hashtags. So maybe that's not a good idea...

dapps/marketplace/src/components/growth/Action.js
```diff
@@ -104,13 +104,28 @@ function Action(props) {
     externalLink = `https://twitter.com/intent/tweet?text=${encodeURIComponent(
       getContentToShare(props.action, props.locale)
     )}`
+  } else if (type === 'FacebookShare') {
+    buttonLink = undefined
+    foregroundImgSrc = 'images/growth/facebook-icon.svg'
+    title = fbt('Share this on Facebook', 'RewardActions.postThisOnFacebook')
+    externalLink = [
```
shahthepro: @sparrowDom Just a thought, What do you think about moving these external links stuff to growth rules?

infra/growth/src/apollo/resolvers.js
sparrowDom: Thanks that one slipped the cracks

packages/graphql/src/typeDefs/Growth.js
sparrowDom: agreed that is more appropriate

dapps/marketplace/src/components/growth/Action.js
sparrowDom: interesting... thinking about it I would rather keep it here since I think this is implementation detail rather than configuration one. For example Origin Wallet depends on Twitter share dialog to match `twitter.com/intent/tweet` when it is intercepting the dialog on Android and triggering native share. 

I would rather see that changing that would require code changes on both sides instead of configuration on dapp/growth part and code in Origin Wallet

infra/growth/campaigns/august.js
sparrowDom: That's a good point. More social networks is going to bring only more redundancy in configuration. Though about it a bit... really like your idea with separate content config. So have changed it to that and separated the post contents from:
```
content: {
  id: "origin_app"
  ....
  post: {
    text: {...}
  }
}
```

to

```
content: {
  id: "origin_app"
  ....
  post: {
    tweet: {...}
  }
}
```
Since the current post config is really applicable only to Twitter. And in case we want to add other more generic sharing contents besides `tweet` like `short_text` or `text` that can be reused in multiple social networks we can.

Do you like this approach?

franckc: Ah yes, good thinking. I like your idea of having different fields if necessary in content, like tweet, post, etc.. Sounds good to me ! :)

PR #2917, opened 2019-08-08
Proxies and messaging

### LINKED ISSUE #2887: Messaging error - messaging enabled, but shows as not enabled for users with proxies
I have a user on my local that does not have messaging enable, yet the dapp thinks it is.  There's nothing in the registry, and the user was never prompted to enable.  

This issue is for investigation since I'm not sure how to repro at all.

mikeshultz:
### Description:

This fixes an issue where users that profiles of proxy contracts always show as they have not enabled messaging.  The dapp will now look up the proxy owner to find the `to` address.

The backend also now refuses to accept messages for proxy contracts.

closes #2887

### Checklist:


infra/messaging/src/index.js
```diff
@@ -77,6 +89,14 @@ app.get('/accounts/:address', async (req, res) => {
 
   address = Web3.utils.toChecksumAddress(address)
 
+  // If it's a proxy, make sure we're checking with the owner account address
+  if (await isContract(address)) {
```
sparrowDom: Would it make sense to put this kind of util functionality somewhere in a class. I expect this kind of `isContract` / `isProxyAddress` check will be  performed on many places.

mikeshultz: hm, actually, I might be able to pull these from the utils in `@origin/graphql` since that's now a dependency.  I keep forgetting about those.  Will see about that change this morning.  Thanks

mikeshultz: Moved `isContract()` to graphql utils

PR #2942, opened 2019-08-13
Fix linting in faucet

tomlinton:
Someone committing to master again? Oh hi @franckc 😁


franckc:
doh. totally my fault. not only I did not get a review but I broke stuff. terrible example :(

PR #2951, opened 2019-08-13
T3 server

tomlinton:
Bulk of the work for the T3 server. There is some more things to tidy up here but I thought we'd get this merged so I can hand off a few tasks to @franckc if he has some spare time. 😀 

- Restructured directories (separated controllers out of main app.js, organized tests)
- Added tests for all of the HTTP API methods
- Added/changed models to support requirements of client
- Refactored the login scheme to allow magic links
- Changed to vesting calculations in memory
- Moved all settings to a common config

Edit: also I cleared out the old models/migrations and started fresh, so before we merge this I'll need to clean up the db dev and staging.


infra/token-transfer-server/src/controllers/transfer.js
```diff
@@ -0,0 +1,92 @@
...
+const { ensureLoggedIn } = require('../lib/login')
+const {
+  asyncMiddleware,
+  isEthereumAddress,
+  getUnlockDate
+} = require('../utils')
+const { unlockDate, networkId } = require('../config')
+const { enqueueTransfer } = require('../lib/transfer')
+
+createProvider(networkId) // Ensure web3 credentials are set up
```
franckc: Remove - this does not seem to be used anywhere ?
And anyway I refactored the token package so this method no longer exist... :)

infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -18,47 +21,50 @@ const ConfirmationTimeoutSec = 20 * 60 * 60
...
+
+  const pendingOrCompleteAmount = grant.Transfers.reduce((total, transfer) => {
+    if (pendingOrCompleteTransfers.includes(transfer.status)) {
+      return (total += BigNumber(transfer.amount))
+    }
+    return total
+  }, 0)
+  logger.info('Pending or transferred tokens', pendingOrCompleteAmount)
+
+  const available = vested - pendingOrCompleteAmount
```
franckc: Should this use BigNumber.minus method rather than JS '-' operator ?
Same comment for the line below. Should use BigNumber.gt

infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -69,44 +75,23 @@ async function _checkTransferRequest(userId, grantId, amount, transfer = null) {
  * @param amount
  * @returns {Promise<integer>} Id of the transfer request.
  */
-async function enqueueTransfer(userId, grantId, address, amount, ip) {
-  await _checkTransferRequest(userId, grantId, amount)
+async function enqueueTransfer(grantId, address, amount) {
+  const grant = await _checkTransferRequest(grantId, amount)
```
franckc: nit: there is probably a potential race condition here in case someone would rapid fire requests to enqueueTransfer. We could serialize this call by adding a watchdog variable or locking the grant row in the DB. Alternatively we could let the script that processes the transaction double check (it should anyway) and refuse to execute the transaction.

infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -69,44 +75,23 @@ async function _checkTransferRequest(userId, grantId, amount, transfer = null) {
...
-  try {
-    transfer = await Transfer.create({
-      grantId,
-      userId,
-      status: enums.TransferStatuses.Enqueued,
-      toAddress: address.toLowerCase(),
-      amount,
-      currency: 'OGN' // For now we only support OGN.
-    })
-    await Event.create({
```
franckc: Why not logging this event ?

infra/token-transfer-server/src/lib/vesting.js
```diff
@@ -1,98 +1,108 @@
...
-      throw new Error(
-        `vested amount ${grant.vested} != calculated amount ${calculatedVested}`
-      )
-    }
-    if (grant.vested > grant.amount) {
-      throw new Error(
-        `vested ${grant.vested} > grant ${grant.amount} for ${grant.email} ${grantedAt}`
-      )
-    }
+  const cliffVestAmount = Math.floor(vestedPerEvent.times(cliffVestingCount))
```
franckc: Does Math.floor work on BigNumber objects ?
Maybe safer to use BigNumber rounding. Something like
`vestedPerEvent.times(cliffVestingCount).integerValue(BigNumber.ROUND_FLOOR)`

infra/token-transfer-server/src/lib/vesting.js
```diff
@@ -1,98 +1,108 @@
...
+  const remainingVestingCount = vestingEventCount - cliffVestingCount
```
franckc: Use BigNumber.minus

infra/token-transfer-server/src/lib/vesting.js
```diff
@@ -1,98 +1,108 @@
...
 
-    await grant.save()
-    await txn.commit()
-    return grant.id
-  } catch (e) {
-    await txn.rollback()
-    throw e
-  }
+  const vestingEvents = [cliffVestAmount, ...remainingVestingAmounts]
+  const roundingError = grant.amount - vestingEvents.reduce((a, b) => a + b, 0)
```
franckc: Use BigNumber.plus

infra/token-transfer-server/src/lib/vesting.js
```diff
@@ -1,98 +1,108 @@
...
+  if (start) {
+    bottomThreshold = start.diff(grant.cliff, grant.interval)
+  }
+  if (end) {
+    topThreshold = end.diff(grant.cliff, grant.interval)
   }
 
-  if (process.env.NODE_ENV !== 'test') {
-    console.log(`\n✅ Vested ${vestedCount} of ${grantCount} grants.`)
+  console.log(bottomThreshold)
```
franckc: nit: replace with logger.debug

infra/token-transfer-server/src/scripts/execute_transfers.js
```diff
@@ -1,153 +0,0 @@
-/*
```
franckc: Was this file deleted by mistake ? I was thinking we would schedule this script as a cron to execute the transfers enqueued... :)

infra/token-transfer-server/src/utils.js
```diff
@@ -0,0 +1,57 @@
...
+  return Account.findOne({
+    where: {
+      userId: req.user.id,
+      address: value
+    }
+  }).then(account => {
+    if (account) {
+      throw new Error('Address already exists')
+    }
+  })
```
franckc: super nit: return false rather than undefined
same comment for isExistingNickname function below.

infra/token-transfer-server/migrations/20181018102506-create-grant.js
```diff
@@ -1,41 +1,66 @@
...
+      purchase_date: {
+        type: Sequelize.DATEONLY
+      },
+      purchase_round: {
+        type: Sequelize.STRING
+      },
+      purchase_total: {
+        type: Sequelize.DECIMAL
+      },
+      investment_amount: {
```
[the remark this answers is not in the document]
tomlinton: Good question! @micahalcorn do you know the answer to this?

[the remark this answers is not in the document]
micahalcorn: I assumed they were all ETH or USD, but at least denominated in USD. 🇺🇸 I'll ask.

infra/token-transfer-server/src/scripts/execute_transfers.js
tomlinton: Yep will restore :grin:

infra/token-transfer-server/migrations/20190719230159-create-transfer.js
```diff
@@ -6,57 +6,55 @@ const tableName = 't3_transfer'
...
 module.exports = {
   up: (queryInterface, Sequelize) => {
-    return queryInterface.createTable(tableName, {
-      id: {
-        allowNull: false,
-        autoIncrement: true,
-        primaryKey: true,
-        type: Sequelize.INTEGER
-      },
-      user_id: {
```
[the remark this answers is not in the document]
tomlinton: I prefer the User -> Grant -> Transfer setup with a join for querying if you don't mind. I like to minimise redundancy in the database models. It'll be every so slightly slower to query but not noticeable with the amount of data we will have.

infra/token-transfer-server/migrations/20190719230159-create-transfer.js
```diff
@@ -6,57 +6,55 @@ const tableName = 't3_transfer'
...
-        allowNull: false,
-        type: Sequelize.DATE
-      },
-      updated_at: {
-        allowNull: false,
-        type: Sequelize.DATE
-      }
-    })
-      .then(() => queryInterface.addIndex(tableName, ['user_id']))
-      .then(() => queryInterface.addIndex(tableName, ['grant_id']))
```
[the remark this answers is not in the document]
tomlinton: Yep I should bring this back for grant_id

tomlinton:
Thanks for the review @franckc, will fix those issues up. I think I got test fever toward the end of this 😂

infra/token-transfer-server/src/lib/transfer.js
tomlinton: Ahh yea this is something I wanted to handle and add a good test for. Thanks for picking that up!

infra/token-transfer-server/src/utils.js
tomlinton: `true` I think you mean? express-validators only cares if it throws I think, `return false` would indicate it isn't valid.

infra/token-transfer-server/src/lib/transfer.js
tomlinton: Done!

PR #2955, opened 2019-08-13
ApolloEngine Support

mikeshultz:
### Description:

ApolloEngine support was added previously, but this just finishes it up.

Mostly, this adds breaking change checks to CircleCI.  I'm not really sure how well this works yet except through my tests, and there haven't been any schema changes yet, so we can remove this if it becomes a hindrance.

A lot of this is mostly trial right now.

I think those with project access should be able to see this: https://engine.apollographql.com/graph/origin-graphql/

### Checklist:


.circleci/config.yml
```diff
@@ -65,6 +66,34 @@ jobs:
...
+    steps:
+      - attach_workspace:
+          at: .
+      - run: sudo npm install -g apollo
+      - run:
+          name: Starting server
+          command: npm run start --prefix packages/graphql
+          background: true
+      - run: sleep 5
+      - run: apollo service:push --tag=master --endpoint=http://localhost:4002
```
franckc: Is it ok to push a new schema on every PR and every unit test run even if the schema hasn't changed ?

mikeshultz: The upload only currently happens on master.  If it makes sense, maybe we can add staging and stable later, but no, it doens't make sense to do the upload on every random branch.  See the filters:

https://github.com/OriginProtocol/origin/blob/mikeshultz/apollo-engine/.circleci/config.yml#L265-L268

The breaking change checks do, however, run on every branch just the same as any other test.

Issue #2970, opened 2019-08-14
White screen on Origin Wallet when doing Telegram Attestation


shahthepro:
When doing Telegram Attestation on Origin Wallet, switching back to the Origin Wallet from Telegram app results in a white screen of death. 

This might be [because of this line here](https://github.com/OriginProtocol/origin/blob/master/dapps/marketplace/src/pages/user/Profile.js#L297-L299). However, even refreshing the page doesn't load anything until you kill and restart the app.

micahalcorn:
It sounds like this is the same as #2983, yeah? 🏳️

shahthepro:
@micahalcorn Yeah. It is not specific to Telegram attestation.

PR #2972, opened 2019-08-14
Minor fixes for Telegram

shahthepro:
- Should remove the hack I did for #2970 since it seems to have no effect on staging and probably caused by other reasons.
- Change URLs to open Telegram app always
- Few styling changes


dapps/marketplace/src/components/growth/Action.js
```diff
@@ -134,7 +134,7 @@ function Action(props) {
     title = fbt('Join us on Telegram', 'RewardActions.followOnTelegram')
     allowInteractionWhenCompleted = true
     // TODO: Move screen name to Enviroment variable
-    externalLink = 'https://web.telegram.org/#/im?p=@originprotocol'
+    externalLink = 'tg://resolve?domain=@originprotocol'
```
franckc: Is this only used on mobile since I assume it won't work on mobile ?

infra/growth/campaigns/august.js
```diff
@@ -442,7 +442,13 @@ const augustConfig = {
             visible: true,
             nextLevelCondition: false,
             scope: 'campaign',
-            statusScope: 'user'
+            statusScope: 'user',
+            unlockConditionMsg: [
+              {
+                conditionTranslateKey: 'growth.attestation.requirement.telegram',
```
franckc: We need to add this string to dapps/marketplace/src/constants/Fbt#enum.js then run "yarn run translate"

dapps/marketplace/src/components/growth/Action.js
shahthepro: It does work on Desktop. But you have to have Telegram Desktop app installed. Perhaps we can have a fallback method that opens external window/tab in desktop

infra/growth/campaigns/august.js
shahthepro: Yes, Done.

dapps/marketplace/public/translations/de_DE.json
```diff
@@ -247,7 +248,7 @@
   "VUfcH": "Promote Origin",
   "9rfJX": "Requires: Twitter attestation",
   "1cVxL1": "Follow Origin",
-  "2AoW77": "Requires: Twitter attestation",
+  "49fgup": "Requires: Twitter and Telegram attestation",
```
franckc: Shouldn't this be "Twitter or Telegram attestation" ?
That string comes from here: https://github.com/OriginProtocol/origin/blob/master/dapps/marketplace/src/components/growth/ActionGroup.js#L48

Or maybe we don't need to lock the Follow section at all since it also has Facebook and no attestations are need for Facebook follow...

@sparrowDom Any input ?

PR #2982, opened 2019-08-15
Multi marketplace contracts

nick:
Support for multiple marketplace contracts in GraphQL and the DApp

- [x] Write tests for marketplace contract changes
- [x] Expand puppeteer tests to include transaction pages
- [x] Add puppeteer tests for reviews
- [x] Add puppeteer tests for notifications


nick:
Since this is almost ready to go, thought I'd start getting some feedback ☺️

packages/graphql/src/mutations/marketplace/createListing.js
```diff
@@ -70,6 +70,15 @@ async function createListing(_, input) {
   let tx
   const deposit = contracts.web3.utils.toWei(String(input.deposit), 'ether')
 
+  let version = input.version
+  if (!version) {
+    version = Object.keys(contracts.marketplaces).sort()[0]
```
franckc: Sorting is by ascendant order so this will use the contract with the oldest version. e.g. '000'
Shouldn't we default to using the most recent contract when creating new listings ?

packages/graphql/src/mutations/marketplace/deployMarketplace.js
```diff
@@ -1,9 +1,11 @@
-import Marketplace from '@origin/contracts/build/contracts/V00_Marketplace'
+import MarketplaceV0 from '@origin/contracts/build/contracts/V00_Marketplace'
+import MarketplaceV1 from '@origin/contracts/build/contracts/V01_Marketplace'
 import txHelper, { checkMetaMask } from '../_txHelper'
 import contracts, { setMarketplace } from '../../contracts'
-const data = Marketplace.bytecode
 
-async function deployMarketplace(_, { token, from, autoWhitelist }) {
+async function deployMarketplace(_, { token, from, autoWhitelist, version }) {
+  const Marketplace = version === '001' ? MarketplaceV1 : MarketplaceV0
```
franckc: nit: maybe throw an exception if version is unknown so that in future we don't deploy MarketplaceV0 again by mistake :)

packages/graphql/src/resolvers/marketplace/listings.js
```diff
@@ -98,6 +102,15 @@ async function resultsFromIds({ after, ids, first, totalCount, fields }) {
   }
 }
 
+async function getEvents(args) {
```
franckc: I think we currently have 5K events on marketplace contract... that's going to be lots of events.
Not sure if it makes sense to expose this without pagination. Perhaps add a comment for later ?

packages/graphql/src/resolvers/marketplace/listings.js
```diff
@@ -111,23 +124,24 @@ export async function listingsBySeller(
...
-    events = events.filter(
-      e => withdrawnListingIds.indexOf(e.returnValues.listingID) < 0
-    )
+  } else if (filter === 'inactive') {
+    events = await getEvents({ event: 'ListingWithdrawn', party })
   }
-  const ids = events.map(e => Number(e.returnValues.listingID)).reverse()
+
+  const ids = Object.keys(events)
+    .map(v => events[v].map(e => `999-${v}-${e.returnValues.listingID}`))
```
franckc: 999 should be replace with networkId

packages/graphql/src/resolvers/marketplace/listings.js
```diff
@@ -160,14 +174,17 @@ export default async function listings(
     // discovery server is up. They are ignored when running in decentralized mode
     // due to performance and implementation issues
 
-    const decentralizedResults = await allIds({ contract })
-    ids = decentralizedResults.ids
-    totalCount = decentralizedResults.totalCount
+    for (const version in contracts.marketplaces) {
+      const curContract = contracts.marketplaces[version].contract
+      const decentralizedResults = await allIds(curContract, version, '999')
```
franckc: Replace 999 with networkId

dapps/marketplace/src/pages/listing/mutations/Buy.js
```diff
@@ -360,12 +360,18 @@ class Buy extends Component {
 
     this.setState({ modal: true, waitForAllow: 'pending' })
 
+    let to = this.props.listing.contractAddr
```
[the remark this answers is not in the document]
nick: Not sure it's worth abstracting yet as it's only used twice... will do if we need to use it again though

packages/graphql/src/resolvers/marketplace/listings.js
nick: This is just a helper method that expects filters to be passed in `args`. Adding a comment...

Issue #2716, opened 2019-07-12
Account Not Loading From Messages Screen


micahalcorn:
Probably related to #2714, refreshing the conversations list (Messages screen) causes the identity to be lost. This results in the conversations never being found (other than the fake one from Origin Support).

micahalcorn:
This just happened to me on navigation from the home screen (rather than on direct page load via refresh). 😬

micahalcorn:
Actually, I haven't seen this happen in Coinbase Wallet, but I don't have a power user account there with a lot of messages. Could this be something specific to our webview➡️native connection, or more likely a graphql problem? cc @tomlinton

shahthepro:
@micahalcorn I fiddled around this bug. Seems like it happens on all mobile DApp browsers. The problem is that the identity loading takes some time on mainnet. Most likely due to missing cache. If you wait for a minute or so, the identity and messages load (but it doesn't happen always either. Perhaps because of timeout while looking for IPFS Hash?).

shahthepro:
@micahalcorn I had been checking out this issue again today. I remember @franckc telling me that @mikeshultz is working on PR that will push blockchain events to DB. If we plan to query these events from DB rather than hitting the blockchain, this delay should be drastically reduced.

mikeshultz:
For the sake of future decentralized mode(I'd really not like to diverge too far from that possibility if possible), I don't think waiting on EC DB is a good play here.  If the identity is that heavy to load, maybe we should be looking into splitting up the queries or something.  

I'm going to be bug hunting this week, let me know if you'd like me to take a deeper look at this.

shahthepro:
> For the sake of future decentralized mode(I'd really not like to diverge too far from that possibility if possible), I don't think waiting on EC DB is a good play here. If the identity is that heavy to load, maybe we should be looking into splitting up the queries or something.

Agree with that. It does work as of now. But takes some time (both identity and the conversations).

> I'm going to be bug hunting this week, let me know if you'd like me to take a deeper look at this.

Thanks. It'd be great if you can take a look at this. A different perspective is always welcome.

mikeshultz:
@micahalcorn, is this mobile app only?  If not, you got repro steps?

micahalcorn:
@mikeshultz I think @shahthepro said that it is.

mikeshultz:
[...]

PR #2986, opened 2019-08-16
Performance fix for identity card

mikeshultz:
### Description:

Basically, this pulls the unnecessary web3 cruft from the GraphQL `Identity` query so it doesn't use the local resolvers.  I don't really know how Apollo makes all its decisions, but apparently since the Web3 types are local-only, the entire query executes locally instead of remote.  I thought schema stitching was a bit more intelligent, but I suppose not in this case.

In this PR, I only added a new query(`SkinnyIdentity`) as opposed to rewriting the previous because I only wanted to clear (#2716).  However, this specific issue will probably impact other identity loads in other parts of the dapp.  On top of that, there are a lot of queries similar to this one that are run locally because they're crafted in such a way that forces Apollo to route them locally.  This is probably something that should be kept in mind in the future by dapp devs, and perhaps worthy of a project in the near future.

Performance mode is having relatively low impact right now because of it.  Though at least the heavyweights like `AllListings` seem to still be getting sent to the graphql server.

This also shamlessly includes a change to `@origin/services` to fix the graphql server port so the dapp can actually use performance mode with a local docker setup.

cc @micahalcorn 

### Checklist:


mikeshultz:
I'm assuming the failing marketplace tests are not related.  Or at least I can't figure how the profile nav card could have an effect on listing purchases.  /shrug

nick:
I think the tests are failing because graphql/src/configs/test.js has the graphql server pointing at port 4007.

I guess the problem with flaky tests is that we assume new changes aren't the cause when sometimes they are. I'll have another pass at fixing the flaky tests... especially since the Marketplace tests are arguably the most important as they test all parts of the system together.

Trying to remember why it was port 4007 instead of 4002 in services in the first place...

mikeshultz:
> I think the tests are failing because graphql/src/configs/test.js has the graphql server pointing at port 4007.

Doh, I'll change up that config as well for this PR.

> I guess the problem with flaky tests is that we assume new changes aren't the cause when sometimes they are.

Truth.  How did you ID this issue?  Another problem I have with these tests is I don't seem to get any visibility into them(errors, traces) to tell me what went wrong.  Occasionally I'll get more detail locally, but when I ran them locally I just got a different random selection of tests failing silently.

mikeshultz:
Fixed the config... there's still a graphql network error in the tests.  hrm.

mikeshultz:
No network errors on that second run, just a ton of timeouts.

nick:
Just lint now...

mikeshultz:
All green

PR #2993, opened 2019-08-16
IPFS port conflicts

mikeshultz:
### Description:

Multiple port conflicts with IPFS.  Not sure how the port `4001` conflict wasn't noticeable?  Port `4002`, seems maybe-maybe not depreciated.  It's explicitly used by the IPFS server in `@origin/services`, but no mention of it in the logs for `prod-ipfs-0`.  I could probably pull `4002/udp` out, but thought there was a chance it could be used or perhaps enabled if we wanted.

Do you have any insights/thoughts, @tomlinton?  Would deploying this be clean?

### Checklist:

### DIFF
--- a/dapps/marketplace/src/hoc/withSkinnyIdentity.js
+++ b/dapps/marketplace/src/hoc/withSkinnyIdentity.js
@@ -1,0 +1,28 @@
+import React from 'react'
+import { useQuery } from '@apollo/react-hooks'
+import get from 'lodash/get'
+
+import IdentityQuery from 'queries/SkinnyIdentity'
+
+function withSkinnyIdentity(WrappedComponent, walletProp = 'walletProxy') {
+  const WithIdentity = props => {
+    const id = get(props, walletProp)
+    const { data, networkStatus, refetch } = useQuery(IdentityQuery, {
+      skip: !id,
+      variables: { id }
+    })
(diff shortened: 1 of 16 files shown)


mikeshultz:
Just checked with `netstat` on `prod-ipfs-0` and IPFS is not currently listening on `4002`.

mikeshultz:
Removed `4002`from the lb and sts configuration.  Looks like it was removed form the default go-ipfs Dockerfile within the last 2 years.  Still think it's a good idea to move graphql off that port to reflect the local docker configuration, so leaving it.

mikeshultz:
graphql and ipfs are definitely running on the same cluster nodes...  Really not sure how kubernetes is avoiding this port `4001` conflict.  I guess maybe they're only exposed on the internal docker network without any host ports?  I still didn't think that was possible.

tomlinton:
Each pod has its own IP so the ports are not all bound on the same (node) IP. What you can't do is have two services in the same pod on the same port. This looks like it should work fine though, if we want to keep things in sync with docker-compose.

tomlinton:
Still keen to merge this?

mikeshultz:
I've been waiting for a slow time where I can watch it deploy to make sure nothing explodes, but if you want to do it, go for it, @tomlinton!

PR #2994, opened 2019-08-16
[Search] Handle malformed listings with no price

franckc:
### Description:

Some listings on staging are test data and are malformed. Their price.amount field is set to null.
It causes the sorting painless script to crash. This could also happen in prod due to bugs or malicious user writing directly listing data to IPFS rather than thru our code.

This PR adds some defensive code.

### Checklist:

### DIFF
--- a/infra/discovery/src/lib/search.js
+++ b/infra/discovery/src/lib/search.js
@@ -1,9 +1,8 @@
 const elasticsearch = require('elasticsearch')
+const get = require('lodash/get')
+
 const scoring = require('../lib/scoring')
 const logger = require('../listener/logger')
-/*
-  Module to interface with ElasticSearch.
- */
 
 const client = new elasticsearch.Client({
   hosts: [process.env.ELASTICSEARCH_HOST || 'elasticsearch:9200']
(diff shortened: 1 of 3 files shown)


franckc:
Would like to test this on staging so merging but happy to address any comment in a separate PR :)

infra/discovery/src/lib/search.js
```diff
@@ -373,9 +372,13 @@ class Listing {
                     type: 'number',
                     script: {
                       lang: 'painless',
+                      // Note: Script is very defensive at checking existence and validity
+                      // of the data to be robust to possibly malformed listings that are on testing
+                      // environments and could also exist in production.
                       source: `
                         if (params._source.containsKey("price") &&
                           params._source.price.containsKey("amount") &&
+                          params._source.price.amount != null &&
```
sparrowDom: The `Float.parseFloat` calls below [can throw](https://www.geeksforgeeks.org/float-parsefloat-method-in-java-with-examples/) `NumberFormatException` and `NullPointerException`.  Would be cool to wrap it in try-catch to also handle the possible wrong format exceptions.

Issue #2771, opened 2019-07-24
Out of gas errors purchasing fractional listing


tomlinton:
I got several out of gas errors attempting to purchase a fractional listing. I incremented the gas manually in metamask and eventually got it to work (it did take about 5 attempts though).

Additionally, once a purchase fails you can't attempt to purchase again as it immediately errors with a "See console" error and diplays "Expected event not found" in the console. This is probably because GraphQL gets confused by the failure.

<img width="1357" alt="Screen Shot 2019-07-24 at 2 20 19 PM" src="https://user-images.githubusercontent.com/489202/61760712-83818200-ae20-11e9-810d-a0e3e4b50bd4.png">

franckc:
From a contract perspective, what is different with fractional listings ? Looking at the contract code, it seems to be the same code path whether offer is for fractional vs traditional listing. So w

tomlinton:
Fractional might not be relevant here, it might be just listings in general.

nick:
Yea no difference between fractional and non-fractional at the contract level. We probably just need to bump the amount of gas up a bit here. Which value worked for you, Tom?

tomlinton:
Tried a few times with the default of `301785`, bumped it to `401785` and still failed, then succeeded with `1001785`. Probably not the most efficient. 😁

DanielVF:
@tomlinton Was this on staging? No relayer, I'm assuming? Was this using proxy contracts?

tomlinton:
It was mainnet, pretty fresh identity deployed with a proxy. Relayer doesn't get used for make offer calls afaik? You have to pay gas for ETH transfer anyway.

DanielVF:
@tomlinton Can you get me a eth TXN hash for that purchase? Or discord message me your ETH address.

tomlinton:
Looking back at the transactions now, it involves a DAI swap so the relayer would have been involved. The swap might be the cause here.

Failures:

[0xa4cf4abfe23822632ea9d2524e2aedc11bd6685da1ad811fcb2282c0afc3bfa9](https://etherscan.io/tx/0xa4cf4abfe23822632ea9d2524e2aedc11bd6685da1ad811fcb2282c0afc3bfa9)
[0x7620ff1942ec08d766034bbd0d701396e09eebb4300f17d0157bea9422ccf701](https://etherscan.io/tx/0x7620ff1942ec08d766034bbd0d701396e09eebb4300f17d0157bea9422ccf701)
[...]

PR #3004, opened 2019-08-17
Send extra ETH with DAI purchase to handle changing exchange rates

DanielVF:
Offers made with an atomic DAI swap were sometimes failing because the exchange price had changed between the transaction being created and transaction being finally mined. See issue 
#2771 

This PR sends along 1% extra ETH with DIA purchased. Any extra after the purchase is refunded to the proxy contract.

For now, a user can manually sweep it back the extra eth to the owner contract. In the future we may modify the proxy contract to automatically sweep it back.

### Checklist:


DanielVF:
It looks like the DAI address? is somehow not getting configured in the tests running on CI, though it works fine in tests locally.

I see the coin and exchange being created before the tests:

```
Deployed DAI stablecoin to 0x345cA3e014Aaf5dcA488057592ee47305D9B3e10
...
Deployed Uniswap Exchange Template to 0x38cF23C52Bb4B13F051Aec09580a2dE845a7FA35
Initialized Uniswap Factory
Created Uniswap Dai Exchange 0x486206CB5844197568feFe430Ef5275AD7246EF1
Approved DAI on Uniswap Dai Exchange
Added liquidity to Uniswap Dai Exchange
```

However, before the services are all setup, I also see:

```
Connecting to network test
Marketplace contract appears to be invalid!
IdentityEvents contract appears to be invalid!
ProxyFactory contract appears to be invalid!
ProxyImp contract appears to be invalid!
```

The when I try to create an offer using DAI, it fails:

```
  1) Marketplace
       DAI swap transaction
         should make an offer:
     Error: GraphQL error: Could not find token address for token-DAI
```

packages/graphql/package.json
```diff
@@ -43,6 +43,7 @@
     "apollo-link-schema": "^1.2.3",
     "apollo-link-ws": "^1.0.18",
     "apollo-server": "^2.6.7",
+    "bignumber.js": "9.0.0",
```
franckc: Just checking, did you run "lerna add bignumber.js --scope=@origin/graphql" ?
It should also update your yarn.lock which is a file I think we commit to the repo.
Though bignumber.js may already be a dependency of graphql since it is a web3 depends on it.

packages/graphql/src/mutations/uniswap/swapToToken.js
```diff
@@ -11,10 +12,20 @@ export async function swapToTokenTx(tokenValue) {
...
 
+  // We send 1% extra value to cover 99% of market price moves.
+  // See https://github.com/OriginProtocol/origin/issues/2771
+  // for how this amount was determined.
+  //
+  // Marketplace prices change on average every 10 blocks.
+  // It takes some time to finish a transaction, and the exchange will
+  // fail our transaction if the user did not send enough eth for
+  // the new price.
+  const value = new BigNumber(marketValue).times(1.01).toFixed(0) // have to be an integer.
```
franckc: I'm personally ok to use BigNumber since we are already using it in other places in our code on the back-end (bridge, growth, payout, ...).

@nick if you feel strongly we should use BN.js then the same result could be achieved by doing BN(marketValue).mul(101).div(100)

PR #2938, opened 2019-08-12
Sort listings frontend

rtman:
### Description:

This is the front end changes for the sort listings update #2414 . I have followed the zeplin screens that @micahalcorn linked in the [issue](https://github.com/OriginProtocol/origin/issues/2414). 

Marketplace tests are intermittently failing, I don't think that is related to this change though.

Currently there isn't a test build as the backend changes haven't been deployed, but I thought since I won't be contributing as much I should put this up.

### Checklist:

- [✅] Test your work and double-check to confirm that you didn't break anything
- [✅] [Wrap any new text/strings with `fbt`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#wrapping-text) for translation
- [✅] If there are any new/changed strings to translate, [`npm run translate`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#integrating-translations)

PR #3014, opened 2019-08-19
Fix dropdown width in Settings Page

shahthepro:
A very minor fix for the dropdown.

Screenshot 2019-08-19 at 3 24 23 PM

Probably broken when #2938 got merged [because of this line](https://github.com/OriginProtocol/origin/blob/master/dapps/marketplace/src/pages/listings/SortMenu.js#L206)


franckc:
My bad  for breaking the CSS... Thanks for fixing this @shahthepro ! :)

PR #3030, opened 2019-08-19
[Relayer] Reduce number of calls to getBalance

franckc:
### Description:

Reduce volume of calls to getBalance by checking balances of the master and children accounts only after a transaction is sent.

### Checklist:


infra/relayer/src/purse.js
```diff
@@ -413,6 +414,9 @@ class Purse {
       throw err
     }
 
+    // Trigger a check of the balances for every new transaction.
+    this.checkBalances = true
```
mikeshultz: I don't think this is the spot for it.  Or maybe not the only spot for it.  I think the best place to do the check would be after the tx is mined, probably by adding an extra callback like this: https://github.com/OriginProtocol/origin/blob/franck/purse_optim/infra/relayer/src/purse.js#L397-L399

Otherwise, it'll always be lagged by at least one transaction.

mikeshultz: Alternatively, you don't even need to add a callback, but just add the switch somewhere in this block after a receipt(any receipt) is found: https://github.com/OriginProtocol/origin/blob/franck/purse_optim/infra/relayer/src/purse.js#L1012-L1057

franckc: Would be ok to lag by one tx but you are absolutely right, it is better to move this trigger after tx is mined. Done !

franckc:
Looks like there is a Relayer test failure that is likely caused by this change... I'll take a look.

franckc:
Test should be fixed now.

PR #3035, opened 2019-08-20
Date package update

nick:
Remove date-fns dependency from marketplace dapp

### DIFF
--- a/dapps/admin/src/utils/formatDate.js
+++ b/dapps/admin/src/utils/formatDate.js
@@ -1,5 +1,5 @@
 import format from 'date-fns/format'
 
 export default function formatDate(timestamp) {
-  return format(new Date(timestamp * 1000), 'MM/DD/YYYY h:mm a')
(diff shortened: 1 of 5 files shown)


dapps/admin/src/utils/formatDate.js
```diff
@@ -1,5 +1,5 @@
 import format from 'date-fns/format'
 
 export default function formatDate(timestamp) {
-  return format(new Date(timestamp * 1000), 'MM/DD/YYYY h:mm a')
+  return format(new Date(timestamp * 1000), 'MM/dd/yyyy h:mm a')
```
shahthepro: Just wondering, why not use something like `dayjs.unix(timestamp).format('MM/DD/YYYY H:MM A')`?

nick: don't currently have dayjs in the admin dapp. I guess we can switch it for consistency

shahthepro: Got it. 👍

PR #3041, opened 2019-08-21
T3 server changes

tomlinton:
- Removed relationship between transfer and grants and replaced it with users because we don't want separate balances for each grant in the UI
- Separated out the enqueue of a transfer into an "add" and "confirm" step. The initial post request just adds it, and the confirm request enqueues it. The confirm request is how we enforce the 2fa requirement in the UI for a new transfer.
- Removed a bunch of unused dependencies.
- Added a fingerprint data function.
- Updated transfers to have fingerprint data in the transfer row rather than the event row so we don't need to query events to get that data (also because the data isn't relevant for a lot events, e.g. TRANSFER_FAILED, TRANSFER_DONE)
- Made the vesting logic use UTC so it is consistent client/server side
- Change to sequelize sync for cleanup
- More tests


infra/token-transfer-server/src/config.js
```diff
@@ -26,7 +26,9 @@ if (!sendgridFromEmail) {
   process.exit(1)
 }
 
-const unlockDate = moment.utc('2020-01-01')
+const unlockDate = process.env.UNLOCK_DATE
```
franckc: Wonder if it's dangerous to have a default in case we somehow misconfigure EnvKey. Should we rather crash if UNLOCK_DATE is not defined ?

tomlinton: 👍

infra/token-transfer-server/src/controllers/transfer.js
```diff
@@ -63,28 +61,62 @@ router.post(
...
           address,
           amount,
-          req.connection.remoteAddress
+          await getFingerprintData(req)
         )
       })
       // TODO: update to be more useful, e.g. users email
-      logger.info(`Grant ${grantId} transferred ${amount} OGN to ${address}`)
+      logger.info(
+        `User ${req.user.email} transferred ${amount} OGN to ${address}`
```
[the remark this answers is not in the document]
tomlinton: Done!

PR #3042, opened 2019-08-21
T3 UI

tomlinton:
This is all the T3 UI. There are a few outstanding things. I'll raise an issue with a catalogue of those when we get this up and running with some sample data on dev/staging.

Since T3 was originally created with `create-react-app` and I ejected it there is a bunch of boilerplate in here that might be confusing. We can safely ignore it!


franckc:
@tomlinton Let me know if ready for review and I can look. Though don't let my review block you - totally fine imho to merge in master and then we can iterate from there...

tomlinton:
Thanks @franckc it is all set 👍You are right. I'll merge it sometime after the eng meeting. Feel free to give it a quick look if you want though, but I wouldn't spend too much time on it.

PR #3053, opened 2019-08-22
T3 router fix

tomlinton:
- Changed T3 to hash router so the URLs in emails will work.
- Finished of send welcome token script to handle sending to all.
- Fixed a port that was wrong in CircleCI due to #2993.


infra/token-transfer-server/src/scripts/send_welcome_token.js
```diff
@@ -81,6 +72,23 @@ async function sendWelcomeEmail(email) {
...
+    const user = await User.findOne({ where: { email: config.email } })
+    if (!user) {
+      logger.error('User with that email does not exist')
+      process.exit()
+    }
+    sendWelcomeEmail(user)
+  } else {
+    logger.info('Sending welcome email to all users')
+    const users = await User.findAll()
+    users.map(sendWelcomeEmail)
```
franckc: Should this await otherwise it will fire hundreds of promises and sendgrid may rate limt ?

PR #3074, opened 2019-08-23
Sparrow dom/mobile wsod fix

sparrowDom:
### Description:
This solves the problem of "white screen" that appears sometimes in the dapp browser component when resuming application from background. As suggested by @tomlinton seems that lock authenticator screen is causing it.

**Solution:**
- do not show lock screen if user has left and re-entered the app within 2 minutes
- if longer than 2 minutes restart the application to still have security step and no white screen
- this way users should have enough time to do a quick action exiting our app (like tweeting, FB following...) and see the result of the action in the Dapp. While still having a security step if they leave the app for a longer time. 

### Checklist:


nick:
Closing in favor of #3072 - please reopen if there are other things this fixed 🙂

sparrowDom:
as per @joshfraser request app restarts if it was in background for more than 1 hour. @tomlinton can you take another look please. If it is ok with you I can try publishing this version on the stores once it gets merged?

PR #3084, opened 2019-08-23
Pared down web3-provider-engine

nick:
Stripped down web3-provider-engine to the bare bones.

The original web3-provider-engine had it's own block tracker, event emitter and a bunch of other stuff we don't need. This PR gives us a bare bones web3-provider-engine for our Metrics, ThrottleRPC and EthGasStation providers.

Will hopefully reduce blockNumber calls significantly...

### DIFF
--- a/packages/web3-provider/src/index.js
+++ b/packages/web3-provider/src/index.js
@@ -1,7 +1,7 @@
 /**
  * Web3 Provider Engine Subproviders
  */
-const ProviderEngine = require('web3-provider-engine')
+const ProviderEngine = require('./ProviderEngine')
 const {
   MetricsProvider,
   ThrottleRPCProvider,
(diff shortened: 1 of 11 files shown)


packages/web3-provider/src/index.js
```diff
@@ -43,39 +43,9 @@ function createEngine(web3Inst, options) {
...
-   * The blockTracker seems be a default function of web3-provider-engine, but
-   * it's not completely clear if it's necessary(emits the 'latest' event).  It
-   * has an internal currentBlock tracker.  Perhaps we should bolt onto it.
-   * Right now, we're specifically initializing the block tracker so we can
-   * disable the `skipConfig` parameter that it sends which breaks Alchemy
-   */
-  const engine = new ProviderEngine({ useSkipCache: false })
-
-  /**
-   * IF YOU REMOVE THIS, EVERYTHING WITH EXPLODE
```
mikeshultz: Except if you completely gut web3-provider-engine!

PR #3087, opened 2019-08-23
Add status bar to mobile app

nick:


tomlinton:
Looks fine to me, although I saw you raise a question in Discord about it. Not sure if we want to merge.

PR #3145, opened 2019-08-29
Infra changes for multi-marketplace support

mikeshultz:
### Description:

Fairly trivial update to listener.  Also adds `V001_Marketplace` to docker config.

Relayer doesn't require any changes.  It doesn't care about addresses, only the ABI.

### Checklist:


infra/discovery/src/listener/listener.js
```diff
@@ -114,9 +114,14 @@ async function main() {
   // List of contracts the listener watches events from.
   const contracts = {
     IdentityEvents: contractsContext.identityEvents,
-    Marketplace: contractsContext.marketplaces['000'].contract,
     ProxyFactory: contractsContext.ProxyFactory
   }
+  // Listen to all versions of marketplace
+  Object.keys(contractsContext.marketplaces).forEach(key => {
+    console.log('key', key)
```
nick: Can we take out the log?

mikeshultz: woops, yeah, I'll fix that before merge.  Thanks

infra/discovery/src/listener/listener.js
```diff
@@ -114,9 +114,13 @@ async function main() {
   // List of contracts the listener watches events from.
   const contracts = {
     IdentityEvents: contractsContext.identityEvents,
-    Marketplace: contractsContext.marketplaces['000'].contract,
     ProxyFactory: contractsContext.ProxyFactory
   }
+  // Listen to all versions of marketplace
+  Object.keys(contractsContext.marketplaces).forEach(key => {
+    contracts[`V${key}_Marketplace`] =
```
nick: One thing I just realized was `key` is 3 characters long (eg '000') but the config if 2 (eg 'V00'). Not sure if that affects this. Also might need to say `Object.keys(contractsContext.marketplaces || {})` incase no marketplaces are set yet for whatever reason

mikeshultz: it doesn't have any effect here, it's using the keys from the same object to reference it.  But maybe that should be standardized across the codebase.

PR #3155, opened 2019-08-30
event-cache PostgreSQL Backend Stuffs

mikeshultz:
### Description:

This is PR #2597 back from the dead.  Bunch of fixes to get the Postgres backend working properly. This includes functional conditional build-time import for `PostgreSQLBackend` that required a relatively minor webpack change.  Was hoping not to have to do that, but alas.  Will want to come back later and enable other services to use the backend(e.g. graphql) once things look good.  It can be enabled by setting `EVENTCACHE_ENABLE_PG=true` and any non-writers should have `EVENTCACHE_SLAVE=true`.

This also has a nice side effect of reducing the dapp bundle by almost 1MB by removing sequelize and its dependencies entirely.

Tested to work in docker-compose services, frontend build, and docker dapp build locally.  I'll want to try and deploy this to dev again on a slow night/day.

Example docker build: `docker build --build-arg ENVKEY="[insert your envkey]" -t origin-dapp -f devops/dockerfiles/origin-dapp .`

### Checklist:


docker-compose.yml
```diff
@@ -166,6 +167,7 @@ services:
       - IPFS_GATEWAY_URL=http://ipfs-proxy:9999
       - IPFS_API_URL=http://ipfs-proxy:9999
       - ENVKEY=
+      - ENABLE_EVENTCACHE_DB=true
```
franckc: If both the listener and graphql use event-cache, will that cause double writes to the event table ?

packages/event-cache/migrations/20190705152855-add-event-id.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.addColumn('event', 'id', {
+      type: Sequelize.STRING
```
franckc: Is index the primary key that replaces (blockNumber, logIndex) as primary key ?

BTW: since this table is empty in prod, as opposed to writing migrations to add columns,  it would be fine to start defining this entire table from scratch.

packages/event-cache/src/backends/PostgreSQLBackend.js
```diff
@@ -2,6 +2,40 @@ const { AbstractBackend } = require('./AbstractBackend')
...
+  if (ret.endsWith('Id')) {
+    ret = ret.replace(/Id$/, 'ID')
+  }
+  return ret
+}
+
+/**
+ * Convert a camelCaseString to an undorscored_string
+ * @param v The camelCase string string
+ * @returns {string} The undersocred string
```
franckc: nit: typo

packages/event-cache/migrations/20190703025936-add-event-prefix.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.addColumn('event', 'prefix', {
```
franckc: Is an index needed ?

docker-compose.yml
mikeshultz: This is an excellent point.  I don't think it ever showed problems in my docker testing, but it's a serious concern.  I think this is still safe to merge and deploy since only one prod system has it enabled and since `PostgreSQLBackend` uses upsert functionality, it shouldn't cause any conflict.  

But that's a waste of resources and there should be some work done to either disable writes on some of the EC consumers, or add some kind of auto-locking feature that prevents more than one instance of EC to write. Probably the former, since I think it'll be a conscious decision of ours to make listener do the writes.

Good catch.  This didn't even cross my mind.

packages/event-cache/migrations/20190705152855-add-event-id.js
mikeshultz: not an index.  This was added because it's data returned from web3 for an event.  Since we want to re-assemble the event objects the same as web3 returns them, we want to include everything.

packages/event-cache/migrations/20190703025936-add-event-prefix.js
mikeshultz: Yeah, this should probably be indexed.  Will add.

docker-compose.yml
franckc: Maybe to start with a simple solution, we could control this via an environment variable. Something like process.env.ENABLE_EVENT_CACHE_DB_WRITE which we would set to true only on either the graphql server or the listener. Then later if needed we can have a lock table or perhaps even move events writing to a completely separate service.

packages/event-cache/migrations/20190705152855-add-event-id.js
franckc: Got it. Sounds good. Maybe add a comment in the model to indicate the role of the id column.

packages/event-cache/migrations/20190703025936-add-event-prefix.js
franckc: sg

mikeshultz:
I've updated this a bit.  Now, if an instance is set as a slave, it won't make any writes(on any backend).  I didn't want to make it default slave because it'll affect everything.  So, if we bring up another instance of EC with postgres backend, it'll need to be set as a a slave.  The listener will be the designated master.

Env vars have also been renamed to use `EVENTCACHE_` as a prefix for uniformity.

Please review when you get a chance, I'd like to get this merged and setup on dev for testing.

packages/event-cache/src/backends/PostgreSQLBackend.js
```diff
@@ -2,6 +2,40 @@ const { AbstractBackend } = require('./AbstractBackend')
 
 const { debug } = require('../utils')
 
+/**
+ * Convert an undorscored_string to camelCase.
```
franckc: nit: underscored

packages/event-cache/src/backends/PostgreSQLBackend.js
```diff
@@ -2,6 +2,40 @@ const { AbstractBackend } = require('./AbstractBackend')
...
+ * @note This converts doubles (listingID) to underscored (listing_id)
```
franckc: nit: fix comment. the function actually does the opposite.

packages/event-cache/src/backends/PostgreSQLBackend.js
mikeshultz: how do I keep messing up that word...

packages/event-cache/src/backends/PostgreSQLBackend.js
mikeshultz: thanks

PR #3175, opened 2019-09-03
Tomlinton/language temporary fix

tomlinton:
Stop mobile app from controlling the language setting in the DApp for the meetup tonight. Will cherry pick to prod and revert after the meetup.


franckc:
On mobile so can’t approve thru GitHub but LGTM.

PR #3177, opened 2019-09-03
Sparrow dom/brave fix

sparrowDom:
### Description:
After some testing it turns out that default shield setting in Brave do not limit our DApp. But if the settings are more aggressive to block all cookies also localStorage doesn't work and our DApp doesn't load:
Screenshot 2019-09-03 17 35 42

There isn't much use in displaying the error modal if the Dapp breaks before that step. 

For now I am removing Brave modal warning. 

### Checklist:


sparrowDom:
I think a better solution would be to have a simple `js` script in `template.html` that would do a quick check if `localStorage` is enabled and display an error message (can be javascript's `alert`) if localStorage is disabled.

franckc:
CC @DanielVF in case he has any thoughts about what solution we should go for.

MarioBlanco:
> I think a better solution would be to have a simple `js` script in `template.html` that would do a quick check if `localStorage` is enabled and display an error message (can be javascript's `alert`) if localStorage is disabled.
> 
> @nick @MarioBlanco thoughts?

Yeah, always better to detect what we actually need for the app to work instead of focus on a given browser.

Issue #3140, opened 2019-08-29
Can't scroll modal on Android


norbertdragan:
**Update** 
Seems like the modal is also not scrollable in the Origin Marketplace app for Android. It means that users can't even complete the sign up flow. Essentially we are using all Android users until it's fixed.

**Original**
When I'm accessing the dApp via another application in WebView, the modal component is not scrollable. I tried it with both the MetaMask and imToken apps, and I got stuck at different screens where the modal component appeared (for example, [here](http://i.norbert.dev/c7e93c)).

I am using a OnePlus 6, Android 9. I also tested on iOS, and this issue was not present.

norbertdragan:
Quote from @shahthepro:

> @norbertdragan I've faced that issue. If we use position: fixed it works on iOS but not on Android and if we use position: absolute it works on Android but not on iOS.
> On iOS, the problem was that body was also scrollable even when there was another element on the front. It was not as bad as scroll completely not working on Android. So I think we decided to use position: absolute in some places
> The problem you are facing might be because of this.

PR #3192, opened 2019-09-04
Update Google Attestation page for WebView users

### LINKED ISSUE #2227: Google Attestation Not Supported In Mobile

norbertdragan:
### Description:

Fixes #2227.

- Detects if the user's browser is WebView for the Google Authetication
- Warns the user that it's unsupported
- Renders a button that sends the user to the app store/our mobile download page

Tested on Android and iOS using Cipher, the only app that allowed me to access my local network (and I tested many). 

[iOS Preview](http://i.norbert.dev/6807ec)

On Android, we have the scroll issue described in #3140.

### Checklist:

~- [ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)~


dapps/marketplace/src/pages/identity/OAuthAttestation.js
```diff
@@ -92,11 +92,27 @@ class OAuthAttestation extends Component {
...
+    const { isMobile, provider } = this.props
+
     const didOpen = !prevProps.open && this.props.open,
       didChangeStage = prevState.stage !== this.state.stage
     if (this.inputRef && (didOpen || didChangeStage)) {
       this.inputRef.focus()
     }
+
+    if (provider === 'google' && !prevProps.isMobile && isMobile) {
+      // Check wether the current browser is WebView in order to block Google OAuth and warn the user
```
shahthepro: When the profile page is being loaded in the mobile, It is always in a webview. User cannot access this page without a DApp browser. 

Instead of checking these, we could just check the `this.props.walletType`. Something like this: 
```
this.props.isMobile && (this.props.walletType === 'Mobile' || this.props.walletType === 'Origin Wallet')
```

norbertdragan: Good point, thanks!

norbertdragan: Addressed in b6d4a5b.

shahthepro: @norbertdragan Sorry if I didn't word it correctly. but I meant to remove the entire code from `componentDidUpdate`. It's complicating things. 

You don't have to hardcode the mobile app links either. `https://originprotocol.com/mobile` will open the corresponding store on mobile.

You could just use `isMobile` and `walletType` props inside the `render` method and render the App download modal.

dapps/marketplace/src/pages/identity/OAuthAttestation.js
```diff
@@ -169,35 +184,81 @@ class OAuthAttestation extends Component {
...
 
     return (
       <>
         <h2>{header}</h2>
-        <div className="help mt-0 mb-3">
-          <fbt desc="OAuthAttestation.neverPost">
-            We will never post on your behalf.
-          </fbt>
-        </div>
+        {!isForbidden && (
```
shahthepro: Adding to my previous comment, I feel it'd be better if we move Forbidden modal to a separate function or make it a separate component that can used in other places.

norbertdragan: Yes, and it's probably better to separate concerns anyway.

norbertdragan: I agree that a separate component makes a lot of sense. I'll go and separate concerns.

PR #3197, opened 2019-09-04
🐃 Deep checking of relay transactions

DanielVF:
Adds deep inspection of transactions we are requested to relay, to recursively ensure that only we only call allowed methods on allowed contracts, even inside inner-transactions.

This checks things like a transaction calling `swapAndMakeOffer` that then nests sub-transactions to the marketplace, to a unisawp exchange, and to a token contract.

In this PR, the validator code runs in logging mode only for now. This way we can see if I have the limits too tight, or am missing anything, without failing any real transactions.


infra/relayer/src/relayer.js
```diff
@@ -334,6 +337,17 @@ class Relayer {
...
 
+    // 2.1 Deep check that the potentialy nested transaction are calling
+    // approved methods. For now, we don't fail on this, since we might
+    // have the whitelist too tight. We'll check the logs after a while
+    // and see if there is anything we should add.
+    try {
+      this.validator.validate(PROXY_HARDCODE, txData)
+    } catch (e) {
+      logger.error('Error in transaction validator', e)
+      throw e
```
mikeshultz: maybe do a `res.status(500).send [...]` instead of killing the relayer

infra/relayer/src/validator.js
```diff
@@ -0,0 +1,240 @@
...
+    } else if (method.name == 'swapAndMakeOffer') {
+      if (!this.isValidTokenAddress(params._token)) {
+        logger.info(
+          `Validation failed. ${params._token} is not an allowed ERC20 address`
+        )
+        return false
+      }
+      validationsNeeded.push([params._marketplace, params._offer])
+      validationsNeeded.push([params._exchange, params._swap])
+    } else {
```
mikeshultz: Do we want to limit this to only these 4 methods?  Just thinking that we were talking about adding `transferToOwner()` for sweeps yesterday.

infra/relayer/src/relayer.js
```diff
@@ -3,6 +3,7 @@
 const memoize = require('lodash/memoize')
 const Web3 = require('web3')
 const utils = require('ethereumjs-util')
+const { Validator, PROXY_HARDCODE } = require('../src/validator')
```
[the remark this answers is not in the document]
DanielVF: Thanks!

infra/relayer/src/validator.js
DanielVF: I've added `transferToOwner` to the list. If there's anything else we need later, we can add it then.

infra/relayer/src/relayer.js
DanielVF: Thank goodness for code reviews!

I've pulled out the throw - that shouldn't have been there. It will now correctly log verifier crashes, but not crash the relayer.

mikeshultz:
Well, I did approve the PR! Yes, g2g. Ship it and let's see what it tells
us in the logs.

franckc:
Actually the Relayer unit tests failed. Let's fix that before we merge.
@DanielVF do you mind taking a look ?

PR #3217, opened 2019-09-05
Add Unsupported Attestation for WebView

### LINKED ISSUE #2227: Google Attestation Not Supported In Mobile

norbertdragan:
### Description:

Fixes #2227.

- Detects if the user's browser is WebView for the Google Authetication
- Warns the user that it's unsupported
- Renders a button that sends the user to the app store/our mobile download page
- As discussed in #3192 yesterday, Unsupported Attestation is a now it's separate modal

Test build is [here](https://originprotocol.github.io/test-builds/unsupported-attestation/). Use MetaMask, imToken, Cipher or similar WebView app to test. I used Cipher to test.

[iOS Preview](http://i.norbert.dev/6807ec)

### Checklist:

~- [ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)~


norbertdragan:
@shahthepro I ended up checking for the `userAgent` to determine WebView, because `isMobile` only checks for the browser width, which can be narrow on the desktop as well.

tomlinton:
Hey @norbertdragan the `withIsMobile` HOC also has ` isMobileApp` prop which you should check here. We fake the UserAgent in our own mobile app because certain (cough) attestation providers won't let us do attestations with a WebView UserAgent.

For other non-Origin wallets I think width is probably the only way we have. Although I'm sure each wallet probably implements something on the `window` object or similar that they user internally. If we could figure out each of those and add them to the `withIsMobile` HOC we could stop relying on width, which is a bit problematic as you note.

tomlinton:
Closing due to inactivity.

Issue #3199, opened 2019-09-04
Improve Sentry integration in the DApp


tomlinton:
The DApp swallows a lot of errors/outputs to console and we have no visibility of these (especially on mobile). There are various places we should add calls to Sentry, e.g. [here](https://github.com/OriginProtocol/origin/blob/d0714f7012a0aee2394324aa32deb50fa5c8873e/dapps/marketplace/src/components/WaitForTransaction.js#L130).

We probably need this anywhere there is a `console.error` in the DApp. We might be able to just use [this](https://docs.sentry.io/platforms/javascript/?platform=browsernpm#captureconsole) or patch console to give us an integration everywhere.

mikeshultz:
PR #3226 for quick & dirty integration for our current issues with `WaitForTransaction`.

Leaving this open because later, I want to take it further and add environments so we can get errors reported from staging and maybe dev.  Add it to the catch-all error component where all unhandled errors bubble up to.  And integrate it across the dapp, perhaps somehow integrated into the apollo-react hook function `useQuery()`.

Perhaps this isn't a P0 any longer after that PR?

micahalcorn:
Taking this one down a couple of notches 👇

mikeshultz:
Grouping should be worked on a bit.

https://docs.sentry.io/data-management/event-grouping/

The following sentry event group govers a lot of unrelated ground with graphql errors:

https://sentry.io/organizations/origin-protocol/issues/1098412085/?referrer=github_integration

It's not very useful to us to have all these grouped together.

mikeshultz:
PR #3535 added captures for QueryError and adjusted grahpql error grouping a bit. If this turns out good, I think that might wrap up this issue unless there are objections.

PR #3226, opened 2019-09-05
Sentry integration into WaitForTransaction

mikeshultz:
### Description:

Quick & dirty integration of Sentry into `WaitFortransaction` component for #3199

### Checklist:

### DIFF
--- a/dapps/marketplace/src/components/WaitForTransaction.js
+++ b/dapps/marketplace/src/components/WaitForTransaction.js
@@ -8,6 +8,7 @@ import MetaMaskAnimation from 'components/MetaMaskAnimation'
 import query from 'queries/TransactionReceipt'
 import withWallet from 'hoc/withWallet'
 import withConfig from 'hoc/withConfig'
+import Sentry from 'utils/sentry'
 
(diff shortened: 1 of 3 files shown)


dapps/marketplace/src/components/WaitForTransaction.js
```diff
@@ -128,6 +129,7 @@ class WaitForTransaction extends Component {
           let content
           if (error) {
             console.error(error)
+            Sentry.captureException(error)
```
tomlinton: Does the <Query /> component from `react-apollo` return errors as something that `Sentry.captureException` won't choke on?

tomlinton: It looks [like it extends](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/errors/ApolloError.ts) `Error` so it should be OK. Not sure if you can verify this easily.

mikeshultz: Yeah, was worried about that too but it returns an `ApolloEror` which extends `Error`.  Should be fine

PR #3238, opened 2019-09-06
Fix Android mobile modal scroll

### LINKED ISSUE #3140: Can't scroll modal on Android

tomlinton:
Fixes #3140. Maybe the `touch-action: none` was there for a reason. Might drop this on dev to test it on iOS?


dapps/marketplace/src/components/MobileModal.js
```diff
@@ -231,7 +231,6 @@ require('react-styl')(`
       opacity: 1
       display: block
   .mobile-modal-light
-    touch-action: none
```
nick: @shahthepro can you remember why this was added? OK to merge?

shahthepro: @nick Not that I can remember. But last time this issue popped up, @tomlinton had [opened this PR](https://github.com/OriginProtocol/origin/pull/2539) and it did work on Android as well as on iOS (with some scroll issues). So I think some other change would have caused this again and not this line. I'm not sure though.

Issue #3237, opened 2019-09-06
VM Execution Error


micahalcorn:
On https://shoporigin.com/#/purchases/1-000-3118-8492700-1...

```
Error: GraphQL error: VM execution error.
    at new t (bundle.esm.js:63)
    at t.getCurrentResult (bundle.esm.js:159)
    at t.getQueryResult (react-hooks.esm.js:245)
    at t.a.getExecuteResult (react-hooks.esm.js:71)
    at t.execute (react-hooks.esm.js:88)
    at c (react-hooks.esm.js:344)
    at ko (react-hooks.esm.js:318)
    at xo (react-hooks.esm.js:353)
    at Xo (react-components.esm.js:8)
    at Qi (react-dom.production.min.js:150)
    at ka (react-dom.production.min.js:166)
    at Oa (react-dom.production.min.js:179)
    at zs (react-dom.production.min.js:232)
    at Hs (react-dom.production.min.js:233)
    at Co (react-dom.production.min.js:249)
    at xo (react-dom.production.min.js:248)
    at To (react-dom.production.min.js:245)
    at Ys (react-dom.production.min.js:243)
    at sa (react-dom.production.min.js:157)
    at Object.error (react-hooks.esm.js:221)
    at b (Observable.js:134)
    at v (Observable.js:165)
    at e.value (Observable.js:224)
    at bundle.esm.js:434
    at Array.forEach (<anonymous>)
    at re (bundle.esm.js:434)
    at Object.r [as error] (bundle.esm.js:385)
    at i (bundle.esm.js:1208)
    at bundle.esm.js:1246
    at bundle.esm.js:1558
    at Set.forEach (<anonymous>)
    at bundle.esm.js:1556
    at Map.forEach (<anonymous>)
    at e.broadcastQueries (bundle.esm.js:1554)
    at bundle.esm.js:1645
    at Object.next (Observable.js:308)
    at b (Observable.js:130)
    at v (Observable.js:165)
    at e.value (Observable.js:219)
    at bundle.esm.js:865
    at Set.forEach (<anonymous>)
    at Object.next (bundle.esm.js:865)
    at b (Observable.js:130)
    at v (Observable.js:165)
    at e.value (Observable.js:219)
    at bundle.esm.js:23

@ breadcrumbs.js:58 

  query Offer($offerId: ID!) {
    marketplace {
      offer(id: $offerId) {
        listing {
          ...basicListingFields
        }
        ...basicOfferFields
      }
    }
  }
[...]

PR #3250, opened 2019-09-06
Sparrow dom/historical listing fix

sparrowDom:
### Description:

 - When user is observing a historical listing replace the purchase option with a "View current listing" button. 
- Also hide total payment amount and `Your Eth will be transferred to escrow` info box.

**Related Issue:** https://github.com/OriginProtocol/origin/issues/3237
Screenshot 2019-09-06 19 47 31

### Checklist:


micahalcorn:
Thanks @sparrowDom 👏 Can we get more whitespace below that yellow alert? Probably about the same as below the blue button.

Also, I'm wondering since you used the word "available", do we necessarily know here that the listing has availability? Or could a buyer click the blue button and then see "sold out" on the next screen. Or will that entire section change state on this initial screen and not display the alert/button at all?

micahalcorn:
@sparrowDom let's also hide the "Your ETH will..." in this case. And for that matter, I would probably recommend hiding all of the quantity+total+payment as we do when the buyer has not yet onboarded. 🤔

sparrowDom:
As requested warning now looks like this @micahalcorn :
Screenshot 2019-09-06 23 38 18

Do tell if any additional tweaks are required. 

@nick PTAL when you can spare some time

dapps/marketplace/src/utils/listing.js
```diff
@@ -0,0 +1,19 @@
+const regex = /(\d*)-(\d*)-(\d*)-(\d*)/
```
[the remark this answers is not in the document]
shahthepro: Also, you may want to specify `^` and `$` characters to be more defensive.

dapps/marketplace/src/pages/listing/_PaymentOptions.js
```diff
@@ -89,7 +91,7 @@ const PaymentOptions = ({
     needsSwap = false,
     noEthOrDai = false
 
-  if (hasIdentity) {
+  if (hasIdentity && !isHistoricalListing(listing)) {
```
shahthepro: Why not move this check out of this component? We could avoid rendering `PaymentOptions` if it is a historical component right?

micahalcorn:
I haven't tested this myself, but I'm under the impression that we still need to make some changes here.

The basic issue (I'm told) is that, if a buyer visits a listing detail with a block number in the URL's listing id param, then after making an offer from that block-specific listing, the user will be routed to a transaction detail page with ***its*** URL offer id param being polluted with an unnecessary block number, which gets interpreted incorrectly and throws the exception described in #3237. This issue applies ***whether or not*** the listing has been updated.

So here is what I was proposing for when the listing detail page is block-number-specific:

1) If the listing ***has*** been updated and I ***am not*** looking at the newest version, then display the new UI as this PR does.
2) If the listing ***has not*** been updated or I ***am*** looking at the newest version, then the transaction detail navigation should be smart enough not to route me using an invalid offer id (polluted by a block number).

dapps/marketplace/src/pages/listing/mutations/Buy.js
```diff
@@ -85,6 +90,12 @@ class Buy extends Component {
 
       if (this.state.error) {
         content = this.renderTransactionError()
+      } else if (isHistoricalListing(this.props.listing)) {
+        action = (
+          <HistoricalListingWarning
+            listingId={currentListingIdFromHistoricalId(this.props.listing)}
```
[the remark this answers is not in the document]
sparrowDom: Yeah agreed. Moving the original listing resolution to historicalListingWarning component

dapps/marketplace/src/utils/listing.js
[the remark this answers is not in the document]
sparrowDom: All good points thanks

dapps/marketplace/src/utils/listing.js
```diff
@@ -0,0 +1,19 @@
...
+
+/* Matches when we are not looking at current listing state, rather a snapshot in the past.
+ * This can happen when for example user clicks on a listing from a Purchase detail view
+ */
+export function isHistoricalListing(listing) {
+  return !!listing.id.match(regex)
```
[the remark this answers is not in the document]
sparrowDom: nice, changed to that

dapps/marketplace/src/pages/listing/_PaymentOptions.js
sparrowDom: `PaymentOptions` has children that it needs to render. If we disable the rendering of payment options then components like `_BuySingleUnit` and `Buy` do not get rendered either: https://github.com/OriginProtocol/origin/blob/5aa1d52bf362f8221ab6258f88f3899c773c461f/dapps/marketplace/src/pages/listing/_BuySingleUnit.js#L62-L75

sparrowDom:
thanks @shahthepro for a great review. Merging in once tests pass

dapps/marketplace/src/pages/listing/_PaymentOptions.js
shahthepro: Yeah, But we don't want users to make offer on historical listing, if I'm not mistaken.

From what I understand, we would have a "View Current Listing" instead of "Purchase" button right?

sparrowDom:
@micahalcorn just saw your comment if you are ok with it going with the following solution: 

If a user is looking at the newest version of the listing in listingDetail the component just redirects to newest version before rendering the "historical view" of the listing version.

/cc @shahthepro the commit doing this: https://github.com/OriginProtocol/origin/pull/3250/commits/b4cc88620f81d7d2021be400361c29195899ccb6

sparrowDom:
Merging this in...

@micahalcorn & @shahthepro if any additional changes requested will be resolved in subsequent PRs

micahalcorn:
I don’t quite understand, @sparrowdom. 😕 The block numbered version only renders if it is NOT the current version? And if it is, it redirects to a URL with no block number?

dapps/marketplace/src/pages/listing/_PaymentOptions.js
sparrowDom: The button that displays the "View Current Listing" is a Buy button component. This ensures us that a user is never able to make an offer on a historical listing since the buy button will always say "View current listing" instead of "Purchase". 

And that buy button is part of the children that get passed to the PaymentOption Component.

sparrowDom: The button that displays the "View Current Listing" is a Buy button component. This ensures us that a user is never able to make an offer on a historical listing since the buy button will always say "View current listing" instead of "Purchase".

And that buy button is part of the children that get passed to the PaymentOption Component.

sparrowDom:
Yes @micahalcorn it works exactly like you describe: 
- listing with a block number in the url that **is also the current listing version** will immediately redirect to the current listing view that has no block number in the url. Rendering the purchase button (the whole redirect happens fast is not something I noticed)
- listing with a block number in the url that **is not the current listing version** will render the `historical listing` having a "View current listing" button instead of a purchase button and a yellow warning. 

Do you think we should take a different approach, or that this one could be confusing?

micahalcorn:
@sparrowDom thanks, that certainly works from the user's perspective, and it sounds like the simplest from the engineering perspective. I would think that it might not be ideologically appropriate, but I don't have strong opinions. 🤷‍♂

sparrowDom:
Yeah it was the simplest solution. Happy to change it to something else if need/wish be.

PR #3254, opened 2019-09-06
Gas price changes and subprovider

mikeshultz:
### Description:

This is quick and dirty, duplicates code, and should be cleaned up later.  Trying to port ZeroClientProvider to our web3-provider will take more effort.  

This adds `EthGasStationProvider` to `mobile-bridge` and adjusts gas price to average across the board.

Here's the Rinkeby test build: https://originprotocol.github.io/test-builds/gas-provider

### Checklist:


nick:
Seems OK... haven't tested it though

mikeshultz:
### DO NOT MERGE YET

mikeshultz:
@nick @franckc, do you think this should be enabled for desktop dapp?  I briefly tested it earlier and it worked okay with MM, but I'm a little worried about unintended consequences of adding `@origin/web3-provider` in such a short time without testing.

mikeshultz:
ok, well, on the urging from @micahalcorn, I've enabled `@origin/web3-provider` across the board, including `EthGasStationProvider`.  Test please.

packages/web3-provider/src/subproviders/EthGasStationProvider.js
```diff
@@ -23,7 +23,7 @@ class EthGasStationProvider extends SubProvider {
     const jason = await res.json()
     if (typeof jason[GAS_PRICE_KEY] !== 'undefined') {
       // values come from EGS as tenths of gwei
-      return String(jason[GAS_PRICE_KEY] * 1e8)
+      return '0x' + (jason[GAS_PRICE_KEY] * 1e8).toString(16)
```
franckc: That won't break the relayer, will it ?

mikeshultz: tests passed

mikeshultz: I feel like I should elaborate that standad JSON-RPC interfaces return in hex.  This shouldn't break anything, tests passed, but it is a valid concern and we should test it when it gets deployed.

PR #3263, opened 2019-09-07
gasPrice suggestions for metamask and others

mikeshultz:
### Description:

This adds the `gasPrice` parameter to non-relayer transactions in `_txHelper`.  I'm not entirely sold on the non-metamask portion of this since it always adds an extra call.  I'd like other peoples' opinion on this.

CC: @micahalcorn 

### Checklist:


packages/graphql/src/mutations/_txHelper.js
```diff
@@ -46,7 +50,11 @@ function useRelayer({ mutation, value }) {
   if (!contracts.config.relayer) reason = 'relayer not configured'
   if (!mutation) reason = 'no mutation specified'
 
-  if (mutation === 'makeOffer' && value && value !== '0') {
+  if (
+    ['makeOffer', 'swapAndMakeOffer'].includes(mutation) &&
```
franckc: So prior to this we were using the relayer for swapAndMakeOffer ? That's odd, I don't remember my offers with DAI payment going thru the relayer...

packages/graphql/src/mutations/_txHelper.js
```diff
@@ -141,6 +167,22 @@ export default function txHelper({
...
+      if (window.ethereum) {
+        // Don't use eth_gasPrice for MM
+        try {
+          gasPrice = await fetchGasPrice()
+        } catch (err) {
+          debug(`Error fetching gas price`, err)
+        }
+      } else {
+        // But use it when we're using our provider
+        gasPrice = await web3.eth.getGasPrice()
```
franckc: Is there a point in us having an EthGasStationProvider ?
Could we get rid of it and just always set the gasPrice here ?

packages/graphql/src/mutations/_txHelper.js
```diff
@@ -141,6 +167,22 @@ export default function txHelper({
...
+    debug(`shouldUseRelayer: ${shouldUseRelayer}`)
+
+    // Get our gas price
+    if (!shouldUseRelayer && ['mainnet', 'rinkeby'].includes(contracts.net)) {
+      if (window.ethereum) {
+        // Don't use eth_gasPrice for MM
+        try {
+          gasPrice = await fetchGasPrice()
+        } catch (err) {
+          debug(`Error fetching gas price`, err)
```
franckc: Should we log to Sentry ?

packages/graphql/src/mutations/_txHelper.js
franckc: We could do this as a separate cleanup PR. Just asking for my own understanding... :)

packages/graphql/src/mutations/_txHelper.js
mikeshultz: No, it just so happened that it was deflected by `useProxy`.  `useProxy` didn't work for my logic so i added it here too.

packages/graphql/src/mutations/_txHelper.js
mikeshultz: `web3.eth.getGasPrice` will use the provider(if it's there).  And so will all the services that don't use GraphQL.

packages/graphql/src/mutations/_txHelper.js
mikeshultz: I don't think there's been any Sentry integration in graphql(yet).

franckc: Ah yes, very good point. You are still very lucid at this late hour you know... :)

Issue #3288, opened 2019-09-08
Missing Shipping Address Option


micahalcorn:
***Steps to reproduce:*** 😬 

- [ ] Kill app
- [ ] Open app
- [ ] Open hamburger menu
- [ ] Click "Add a listing"
- [ ] Click "For Sale"
- [ ] Pick a category, any category
- [ ] Add title and description
- [ ] See quantity field
- [ ] Wonder where the shipping address requirement option is
- [ ] Pull to refresh
- [ ] See shipping address requirement option

Related: #3287

-----

At Origin, we use [ZenHub](https://www.zenhub.com/) to manage our engineering tasks and product development. [Download their browser extension](https://www.zenhub.com/extension) and check out our open workspace at [github.com/originprotocol/origin#zenhub](https://github.com/originprotocol/origin#zenhub).

franckc:
Per discussion on engineering channel, we'd like to first focus on resolving this issue on the Origin mobile app. Then we can look at other platforms.

Issue #3287, opened 2019-09-08
Editing Shippable Listings


micahalcorn:
I'm fairly certain that, when editing a listing that requires a shipping address from the buyer, the "Listing Details" step of editing does not default to "Yes" on the radio button. I haven't checked to see whether this would lead to accidentally disabling the shipping address requirement, but I suspect that it would. Let's ensure that it defaults to the saved value. 🔘

-----

At Origin, we use [ZenHub](https://www.zenhub.com/) to manage our engineering tasks and product development. [Download their browser extension](https://www.zenhub.com/extension) and check out our open workspace at [github.com/originprotocol/origin#zenhub](https://github.com/originprotocol/origin#zenhub).

micahalcorn:
I ***_think_*** that we also want to default to "Yes" on all For Sale listing creation as well. 🤔

PR #3293, opened 2019-09-08
Shipping Address refinement

shahthepro:
- [x] Remove "Enable Encryption" modal (#3288)
- [x] Redirect to onboarding, if messaging keys not found, while trying to create a listing (Closes (#3288)
- [x] "Require Shipping" defaults to yes for "For Sale" category (#3287)
- [x] Retain "Require Shipping" value when editing (#3287)
- [x] Use separate keys to cache existing and new listings (#3295)

### Checklist:


dapps/marketplace/src/hoc/withMessagingStatus.js
```diff
@@ -5,9 +5,9 @@ import query from 'queries/WalletStatus'
 
 import get from 'lodash/get'
 
-function withMessagingStatus(WrappedComponent) {
+function withMessagingStatus(WrappedComponent, { excludeData } = {}) {
```
sparrowDom: Maybe only 1 comment. Is there any substantial effect to `excludeData`? I mean it is still being queried, the only difference is that is not being passed down to the wrapped component. Is the idea that this conserves memory?

shahthepro: Added that as `messagingStatus` prop causes some components to rerender after a few milliseconds. This sometimes results in failing marketplace tests. But it won't be necessary once Nick's #3255 PR gets merged.

sparrowDom: Ahh haven't thought of that! Makes sense, thanks for explaining it.

PR #3297, opened 2019-09-08
Mobile refactor

### LINKED ISSUE #2518: Migrate Origin Marketplace to AndroidX
From 0.60 react-native is moving to AndroidX which is a new Android [support library ](https://developer.android.com/jetpack/androidx). I tested out a few builds and it looks like only one or two packages are causing issues.

### LINKED ISSUE #2625: Identity published in mobile onboarding does not trigger a refetch of identity in the DApp
The DApp looks like it triggers a refetch when it [deploys an identity](https://github.com/OriginProtocol/origin/blob/57f683a6489fb144410b850b64b04840c6cfe041/dapps/marketplace/src/pages/user/Profile.js#L395). The mobile app can't do this so when mobile deploys an identity it sometimes requires a hard refresh of the DApp before the identity displays. I'm not sure of a good solution here but maybe exposing a refetch method on window.appComponent (like for [onLocale](https://github.com/OriginProtocol/origin/blob/master/dapps/marketplace/src/index.js#L66)) might work?

Note I seem to recall that in some cases it displayed an identity without a refresh.

@nick I know you thought #2611 might fix this but it doesn't appear to have.

### LINKED ISSUE #2731: Deferred Native Onboarding
To be consistent with [our philosophy from our Q4 desktop user activation](https://docs.google.com/document/d/1Sma1yr_Swa6NtKYnaiEDqzOiWvJSAz0rClmZVDgAu0k/edit?usp=sharing), we should only introduce the native onboarding steps when a user expresses intent to buy, sell, or contact another user. A new installation of our mobile app should open directly to the marketplace webview. When a user initiates an action that requires an Ethereum account, a modal should open from the bottom guiding the user through wallet establishment, profile creation, etc.

### LINKED ISSUE #2884: Mobile app messaging conversation page scroll and refresh
The message conversation window has a scroll area outside of the viewport.  When trying to scroll it down, it will more-often-than-not refresh.  This doesn't seem to happen every time, however.

messaging-scroll-bug

### LINKED ISSUE #3272: Mobile app balance error
Seen when running the app:

    YellowBox.js:67 Could not retrieve balances using GraphQL:  Error: TypeError: Cannot read property 'toString' of null Given value: "null"
        at Object.toBN (utils.js:64)
        at tokenBalanceFromGql (currencies.js:30)
        at _callee7$ (marketplace.js:793)
        at tryCatch (runtime.js:45)
        at Generator.invoke [as _invoke] (runtime.js:271)
        at Generator.prototype.<computed> [as next] (runtime.js:97)
        at tryCatch (runtime.js:45)
        at invoke (runtime.js:135)
        at runtime.js:145
        at es.promise.js:105

It repeats often and doesn't seem to clear up.  Balances do not update on the account page of the app, but they are accurate on the identity dropdown.  Chance it's related to #3266

### LINKED ISSUE #3341: Mobile App Refactor
This tracks the work on refactoring the mobile app to support Samsung integration and removal of native onboarding. 🚀

-----

tomlinton:
- Updated all packages, including RN 0.60 with autolinking instead of using `react-native-link`
- Migrated to AndroidX
- Removed native onboarding
- Disabled pull to refresh on Android until we find a better solution
- Heavy refactoring to make code simpler (especially the WebViews)
- Replaces all of web3js with ethers (which is awesome) and removes all the scaffolding necessary to make web3js work in react-native
- Renamed from `OriginCatcher` to `OriginMarketplace`
- Samsung Blockchain Keystore integration

Closes #3025 
Closes #2518 
Closes #2731 
Closes #2884 
Closes #2625
Closes #3272 
Closes #3341


nick:
Whoa new candidate for heftiest PR of all time 😅

tomlinton:
Not an award I wanted to be winning 😬

mobile/src/screens/welcome.js
```diff
@@ -0,0 +1,180 @@
...
+    return (
+      <SafeAreaView style={styles.content}>
+        <View style={{ ...styles.container, flexGrow: 2 }}>
+          <Image
+            resizeMethod={'scale'}
+            resizeMode={'contain'}
+            source={require(IMAGES_PATH + 'origin-dark-logo.png')}
+            style={styles.image}
+          />
+          <Text style={styles.title}>
```
micahalcorn: I think this one's supposed to be `#0b1823`. 😐

tomlinton: Is that the colour of the title?

micahalcorn: Yeah, but apparently not the titles at the top of other screens 🤷‍♂️ I like the darker one anywhere.

tomlinton:
Going to merge this as it is causing some confusion and has been hanging around way too long!

PR #3331, opened 2019-09-10
Default to using ETH rather than DAI

franckc:
### Description:

Over the last few days we found a number of contract related issues when doing DAI payments. It is certainly a more complex flow that gets executed when a DAI swap is involved.
For now there is no real reason to push buyers to use DAI if the seller accepts ETH.

This change updates the DApp code to default to ETH.

Side note: this is a short-term fix with the goal to achieve the desired behavior with minimum code change. I'd love for us to spend time cleaning up and commenting that code further. For someone unfamiliar with this code like I was, I found it hard to understand the logic that takes place.

### Checklist:

- [X] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)


micahalcorn:
***Hallelujah!***

hallelujah

franckc:
@shahthepro @micahalcorn I'm going to merge this and test on staging. It's a small PR so can be reverted if it ends up causing issues...

PR #3386, opened 2019-09-13
Adds Sentry context

mikeshultz:
### Description:

Adds some context to Sentry error reports including:

- User address when available (set in Profile nav)
- Whether or not its a WebView

Any other suggestions for useful context, @micahalcorn @franckc @shahthepro @sparrowDom 

### Checklist:


franckc:
A couple suggestions for other useful tags:
 - If mobile, identifier of the waller: Origin, Coinbase, ImToken, MetaMask, etc... (does the web3 provider gives some sort of identifier ?)
 - If Origin mobile, version of the app

Not blocking for merging this PR though. These could get added in a subsequent PR.

mikeshultz:
> If mobile, identifier of the waller: Origin, Coinbase, ImToken, MetaMask, etc... (does the web3 provider gives some sort of identifier ?)

User-Agent is alraedy included.  I'm not sure if there's any way to go further than that.  We could include the web3 version, but I thought we replaced any global web3 with our own using their provider?

> If Origin mobile, version of the app

Is there a way to glean this from the dapp?  @tomlinton, is there anything already, or do you think it would be worth injecting this?

tomlinton:
The web3 provider has an [isOrigin](https://github.com/OriginProtocol/origin/blob/32002684d708e662d02e03d6a1ec1fdc45bd22d4/packages/mobile-bridge/src/index.js#L113) attribute. I also use `window.ReactNativeWebView` to detect [our own app](https://github.com/OriginProtocol/origin/blob/32002684d708e662d02e03d6a1ec1fdc45bd22d4/dapps/marketplace/src/hoc/withIsMobile.js#L34) which seems to work pretty well.

mikeshultz:
Already have `window.ReactNativeWebView` to detect WebViews, but wondering if that might not be unique to our mobile app.  Will try to detect `isOrigin`.  Thanks for the info.

mikeshultz:
I'm not convinced the `isOrigin` detection is gonna work.  I doubt the web3 from the app is exposed to the dapp through the WebView...

sparrowDom:
looks good @mikeshultz. Maybe it would make sense in the future that we expose the build version of Origin Marketplace in the app and add that to sentry context.

dapps/marketplace/src/pages/nav/Profile.js
```diff
@@ -42,10 +43,16 @@ const ProfileNav = ({
...
           return null
         }
-        if (!get(data, 'web3.primaryAccount')) {
+        const accountID = get(data, 'web3.primaryAccount.id')
+        if (!accountID) {
           return null
         }
 
+        // set the user for sentry
+        Sentry.configureScope(scope => {
```
shahthepro: I'd suggest to move this out of this HOC. We could add a component (added to root inside App) that uses this HOC and set the scope inside `useEffect` hook of that component.

shahthepro:
@mikeshultz I just pushed a change that moves the logic to App component. This is because we don't render the `nav/Profile` component in every page (in /user and /profile pages to be exact). 

Hope you don't mind me hijacking your branch. :P

mikeshultz:
@shahthepro, I don't mind the changes, thanks.  No concerns over the extra hoc+query?

shahthepro:
@mikeshultz 
>  No concerns over the extra hoc+query?
It shouldn't be a problem since there is no other component that is guaranteed to render on every page under every circumstance.

mikeshultz:
I was thinking performance, not effectiveness.  But if it's no concern to you all, it's none of mine.

Unless there's any additions or corrections, I'll move to get this merged today.  Mobile version would be cool, but will save that for another PR since it'll probably be a bit more complex.

Issue #3229, opened 2019-09-05
Transaction Signing Delay


sparrowDom:
Android Coinbase Wallet, Mainnet.

When onboarding trying to publish a profile the DApp prompts me to confirm transaction in my Coinbase Wallet but the actual Coinbase prompt for signing shows up 30 seconds late. As a normal user I would probably assume something is wrong and quit (since the dapp communicates I should see the prompt immediately) 

Do not know if this has anything to do with relayer or meta-transactions...

![Screenshot_20190906-003353](https://user-images.githubusercontent.com/579910/64388477-85557c80-d03f-11e9-9f78-d9058fdfdb17.jpg)

micahalcorn:
@sparrowDom I experienced the same in basically all of the mobile DApp browsers. ☹️ I had similar experiences when conducting actual transactions, so I suspect that this is unrelated to identity.

micahalcorn:
<img width="868" alt="Screen Shot 2019-09-07 at 6 41 18 AM" src="https://user-images.githubusercontent.com/273937/64475678-8919f980-d13a-11e9-9d2b-7a7853993b6a.png">

micahalcorn:
I actually just saw this in our app and I got tired of waiting for the confirmation prompt to appear. 😫 I swiped backward to the previous screen, tried again, and couldn't reproduce the seemingly-endless spinner.

tomlinton:
There is a preflight request to the relayer to see if it is going to accept the transaction, so there is at least a few round trips that have to happen before the request gets to the signing phase. Maybe we could remove that preflight request.

sparrowDom:
If a user is using a relayer and we are supplying the gas for the transaction execution would it make sense that we just communicate that part?

Some kind of a modal saying: `Negotiating with Origin to sponsor the gas for this transaction.` After that we could briefly display we have approved the gas for the transaction. (though that might be overwhelming to the user)

With the idea to communicate what is going on, why the user is waiting. If we are sponsoring the gas, he might not have problems waiting a bit more. In the end maybe a user could also feel rewarded since Origin has given him the gas.

micahalcorn:
Yeah @sparrowDom I would prefer to prevent the delay, but if there's going to be one, we should be more accurate about what to expect. 😕

mikeshultz:
[...]

PR #3394, opened 2019-09-14
Purse+relayer timeout on account acquisition

mikeshultz:
### Description:

Adds a timeout on account acquisition to beat the nginx HTTP timeout and potential incongruity of a transaction being sent without the client knowing.

The timeout is currently set at 15s.  I'm open to arguments on changing that.

I would recommend this is accompanied by more handling in the dapp, but I'm not sure what that should look like.  Right now, I think it would just move right along and prompt the user to pay for gas.  @micahalcorn may have opinions on that.

Ref: #3229 

### Checklist:


mikeshultz:
CC @sparrowDom

infra/relayer/src/purse.js
```diff
@@ -59,6 +59,7 @@ const REDIS_PENDING_PREFIX = `pending_tx_`
 const REDIS_PENDING_TX_PREFIX = `pending_txobj_`
 const JSONRPC_QPS = 100
 const JSONRPC_MAX_CONCURRENT = 25
+const ACCOUNT_ACQUISITION_TIMEOUT = 15000
```
franckc: Super nit for when someone looks is on-call and gets woken up because nothing is working: add a comment mentioning the unit is msec or change naming to ACCOUNT_ACQUISITION_TIMEOUT_MSEC

infra/relayer/src/purse.js
```diff
@@ -327,10 +345,16 @@ class Purse {
...
       } while (await tick())
     } catch (err) {
+      // We want to throw on timeouts so it can be handled by the sender
+      if (err.message.indexOf('timeout') > -1) throw err
+
       logger.error('Unhandled error in getAvailableAccount()')
       handleError(err)
     }
 
+    // Clear the acquistiion timeout
```
franckc: typo. acquisition

infra/relayer/src/purse.js
mikeshultz: Considering all time-based units in JS are in millisecond, is this really something we need to comment every time?  I wonder if it shouldn't be the other way around, comment if it's not millisecons.

franckc: I personally favor being explicit for all units.
My preference comes from years ago working in the space industry... See https://www.simscale.com/blog/2017/12/nasa-mars-climate-orbiter-metric/ :)

mikeshultz: Hungarian notation bothers me even more so I'll add a comment under protest. :black_flag:

Issue #3422, opened 2019-09-17
Mobile Onboarding Struggle


micahalcorn:
I believe that I've reported this before, but it may have only been in the 0.23.x Sheet tab, which has been overwritten at times.

**Environment**

- Origin Marketplace 0.23.15
- iPhone X
- Account 0x636e5Ca47EF50265CD254383B1F710Cd85E297ba

**Experience**

1. Install app from TestFlight
1. Create a wallet
1. Set up authentication
1. Click "Add a listing"
1. Be redirected to onboarding
1. Input email address
1. Verify email
1. Input names
1. Click'Publish"
1. Click "Got it" on the signing prompt
1. Watch confirmation spinner
1. After some time passes, be redirected back to the email step with an empty input field
1. Watch the webview reload the component again from the bottom of the screen (redundantly)
1. Input email address again
1. Don't get a verification code
1. Return to app to see name & photo step with fields already populated
1. Repeat steps 12-17
1. `stack overflow`

♻️

-----

At Origin, we use [ZenHub](https://www.zenhub.com/) to manage our engineering tasks and product development. [Download their browser extension](https://www.zenhub.com/extension) and check out our open workspace at [github.com/originprotocol/origin#zenhub](https://github.com/originprotocol/origin#zenhub).

micahalcorn:
I killed the app, started over (without deleting the app and needing a new wallet), got to 1️⃣:three:, and it reloaded to the name & photo step this time.

micahalcorn:
The resulting error is https://github.com/OriginProtocol/origin/issues/2944 ☹️

micahalcorn:
Considering this resolved 🤞

tomlinton:
This issue is still occurring, as per the demo on engineering meeting today. In that demo it kicked me back to the name entry rather than the email flow.

micahalcorn:
This affected me again today too. 😞

micahalcorn:
Closing this with the assumption that it's irrelevant with zippy identity 🤐

PR #3435, opened 2019-09-19
Removes internal identityEvents resolver reference for perfmode

mikeshultz:
### Description:

There was still an internal reference to the `identityEvents` resolver in the `deployIdentity` mutation.  This forces a remote query when perfmode is enabled.  The technique is less than ideal, but it does the job.

Ref: #2944 #3229 #3422

### Checklist:


nick:
I'm thinking this mutation should probably accept an optional argument for the old profile IPFS hash  which can then be fetched directly and updated. This is probably OK for right now as a quick fix

mikeshultz:
Interesting idea, but it wold still require whatever runs the mutation to perform this same lookup ahead of time.  Might be a good addition if whatever's running this mutation is already doing that at some point, but I'd still like to get this merged in now and it would probably still be a good fallback.

Issue #3312, opened 2019-09-09
Transaction sent to proxy that does not exist


franckc:
While Coleman was testing with (what we think) is poor connectivity, he reported hitting this error: https://github.com/OriginProtocol/origin/blob/master/dapps/marketplace/src/components/WaitForTransaction.js#L143

This would occur if getting the transaction receipt fails. I suspect it is due to a call to Alchemy that failed due to network error.

We should try to harden this code. Perhaps it could retry N times before showing an error to the user.

mikeshultz:
Is there a tx hash?

micahalcorn:
We can get it from @coleman415 but we may need to coach him through finding it in MetaMask. 🦊

franckc:
@coleman415 Here is how to get the tx hash:
- Open Metamask
- Find the transaction in the History
- Click on specific transaction to expand it and see the details
- Click on little arrow on the right of the Details. See screenshot

<img width="358" alt="Screen Shot 2019-09-09 at 10 40 57 AM" src="https://user-images.githubusercontent.com/945910/64554259-e6a37700-d2ef-11e9-945a-60a45491597d.png">

coleman415:
I think this is the tx hash (based on timestamp, I had many tx's):

https://etherscan.io/tx/0x76260ee603555df2ea71e6b89241abad8a2363fea1c5a3e310a7eaec0cc461d1

Here is my IP right now (I am using VPNs with different servers and forgot what server I was using then): 156.234.193.194

Note that it says Los Angeles but I am in China using a Hong Kong VPN server

tomlinton:
@coleman415 who is the VPN provider? It might be worth one of us setting up a VPN and trying out the DApp.

franckc:
@mikeshultz Do you mind taking a look at this issue ? First could you confirm that the DApp fails hard (as opposed to retying) if any call to getTransactionReceipt fails ?

mikeshultz:
I don't think connectivity would cause the event not found issue.  This tx just doesn't have events, which is probably weird in and of itself since I assume this should at least have a `OfferCreated`.  

This tx was supposed to be an offer, right?

micahalcorn:
Yes @mikeshultz 😕

mikeshultz:
Yep, it's an offer and a new mystery, it seems.

tx decoded:

    --> 1. transferTokenMarketplaceExecute
[...]

PR #3439, opened 2019-09-19
Repair DAI Purchases

### LINKED ISSUE #3419: Repair DAI Purchases
It sounds like https://github.com/OriginProtocol/origin/pull/3331 unintentionally prevented people from making offers using DAI. To recap, here is the current expected behavior:

**Seller Accepts DAI & ETH**
- If buyer has sufficient DAI, payment will be made using DAI
- If buyer has insufficient DAI and sufficient ETH, payment will be made using ETH **_and not converted to DAI_**
- If buyer has insufficient DAI and insufficient ETH, funding message will be displayed

**Seller Accepts DAI Only**
- If buyer has sufficient DAI, payment will be made using DAI
- If buyer has insufficient DAI and sufficient ETH, payment will be made using ETH _and automatically converted to DAI_
- If buyer has insufficient DAI and insufficient ETH, funding message will be displayed

**Seller Accepts ETH only**
- If buyer has sufficient ETH, payment will be made using ETH
- If buyer has insufficient ETH, funding message will be displayed

This will all be simplified by the move to [buyer-controlled payment methods](https://github.com/OriginProtocol/origin/issues/3383). If that isn't going to happen very soon, then we should fix this issue immediately.

😕

-----

nick:
Repair DAI Purchases

Closes #3419

- [ ] Puppeteer tests


nick:
Adding @mikeshultz as I made a few small changes to txHelper

nick:
Going to merge this now as tests pass. Will continue adding more tests and integrate any feedback in a separate PR.

mikeshultz:
What's with the divergence between `transferTokenMakeOffer` and `transferTokenMarketplaceExecute`?  Since they're the same proxy method call, why can one be sent via proxy but the other can not?  Is one considered "no proxy but needs one" and the other "proxy already exists"?

Since the mutation names seem kind of arbitrary, I'm wondering if we can clean this up somehow.  Maybe instead of basing decisions off of mutation names, the mutation itself can just make that call by setting these options?  That might be a conversation for another day, however.

Looks like a proxy would now be created in the case of #3312.  Can we consider that resolved?

PR #3447, opened 2019-09-20
Mobile registry fix

### LINKED ISSUE #3438: [Growth] Users not getting rewarded for installing the mobile app
Some users are complaining they installed the mobile app but did not the 150 OGN reward.

Here are some examples of accounts impacted:
0x0a8b61cAb23E4c54d22a5cD8C1e2E6a0Aa5Df12c
0x1BAdF402D1F9cF13cbe6D4F9297552b0844EC190
0x225C18Dd2399b7f9A1f0617DC538efF040120979

Some users also provided screenshots of the mobile app running to prove they did indeed install the app. See this [doc](https://docs.google.com/spreadsheets/d/1sjQFfBfHXrjso3vE5Zz_YNbW0UF-XZhN0BlVSY9gbSs/edit#gid=0).

tomlinton:
- Removed requirement for a device token to be available for mobile to register. This only happens when the register callback fires which is never on iOS if the user never gets a notification prompt.
- Removed in code rate limiter because we have one in nginx ingress.
- Bump mobile version
- Fix Sentry import

Closes #3438.

### DIFF
--- a/infra/notifications/package.json
+++ b/infra/notifications/package.json
@@ -1,4 +1,4 @@
-{
+{te
   "name": "@origin/notifications",
   "version": "0.0.1",
   "description": "Origin Notifications Service",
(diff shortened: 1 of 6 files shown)


tomlinton:
Haven't had much time to test the notification server side of this but it isn't too complicated. The notifications server will show errors in the logs on prod until this gets deployed.

PR #3450, opened 2019-09-20
DISABLE_GPU for puppeteer Marketplace tests 

mikeshultz:
### Description:

I was having issues getting chrome/puppeteer rendering properly on my system because Chrome/Chromium doesn't work well with hardware acceleration enabled and my Intel HD Graphics 620 graphics drivers.

This adds support for the `DISABLE_GPU=false` env var which will cause puppeteer to launch Chrome without the `--disable-gpu`.

There's a small chance using `--disable-gpu` might improve things on CircleCI(I don't know what kind of GPUs they offer) but I doubt it.  The tests still seemed to run okay, but I couldn't actually see what was going on during `test:watch` because the window never re-rendered. 

### Checklist:


dapps/marketplace/test/_services.js
```diff
@@ -38,7 +38,11 @@ export default async function() {
           width: 1280,
           height: 1024
         },
-        args: ['--no-sandbox', '--disable-setuid-sandbox']
+        args: [
+          '--no-sandbox',
+          '--disable-setuid-sandbox',
+          process.env.DISABLE_GPU === 'true' ? '--disable-gpu' : ''
```
nick: I think we can just disable it globally - no need to put behind a flag. I don't think it actually gets used in headless mode anyway according to some GitHub comments I read

mikeshultz: I think you OSX folks (and none linux+intel folks) would actually be getting some benefit from this in test performance.  This was something I explicitly needed for myself because of this stupid driver issue(I have the same problem with regular Chromium and electron apps).

nick: Using the flag didn't have a noticeable effect for me on OSX 🤷‍♂️

mikeshultz: alright, doing the reverse.  if `DISABLE_GPU` is `false` it won't use the flag.  Don't come crying to me when your CPUs melt a hole in your laptop!

Issue #3392, opened 2019-09-14
ProxyFactory create2 and blackholing of funds


mikeshultz:
This is a bit of an edge case, but does actually waste money.  If a user attempts to deploy a proxy (with the same proxyfactory nonce) when one already exists, it dies.  When it dies, it continues running the `initializer` (wrapped transaction) to `0x0000000000000000000000000000000000000000`.  Of course this isn't an actual contract, and if the the transaction carried any value(ETH), that value would be sent to the black hole that is `0x00..00`.  Example tx:

`0x8a9650f337466d6d4ecda51ec95fd500d914ad5b3e5abb50eb253eea4240f4fc`

1 event was fired:

    topics[0] 0xa38789425dbeee0239e16ff2d2567e31720127fbc6430758c1a4efc6aef29f80 // ProxyCreation
    topics[1] 0000000000000000000000000000000000000000000000000000000000000000

I propose adding a check to see if the address returned by `create2` is zero. If it is, revert.

I think uprading will be clean.  We will need to add support for multiple `ProxyFactory` contracts for anything that tracks events(e.g. listener).  Already deployed user proxies are not affected.

franckc:
CC @DanielVF @nick

DanielVF:
I'm in favor. It's quite painful the way it fails now - it basically fails silently. Not to mention wasting a pile of gas.

nick:
I think this works. Just need to verify that the predicted proxy address doesn't change if we start using a new proxy factory.

eg address of deployed proxy via factory 1 == deployed proxy via factory 2

mikeshultz:
There seems to be more agreement that this isn't urgent, dropping to a P2.

PR #3452, opened 2019-09-20
[WIP] Deal with create2 failures in ProxyFactory

mikeshultz:
### Description:

`revert` on `create2` failure in ProxyFactory proxy creation methods.

Ref #3392

### Checklist:

### DIFF
--- a/packages/contracts/contracts/proxy/ProxyFactory.s
+++ b/packages/contracts/contracts/proxy/ProxyFactory.s
@@ -49,6 +49,12 @@ contract ProxyFactory {
         assembly {
             proxy := create2(0x0, add(0x20, deploymentData), mload(deploymentData), salt)
         }
+
+        // create2 failed
+        if (proxy == address(0)) {
+            revert("proxy-deploy-failed");
+        }
+
         if (initializer.length > 0)
             // solium-disable-next-line security/no-inline-assembly
             assembly {
@@ -73,6 +79,12 @@ contract ProxyFactory {
         assembly {
             proxy := create2(0x0, add(0x20, deploymentData), mload(deploymentData), salt)
         }
+
+        // create2 failed
+        if (proxy == address(0)) {
+            revert("proxy-deploy-failed");
+        }
+
         if (initializer.length > 0) {
             uint256 value = msg.value;
             // solium-disable-next-line security/no-inline-assembly


mikeshultz:
I don't think the Crytic stuff is gonna finish, so opening this up for eyeballs now.  Did run slither, it's mostly style and efficiency stuff and nothing relevant to my changes.  I don't think it would be appropriate to fix any of that in this PR.  If anyone thinks it needs to be addressed, it can be done in another PR before we actually deploy it.

franckc:
Ah also if we decide to go ahead and update the ProxyFactory contract, it would be nice to run it through [Slither](https://github.com/crytic/slither). Just in case it finds a security issue that we should fix at the same time...

mikeshultz:
> What if as opposed to changing the ProxyFactory contract we add extra checks on the client side to verify whether there is already a proxy ?

We're already checking that now.  Not working so hot.  Never seems to fail that users somehow find a way to rapid fire transactions at the relayer, and I'm not confident we'll ever fully get rid of these.  This has been seen to burn money and I think that's worth mounting a defense against that at the contract level.

> Ah also if we decide to go ahead and update the ProxyFactory contract, it would be nice to run it through Slither.

See my previous comment.

mikeshultz:
I'm not thrilled by this myself.  But I'm less thrilled by wasting money of users and I'm not confident we can 100% prevent this on the client-side.

I'm not sure we need to keep separate versions of the contract, though I'll defer to everyone else on that.  I think keeping a configuration with the previous ProxyFactory address would be necessary, though for address prediction.

nick:
I think we should rename this V01_ProxyFactory and keep a copy of the old one so it's easy to refer to what code is deployed an in use on mainnet without having to refer to git history.

Perhaps we should hold off on deploying a copy and actually using the new ProxyFactory until some of the other fixes we have going on are in place, such as the Dai repair I hope to get merged soon. Even if we don't end up using a new ProxyFactory immediately we should keep this code around for when we eventually do a new deploy.

One potential other change I'm thinking it'd be nice to get in before deploying is having the same predicted address regardless of which factory deployed the identity.

mikeshultz:
Yeah, I'm not really in a rush to deploy this since the impact has so far been low(that we know of), but I still think burning money to `0x00..00` is a pretty big deal.

> One potential other change I'm thinking it'd be nice to get in before deploying is having the same predicted address regardless of which factory deployed the identity.

I don't think that's possible since the addresses are bassed upon the calling contract's address.  Here's a snippet for `create2` from the solidity assembly docs:

> create new contract with code mem[p…(p+n)) at address keccak256(0xff . **this** . s . keccak256(mem[p…(p+n))) and send v wei and return the new address, where 0xff is a 8 byte value, **this is the current contract’s address** as a 20 byte value and s is a big-endian 256-bit value

And since ProxyFactory isn't in-place upgradable through `selfdestruct` + `create2` I think we're out of luck.

PR #3453, opened 2019-09-20
[Notification server] Handle null deviceToken and cleanup

franckc:
### Description:
 - After the changes in the mobile app made in #3447 , the notification server /mobile/register route may get called with a null deviceToken from iOS devices. This PR update the logic to handle that case.
- Did a bunch of refactoring and cleanup in the email sending and mobile push notification code. 
- Now storing the Eth address in lower case in the mobile_registry table. This was causing a bug where mobile notification were not sent. I will cleanup the data in prod by converting all existing mobile_registry.eth_address entry to lower case in the DB.

### Checklist:


infra/notifications/src/app.js
```diff
@@ -165,68 +165,107 @@ app.post('/', async (req, res) => {
...
   }
+  logger.info(`POST /mobile/register for ${mobileRegister.ethAddress}`)
 
-  // See if a row already exists for this device/address
-  let registryRow = await MobileRegistry.findOne({
-    where: {
-      ethAddress: mobileRegister.ethAddress,
-      deviceToken: mobileRegister.deviceToken
+  if (!mobileRegister.ethAddress) {
+    return res.status(400).send({ errors: ['Invalid Eth address'] })
```
tomlinton: Maybe sometime later we clean this up with the same validation library we use in t3, not important now but would be nice

infra/notifications/src/app.js
```diff
@@ -165,68 +165,107 @@ app.post('/', async (req, res) => {
...
+    })
+
+    if (!registryRow) {
+      // Nothing exists, create a new row.
+      await MobileRegistry.create(mobileRegister)
+      res.sendStatus(201)
+      logger.debug('Added new mobile device to registry: ', req.body)
+    } else {
+      // Update existing row since permissions might have changed.
+      await MobileRegistry.upsert(mobileRegister)
```
mikeshultz: if you're doing an upsert anyway, is there any added benefit to the insert above?

infra/notifications/src/app.js
```diff
@@ -165,68 +165,107 @@ app.post('/', async (req, res) => {
...
+ */
 app.delete('/mobile/register', async (req, res) => {
-  logger.info('Call to delete mobile registry endpoint')
+  const mobileRegister = {
+    ethAddress: _.get(req.body, 'eth_address', null),
+    deviceToken: _.get(req.body, 'device_token', null)
+  }
+  logger.info(`DELETE /mobile/register for ${mobileRegister.ethAddress}`)
+
+  if (!mobileRegister.ethAddress) {
```
mikeshultz: might be worth doing a little more validation here, but not a deal breaker.

infra/notifications/src/app.js
franckc: Yes good idea. Would be nice to use that express-validator library everywhere. I added a TODO.

infra/notifications/src/app.js
franckc: Good catch. I blindly copied the existing code. Does not make sense. Fixed to do an update if permissions change.

infra/notifications/src/app.js
franckc: Good point. Added address validation.

Issue #3310, opened 2019-09-09
External Links should be opened in native browser on Origin Marketplace app


shahthepro:
When users click on any link when within the app, we should probably prevent and open that link in native browser if it is a different domain.

### Steps to reproduce
1. Go to Messages in Origin Marketplace app.
2. Send any link 
3. Click on the sent link
4. You would be redirected to that link within the app

PR #3454, opened 2019-09-21
Intercept Telegram URL

shahthepro:
Adds back telegram URL interception that was reverted in #3446 

### Checklist:


mobile/src/screens/marketplace.js
```diff
@@ -584,6 +626,7 @@ class MarketplaceScreen extends Component {
         onLoadEnd={this.onLoadEnd}
         onError={this.onError}
         onNavigationStateChange={this.onNavigationStateChange}
+        onShouldStartLoadWithRequest={this.onNavigationStateChange}
```
tomlinton: I think we can probably use onShouldStartLoadWithRequest everywhere here and not have to use the `this.goBackToDapp()` call. Do you think that'd work @shahthepro? It would be a lot cleaner.

mobile/src/screens/marketplace.js
```diff
@@ -360,18 +366,26 @@ class MarketplaceScreen extends Component {
       if (url.hostname === 'twitter.com') {
         if (url.pathname === '/intent/tweet') {
           // Natively Tweet on Android
-          this.openNativeDeepLink(
+          const success = this.openNativeDeepLink(
```
tomlinton: If we clean up the if/else logic in this method we can probably just return straight from the `this.openNativeDeepLink` call in each case.

mobile/src/screens/marketplace.js
```diff
@@ -418,6 +436,20 @@ class MarketplaceScreen extends Component {
...
+      const success = await this.openNativeDeepLink(
+        url,
+        'lastOpenTelegramProfileTime',
+        true
+      )
+      if (success) {
+        return true
+      }
+    }
+
```
tomlinton: While we are here we might as well solve #3310. Happy to do the work on that.

mobile/src/screens/marketplace.js
shahthepro: @tomlinton Yeah, I did try that one. But as I remember, it worked only on one platform (Android). It didn't work on iOS.

mobile/src/screens/marketplace.js
shahthepro: Let me try this.

mobile/src/screens/marketplace.js
shahthepro: #3310 might be a little tricky. We shouldn't break test-build links (under Developer's mode) and attestations (we redirect users for attestations on mobile).

tomlinton: Good point, let’s leave that for a separate PR.

tomlinton: Good point, let’s leave that for a separate PR.

mobile/src/screens/marketplace.js
tomlinton: Does that mean we are using two different paths here, one for iOS and one for Android? If possible we should just use the same for both (i.e. if `onShouldStartLoadWithRequest` doesn't work on iOS we probably should just use `onNavigationStateChange`).

shahthepro: Okay, going through react web view docs again, [`onShouldStartLoadWithRequest`](https://facebook.github.io/react-native/docs/webview#onshouldstartloadwithrequest) works in iOS but not on Android. Sorry, a blunder from my side.

For Android, we use the [`stopLoading()`](https://facebook.github.io/react-native/docs/webview#stoploading) method to prevent the page from continuing further. 

https://github.com/OriginProtocol/origin/pull/3454/files#diff-c0d3f54b8a00383ba9f0b8ce455b87b4R519

`stopLoading()` works both on iOS and Android. But on iOS, `onShouldStartLoadWithRequest` seemed like a cleaner way to do it.

Can change this though, if necessary.

tomlinton: The WebView package has been extracted from `react-native` and has now diverged (the WebView component will be removed from `react-native` core). We use the one here [which is a lot more up to date](https://github.com/react-native-community/react-native-webview). I believe it should work for both iOS and Android. See https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#onshouldstartloadwithrequest

tomlinton: I played around with `onShouldStartLoadWithRequest` but not sure if we can get it to work as it appears to not support async functions.

tomlinton:
This works for me on all devices now so going to merge. It is a bit hacky due to bugs in `react-native-webview` but hopefully we can clean those up when they fix the issues with `onShouldStartLoadWithRequest`.

tomlinton:
Thanks @shahthepro for all the help on this one!

PR #3471, opened 2019-09-23
[Notification server] Fix error handling

franckc:
### Description:

Due to the refactoring in #3453 the notification was not returning a response when called by the listener for an event such as ListingCreated which does not require to send any notification.

### Checklist:

### DIFF
--- a/infra/notifications/src/app.js
+++ b/infra/notifications/src/app.js
@@ -329,7 +329,7 @@ app.post('/events', async (req, res) => {
   const { event = {}, related = {} } = req.body
   const { returnValues = {} } = event
   const eventName = event.event
-  const { listing, offer } = related
+  const { listing = {}, offer = {} } = related
   const { seller = {} } = listing
   const buyer = offer ? offer.buyer : {} // Not all events have offers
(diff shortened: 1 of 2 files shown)


infra/notifications/src/app.js
```diff
@@ -329,7 +329,7 @@ app.post('/events', async (req, res) => {
   const { event = {}, related = {} } = req.body
   const { returnValues = {} } = event
   const eventName = event.event
-  const { listing, offer } = related
+  const { listing = {}, offer = {} } = related
```
mikeshultz: Is this valid? I didn't think defaults were legit for object unpacking.

mikeshultz:
I checked.  node seems okay with it.

    > x = { a: 1 }
    { a: 1 }
    > const { a = 0, b = 1 } = x
    undefined
    > a
    1
    > b
    1

infra/notifications/src/app.js
franckc: Yeah. Seems to work... ah the joy of JS ! ;)
```
> const { listing = {}, offer = {} } = {}
undefined
> listing
{}
> offer
{}

```

PR #3486, opened 2019-09-25
Ensure tx errors are passed back to mutations

nick:
txHelper errors were not getting bubbled back up to mutations resulting in un-closable modals, eg when rejecting tx signatures.

### DIFF
--- a/packages/graphql/src/mutations/_txHelper.js
+++ b/packages/graphql/src/mutations/_txHelper.js
@@ -527,7 +527,8 @@ async function sendViaWeb3({
   hashCallbacks,
   receiptCallbacks,
   confirmCallbacks,
-  mutation
+  mutation,
+  reject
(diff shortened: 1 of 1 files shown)


packages/graphql/src/mutations/_txHelper.js
```diff
@@ -746,11 +744,9 @@ export default function txHelper({
       hashCallbacks,
       receiptCallbacks,
       confirmCallbacks,
-      mutation
+      mutation,
+      reject
```
mikeshultz: Should maybe this be a try/catch around the `sendViaWeb3()` call instead of passing it as another argument?

nick: I tried that initially but for some reason the error wasn't bubbling up 🤷‍♂️

PR #3511, opened 2019-09-26
Remove unnecessary calls to EventSource to prevent logs download

nick:
Try to avoid calls to EventSource in mutations to prevent event log downloads.

There are a few more which I'll look at next but this is at least a start.


packages/graphql/src/mutations/marketplace/acceptOffer.js
```diff
@@ -12,10 +12,11 @@ async function acceptOffer(_, data) {
...
 
-  const offer = await marketplace.eventSource.getOffer(listingId, offerId)
-  if (!offer.valid) {
-    throw new Error(`Invalid offer: ${offer.validationError}`)
-  }
+  // Disable for now as this is a heavy call...
+  // const offer = await marketplace.eventSource.getOffer(listingId, offerId)
+  // if (!offer.valid) {
+  //   throw new Error(`Invalid offer: ${offer.validationError}`)
+  // }
```
mikeshultz: Could move this to that `remoteQuery` thing I added.  Probably not that important though.

franckc:
Very cool. Thanks for doing that @nick ... imho we should strongly consider completely disabling EventCache when the DApp is in performance mode... otherwise I bet you 1,000 OGN that one day or another we'll add another GraphQL query that will trigger loading those events again :)

nick:
Yea agree... once all calls are removed I can add a puppeteer test to throw an error if EventSource is called client-side while in performance mode.

Issue #3512, opened 2019-09-26
Purse: child funding transactions are never finalized


mikeshultz:
It appears purse is never figuring out that child account funding transactions are mined.  This appears to cause relayer to never fund any further children.

mikeshultz:
Could be a dupe of #3355

mikeshultz:
This appears to be blocking somewhere in here and never recovering:

https://github.com/OriginProtocol/origin/blob/0bc3c859feba2b07b41918d0e0bb82b6ab654835/infra/relayer/src/purse.js#L841-L870

mikeshultz:
This may be fixed by #3514, but it might be a while before 2 accounts need funding at the same time to verify.

mikeshultz:
The relayer doesn't appear to notice that funds arrived now.  The `_process` loop is still jacked.

PR #3514, opened 2019-09-26
Adds fetch timeout for EthGasStationProvider

mikeshultz:
### Description:

This might solve .:. #3512.  Doesn't appear that cross-fetch will timeout if the HTTP server doesn't timeout.  I suspect ethgasstation might not and this is blocking for longer than it should.

### Checklist:

### DIFF
--- a/packages/web3-provider/src/subproviders/EthGasStationProvider.js
+++ b/packages/web3-provider/src/subproviders/EthGasStationProvider.js
@@ -14,9 +14,7 @@ const GAS_PRICE_KEY = process.env.GAS_PRICE_KEY || 'average' // 'safeLow'
  */
 function timeoutFetch(url, ms = 10000) {
   return Promise.race([
-    fetch(url).then(res => {
-      return res
-    }),
+    fetch(url),
     new Promise((resolve, reject) => {
       setTimeout(() => reject(new Error('timeout')), ms)
     })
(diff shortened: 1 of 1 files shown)


packages/web3-provider/src/subproviders/EthGasStationProvider.js
```diff
@@ -5,6 +5,24 @@ const CACHE_TIME = 15000
...
+/**
+ * Wraps a fetch with a timeout
+ *
+ * @param url {string} URL to fetch
+ * @param ms {number} timeout in milliseconds
+ * @returns {object} fetch response object
+ */
+function timeoutFetch(url, ms = 10000) {
+  return Promise.race([
+    fetch(url).then(res => {
```
nick: I think you can just say `fetch(url)` here and skip out the then

mikeshultz: The relationship of `Promise` and `await` still trips me up.  Would `await` resolve nested promises?  Or would `Promise.race()`?

nick: I think you can either `await Promise.race(promise1, promise2)` or `Promise.race(promise1, promise2).then(result => doSomething())`

mikeshultz: I did a quick test to be sure and it looks like await resolves nested Promises.  Stripped off that `then()`.

PR #3515, opened 2019-09-26
Remove remaining eventSource calls in mutations

nick:
Remove validation calls from mutations as they require expensive calls to eventSource

### DIFF
--- a/packages/graphql/src/mutations/marketplace/updateListing.js
+++ b/packages/graphql/src/mutations/marketplace/updateListing.js
@@ -26,15 +26,6 @@ async function updateListing(_, args) {
     'ether'
   )
 
-  // Ensure that the new total units exceeds the sum of units purchased through
-  // all valid offers for this listing. This prevents unitsAvailable from going
-  // negative.
-  const newUnitsTotal = Number(ipfsData.unitsTotal) || 0
-  const listing = await marketplace.eventSource.getListing(listingId)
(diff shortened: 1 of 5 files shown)


packages/graphql/src/mutations/marketplace/updateListing.js
```diff
@@ -26,15 +26,6 @@ async function updateListing(_, args) {
     'ether'
   )
 
-  // Ensure that the new total units exceeds the sum of units purchased through
-  // all valid offers for this listing. This prevents unitsAvailable from going
-  // negative.
-  const newUnitsTotal = Number(ipfsData.unitsTotal) || 0
-  const listing = await marketplace.eventSource.getListing(listingId)
-  if (newUnitsTotal < listing.unitsPending) {
```
mikeshultz: Is not having this validation healthy?  What's the worst case of this check failing?

nick: Worst case the updated listing will be invalid, so we should just move the validation to the DApp as it already has access to the listing and prevent the mutation being called from there.

franckc: Sounds good to me in short-term in order to get EventCache disabled. But long-term I think it is always preferrable to have the business logic doing these kind of checks on number of units available rather than the DApp.... mayeb add a TODO ?

Issue #3186, opened 2019-09-04
Focused Search UX


micahalcorn:
When the search field is focused, there needs to be a way to cancel and return to the listings.

For example, when I first focus on the field, I can't find a way to exit without hitting `return` or choosing a filter.

**Bad** ↘
![D51C1182-DEDA-4F26-ADC4-E4918C7D10C5](https://user-images.githubusercontent.com/273937/64223911-71b4e500-ce8a-11e9-8ccc-59fc12bf373a.png)

Once there is a value in the field, we want the 🔍 to switch to a clear icon.

**Half Good** ↘
![28AC86A5-3B06-4740-8AC1-718F566A8DD6](https://user-images.githubusercontent.com/273937/64223910-71b4e500-ce8a-11e9-86f7-ae28df52ebba.png)

But all along, we want the user to be able to just close the modal and forget about it. So there should be an unfortunate second Ｘ visible the entire time.

**Mostly Good** ↘ (https://zpl.io/V0wYBQE)

<img width="335" alt="Screen Shot 2019-09-03 at 8 43 03 PM" src="https://user-images.githubusercontent.com/273937/64224187-74fca080-ce8b-11e9-95f8-a400cb5e106d.png">

Of course, there should also be a state where the field is empty with the 🔍 is still in place and the ever-persistent Ｘ is there to close the modal.

-----

At Origin, we use [ZenHub](https://www.zenhub.com/) to manage our engineering tasks and product development. [Download their browser extension](https://www.zenhub.com/extension) and check out our open workspace at [github.com/originprotocol/origin#zenhub](https://github.com/originprotocol/origin#zenhub).

franckc:
Talked with @micahalcorn - We would like to implement the "Mostly Good" UI referenced above. Mocks is here: https://zpl.io/V0wYBQE

micahalcorn:
Note that we've replaced the "clear" ✖️ icon with a link. 😐

angwangri:
I fixed the issue and pushed the code here. 
https://github.com/OriginProtocol/origin/commit/9cb034d30edefb6b14373eded48f9e0d171d9b97

micahalcorn:
Thanks @angwangri 🙂 When you're ready to submit a PR, can you add a test build here for us to check out? https://github.com/OriginProtocol/test-builds#origin-dapp-build-previews

franckc:
[...]

PR #3526, opened 2019-09-27
Fixed issue 3186

angwangri:
Link for test: https://originprotocol.github.io/test-builds/angwangri/#/


dapps/marketplace/src/pages/listings/_Search.js
```diff
@@ -260,16 +262,16 @@ require('react-styl')(`
...
         .featured-categories-wrapper
-          margin: 0 -2rem
-          overflow-x: scroll
+          margin: 0 0rem
 
         .featured-categories
           display: inline-flex
-          padding: 0 2rem
+          margin: 0 -10px
+          flex-wrap: wrap;
```
franckc: nit: the convention in our codebase is to leave the semicolons (e.g. character ";") out. please remove.

micahalcorn:
Much improved 👍 It would also be preferable for those filter icons (Rewards, Housing, etc) to be justified horizontally with equidistant horizontal margin as described [here](https://github.com/OriginProtocol/origin/issues/3186#issuecomment-535721320). I think I would also prefer the "Clear" to be right-aligned in that field on small screens instead of being padded.

micahalcorn:
@franckc let me commit a small change and then I'll merge it. 🙂

micahalcorn:
I've made a small change with 9f218f3. @angwangri will you make one more tweak to the margin/padding so that these icons light up with the nav controls?

Screen Shot 2019-09-29 at 12 52 06 PM

micahalcorn:
@angwangri let us know if you update the test build again, please. 🏗

angwangri:
@micahalcorn I fixed the issue. 
Here is the updated test-build.
https://originprotocol.github.io/test-builds/angwangri/#/

micahalcorn:
I made another change in https://github.com/OriginProtocol/origin/pull/3526/commits/e6e4c71040dc6cb546d1e4130d5e3c42d1adb42f so that we don't end up with this 👇

Screen Shot 2019-10-01 at 10 57 32 AM

Issue #3460, opened 2019-09-22
Enable Meta Transactions?


micahalcorn:
I'm unable to create a listing with a new 0.23 (18) installation and a new wallet. 😕

![IMG_8863](https://user-images.githubusercontent.com/273937/65390361-b954c100-dd12-11e9-9294-385046c3c989.PNG)

-----

At Origin, we use [ZenHub](https://www.zenhub.com/) to manage our engineering tasks and product development. [Download their browser extension](https://www.zenhub.com/extension) and check out our open workspace at [github.com/originprotocol/origin#zenhub](https://github.com/originprotocol/origin#zenhub).

tomlinton:
Relayer error for this:

```[ERROR] relayer: Proxy already exists at 0xA8D6FA177509b27A9608f09E0F4C57b9E08C2726 for user 0x23d61bF86E64fEe96BE0eb61784dEd0e0c554bbA```

This is tracked in #3179 which we thought was resolved, but will reopen.

mikeshultz:
The txHelper refactor and the work on cache resetting should have helped this...  Can you remember anything from this onboarding that might shed light on this?  This doesn't make any sense on a new install with a brand spanking new wallet.

- Did you refresh while a tx was being sent?
- Any double-prompts?
- You completely sure you didn't re-use a private key?
- Any error modals at all, ever?

For this to have happened, you must have sent a transaction(through the relayer or not), that created your proxy before this prompt.  So, any light you can shed on how you went from that to this would be helpful.

Also, what's with the stupid title of this modal?  "Enable Meta Transactions"?

tomlinton:
I didn’t know what to call it, Create Proxy With Sender Nonce seemed a bit abstract 😂

micahalcorn:
@mikeshultz I 100% did not import a key or phrase. I don't remember refreshing at all. I did publish an identity prior to having this experience, so it's possible that something about that transaction did not work? But I don't remember any other errors. 😕

micahalcorn:
@mikeshultz I was able to publish a listing just now using that account and that app installation. ✅ I basically resumed the session from three days ago. The process was probably killed, resulting in a new webview. But I didn't consciously alter anything.

mikeshultz:
If I had to guess, that's probably because you were no longer on a spoiled cache.  The app now knew about your proxy.  I'm not convinced this is fixed, but thanks for the update.

mikeshultz:
[...]

Issue #3179, opened 2019-09-03
Relayer fails to handle meta transaction because proxy already exists for account


tomlinton:
This has been reported by @joshfraser and @milachoi. There are a lot of errors in the relayer logs like:

`[ERROR] relayer: Proxy already exists at 0x... for user 0x...`

Which results in meta transactions being rejected. I'm not sure what the conditions are that would cause that, any ideas? cc @mikeshultz

tomlinton:
Possibly relevant #2893.

mikeshultz:
This is caused when a client sends a transaction to create a proxy but one
already exists. This is an error state that the relayer can not handle. If
it sent the transaction it would just revert.

>
It's likely causes by a client resending these proxy creation transactions.
Either it thinks the proxy doesn't exist, or it's indescriminately
resending transactions.  If you have reason to suspect the proxies don't
actually exist, we should investigate the relayer.

The other issue you linked is likely an error in my script but I never did
get the chance to investigate, so there's a chance it's related. If
predicted proxy address was getting derived wrong a client may be checking
the wrong address for an existing proxy with `eth_getCode`.

mikeshultz:
I looked at #2893 again.  That appears to be a non-issue.  it was due to the predicted addresses being derived on different networks(different proxy factory addresses).

Looking at that does make me wonder though if the use of memozie in these utilities used by the resolver could be making the dapp think that a proxy was not deployed.

Hypothesis:

1)  `hasProxy()` returns false before a proxy is deployed
1) Dapp deploys proxy via relayer
1) Dapp calls `hasProxy()` again, sees the proxy not deployed and tries again

I haven't dug into the dapp code yet so that's entirely unsubstantiated.  Will look a little further this afternoon.

mikeshultz:
Does this only happen on listing creation?  This logic doesn't really make sense to me:

https://github.com/OriginProtocol/origin/blob/a716c6753618759fffbb25e948c492b0279424ab/packages/graphql/src/mutations/_txHelper.js#L103-L106

tomlinton:
They look to be all be `createProxyWithSenderNonce` transactions based on the function signature in the transaction data pulled from the logs, e.g. one example below. So at some point we must be trying to create another proxy when one was already created. We just need to figure out what steps will get an account into that state. It seemed to be happening reasonably regularly at the meetup.

[...]

PR #3532, opened 2019-09-27
Don't cache falsey values for proxy checks

mikeshultz:
### Description:

Negative responses from the proxy utils are generally usually temporary, and caching them doesn't do much for us but create issues (e.g. #3179, #3460).  This changes these proxy utils to use a modified form of lodash's `memoize()` to prevent caching of falsey values.  This should reduce the chances of clients repeatedly thinking they don't have a proxy.

Also, I originally was under the impression that `memoize` would cache error states but I was unable to reproduce that behavior in tests, so that's not considered here.

### Checklist:


mikeshultz:
@nick would appreciate your review on this since you've been critical of custom caching in the past.  I really think this is the only decent way forward for proxy data because the current reset cache functionality has been plagued by race conditions and other failures.

mikeshultz:
Will add tests. :+1:

mikeshultz:
Adding tests revealed a couple of issues.  Looks good now though.  :heavy_check_mark:

packages/graphql/src/utils/memorize.js
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * memorize is essentially memoize but only caches truthy values.
+ *
+ * @param {function} the function to wrap
+ * @param {function} function to use to derive the cache key
+ * @returns {function} the wrapped funtion
+ */
+export default function memorize(func, resolver) {
+  if (
+    typeof func != 'function' ||
```
nick: `!==`

mikeshultz: Blame lodash!  will change though.  :+1:

PR #3549, opened 2019-09-28
[WIP] Bridge: Implement a route for logging events in the DB

franckc:
### Description:

Implemented a new route /utils/log for the bridge that can be called by the DApp to log any kind of event into a new "log" DB table. The plan is to use that to instrument the DApp core transaction flow.

Still WIP since I want to work a bit on the DApp side instrumentation (as a separate PR) to determine if the log table schema is good enough.

### Checklist:

### DIFF
--- a/infra/bridge/migrations/20190927220008-create-log.js
+++ b/infra/bridge/migrations/20190927220008-create-log.js
@@ -1,0 +1,36 @@
+'use strict'
+
+const tableName = 'log'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable(tableName, {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER
+      },
+      data: {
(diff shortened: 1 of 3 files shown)


tomlinton:
Sentry has support for capturing other non-exception data too. Might not be suitable but might be worth checking out.

franckc:
@tomlinton Yes, thanks for pointing the possibility of capturing Events with tags using Sentry.
For context, I'm thinking using this logging to track the various steps in the purchase transaction flow. So that for example we can get the percentage of success/failure at various stages in the pipeline.
Sentry's capabilities for querying event data are limited. Afaict it only provides a search with filters.  I'm thinking having events in the DB will be more versatile to analyze the data.
But let me work a bit on the purchase flow instrumentation in the DApp to understand better the data we'll track before making a final decision...

tomlinton:
Yep it won't really work well if we are wanting to do structured queries. As you note in the PR we might need to make the schema more transaction specific to be able to get nicer queries.

franckc:
Abandoning this since it is no longer a priority.

PR #3575, opened 2019-09-30
Relayer fallback fix

nick:
Wrap transactions with Proxy.execute if relayer attempt fails


packages/graphql/src/mutations/_txHelper.js
```diff
@@ -737,6 +720,25 @@ export default function txHelper({
       }
     }
 
+    if (shouldUseProxy === 'execute') {
```
mikeshultz: Shouldn't this be mutually exclusive from `create`?

mikeshultz: ah, nevermind, I see it now.

Issue #3574, opened 2019-09-30
Handle null transaction hash returned by web3


mikeshultz:
https://sentry.io/organizations/origin-protocol/issues/1247501267/events/848d565c124c49b7966b976ac9dcf85a/?project=1402589

I didn't think this was supposed to be a thing, but it appears some versions of web3 will return a `null` value for transaction has on the `.once('transactionHash')` event handler.

Ref #3275

mikeshultz:
Added some instrumentation with #3589.  This issue will get a unique error now and also dump the transaction to console for diagnostics.  According to other reports on the interwebs, this can be caused by a gas prices being too low for the node to accept, and possibly other transaction errors.

Issue #3573, opened 2019-09-30
Handle "user denied transaction signature"


mikeshultz:
https://sentry.io/organizations/origin-protocol/issues/1247501267/events/3d74915f2a66437f9d6a48e5ab198556/

Some(?) web3's are returning the following when a user rejects a transaction:

    {"message":"User denied transaction signature"}

This needs to be handled in `txHelper()`

Ref: #3275

tomlinton:
Most likely culprit for this is mobile. It's probably a promise that is being resolved with that message instead of rejected.

mikeshultz:
This error-like object that's bubbling up to WaitForTransaction is weird.  It appears this should happen in the mutation component where the graphql mutation would call `onError` but for some reason it's coming back in receipt polling instead for some clients.

The example event is identified as a webview, and (if it's not broken) `isOrigin` is `false`.

With #3589 this will be handled a little differently by throwing a relevant error, but not solved.

PR #3589, opened 2019-10-01
Transaction hash handling

mikeshultz:
### Description:

Small logic adjustments to how deal with known weird transaction hash values.  Some just for debugging purposes.

I was looking into potentially just closing `WaitForTransaction` when a user declines the tx instead of treating it like an error but I didn't really see a clear way of doing that.  Might require adjustments to the consumers of that component?

Ref: #3573 #3574 #3275

### Checklist:


packages/graphql/src/resolvers/web3/transactions.js
```diff
@@ -6,6 +6,7 @@ import get from 'lodash/get'
 
 export async function getTransactionReceipt(id) {
   if (!id || typeof id !== 'string' || ![66, 64].includes(id.length)) {
+    console.warn(id)
     console.warn('Invalid transaction hash')
```
nick: `console.warn('Invalid transaction hash', id)` rather than two separate warnings?

packages/graphql/src/mutations/_txHelper.js
```diff
@@ -557,7 +557,18 @@ async function sendViaWeb3({
 
   tx.once('transactionHash', async hash => {
     txHash = hash
-    if (typeof txHash !== 'string' || ![66, 64].includes(txHash.length)) {
+    if (
+      typeof txHash === 'object' &&
+      Object.prototype.hasOwnProperty.call(txHash, 'message')
```
nick: could we say `if (get(txHash, 'message'))`? Object.prototype.hasOwnProperty.call seems a bit verbose?

mikeshultz: Curious about performance(bool check vs string comparison).  Probably trivial, will change it for clarity.

mikeshultz:
Integrated your feedback and added a refetch prevention when `hash === 'pending'`.

PR #3583, opened 2019-10-01
[WIP] IPFS Hash validation for identity events

mikeshultz:
### Description:

adds IPFS hash validation to identity event processing.  I guess we got our first garbage identity update event?

Ref #3581

### Checklist:

- [ ] Test your work and double-check to confirm that you didn't break anything
- [ ] [Wrap any new text/strings with `fbt`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#wrapping-text) for translation
- [ ] If there are any new/changed strings to translate, [`npm run translate`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#integrating-translations)
- [ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)

Issue #3581, opened 2019-10-01
Create proxy inside an identity updated event


mikeshultz:
There's no guarantee this came from our systems, but this appears like a legitimate user at first glance.  This transaction makes no sense though.  `createProxyWithSenderNonce` txData was sent to an `identityUpdated` transaction call.

https://etherscan.io/tx/0xa7ff157775d675266ba49bd457302a0553bd8aca79d874b8aa9adf989032efab

Impact is currently unknown.

CC @franckc

franckc:
This is causing the event-handler to choke and not process other events.
Here are the logs from the listener:
```
[INFO]  event-listener: Processing event: blockNumber=8646689     transactionIndex=97     eventName=IdentityUpdated
[INFO]  event-listener: Processing Identity event for account 0x438950fc787a0Db93c3392Ce92Df76017436a12d
2019-10-01T04:03:36.634Z event-cache noop, current
(node:122) UnhandledPromiseRejectionWarning: undefined
(node:122) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:122) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.```

**Comment by franckc** [comment]
Was able to reproduce locally.
Steps to reproduce:
- From a psql shell:
`update listener set block_number = 8646686 where id='IdentityEvents_main';`

- From a shell:
```export DATABASE_URL=postgres://origin:origin@localhost/origin
export DISABLE_CACHE=true
export PROXY_ACCOUNTS_ENABLED=true
export ENABLE_METRICS=false
nohup node listener.js --network=mainnet --identity
```

franckc:
Found another malformed event:
```
[INFO]  event-listener: Processing event: blockNumber=8649840     transactionIndex=196     eventName=IdentityUpdated
[INFO]  event-listener: Processing Identity event for account 0xE05c6a798897C5c424cE2cfe6a3D5282Aed8D944
```

mikeshultz:
I think #3610 should take care of the bad event handling, though the source of this issue still needs to be found.

franckc:
[...]

PR #3610, opened 2019-10-02
Use the latest working IPFS hash from IdentityUpdated events

mikeshultz:
### Description:

Instead of trying to look for errors in IPFS hashes, just test them.  The only way to really be sure it works.

Closes #3583 
Ref #3581

### Checklist:


packages/ipfs/src/index.js
```diff
@@ -113,7 +113,7 @@ async function post(gateway, json, rawHash) {
 //   return decrypted.data
 // }
 
-async function getTextFn(gateway, hashAsBytes) {
+async function getTextFn(gateway, hashAsBytes, timeoutMS) {
```
nick: should we set the default for this too?

packages/graphql/src/resolvers/IdentityEvents.js
```diff
@@ -208,17 +209,29 @@ export function identity({ id, ipfsHash }) {
...
         return resolve(null)
       }
+    } else {
+      ipfsHashes.push(ipfsHash)
     }
 
-    const data = await originIpfs.get(contracts.ipfsGateway, ipfsHash)
+    // Go through each hash until we get valid data
+    let data
+    for (const hash of ipfsHashes) {
```
nick: Maybe set a max of 3 to avoid too many lookups?

packages/ipfs/src/index.js
mikeshultz: There aren't any direct consumers of this function according to my grepping. Just internal in `@origin/ipfs`.

packages/graphql/src/resolvers/IdentityEvents.js
mikeshultz: sg

mikeshultz:
Fun fact, I think this accidentally fixes a potential DoS with overlarge files as well.

PR #3591, opened 2019-10-01
Offer rejection fix

mikeshultz:
### Description:

Don't use marketplaceFinalizeAndPay if withdrawOffer is a seller rejection.  

Ref #3283 

### Checklist:

- [ ] Test your work and double-check to confirm that you didn't break anything
- [ ] [Wrap any new text/strings with `fbt`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#wrapping-text) for translation
- [ ] If there are any new/changed strings to translate, [`npm run translate`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#integrating-translations)
- [ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)

PR #3618, opened 2019-10-02
withdrawOffer by seller should sweep to buyer

mikeshultz:
### Description:

Fix suggested by @nick

closes #3591

### Checklist:


dapps/marketplace/test/multiUnit.js
```diff
@@ -171,6 +173,70 @@ export function multiUnitTests({
...
+    it('should allow a new offer to be withdrawn', async function() {
+      await withdrawOffer({ page, buyer })
+    })
+
+    /**
+     * TODO: Is this a bug?  going home first I guess kind of requires a
+     * refetch?  If we don't, the item will show "Sold Out"
+     */
+    it('should navigate back home', async function() {
+      await page.evaluate(l => {
```
nick: You can prevent a full refresh by using the hash in the url... ie `/#/` instead of just `/`

mikeshultz: Made the change, but it doesn't appear to cause a refetch which is what I was going for here.  Listing will still show as "Sold Out" in this case.

PR #3630, opened 2019-10-03
Sparrow dom/on screen console

sparrowDom:
### Description:
This PR adds the ability to see Dapp JS console within the HTML application. This is especially useful when debugging on mobile devices where accessing console might be extra tedious. Can be also used by our users for reporting problems. A few things about this: 
- used localStorage for storing logs instead of state, so it survives any possible reloads. 
- has to be turned on in /#/settings to start functioning 

**This is how it looks:**
_Turning it on in settings (is not on by default because of performance impact)_
Screenshot 2019-10-04 00 42 03
_When turned on another option appears in the navigation:_ 
Screenshot 2019-10-04 00 41 57
_And the console itself:_ 
Screenshot 2019-10-04 00 41 49

 
### Checklist:


sparrowDom:
Not sure if we can merge stuff in now because of the broken messaging...

dapps/marketplace/src/pages/nav/Nav.js
```diff
@@ -66,12 +67,67 @@ const Nav = ({
...
+  const [consoleLogConnected, setConsoleLogConnected] = useState(false)
+
+  const screenConsoleEnabled = localStorage.screenConsole
+
+  useEffect(() => {
+    if (wallet && !consoleLogConnected && screenConsoleEnabled) {
+      ConsoleLogCatcher().connect((method, logString) => {
+        const logs = JSON.parse(localStorage.getItem('capturedLogs') || '[]')
+        // only keep max 15 items in the logs
+        logs.slice(Math.max(0, logs.length - 15))
```
franckc: nit: slice accepts a negative index. So you can just use logs.slice(-15) and it will only keep the last 15 elements.

dapps/marketplace/src/pages/nav/Nav.js
```diff
@@ -66,12 +67,67 @@ const Nav = ({
...
+        logs.push({ method, log: logString })
+        localStorage.setItem('capturedLogs', JSON.stringify(logs))
```
franckc: This is rewriting the entire log in localstorage for every single line logged. I have no idea about the performance of localStorage on mobile... but if this turns out to be super slow, here is a possible option to make this a less io intensive: get the date in msec, modulo it 15 and use that as part of the local storage key. Pseudo code:
```
timestamp = Date.now()
key = 'capturedLogs' + timestamp  % 15
val = JSON.stringify( { timestamp, method, log: logString } )
localStorage.setItem(key, val)
```

Then at display time, read all the logs and sort by timestamp.
```
logs = []
for (i=0; i< 15; i++) {
   key = 'capturedLogs' + i
   logs[i] = localStorage.getItem(key)
}
logs = logs.filter(l => l).sort((l1, l2) => l1.timestamp > l2.timestamp)
```

dapps/marketplace/src/pages/nav/Nav.js
sparrowDom: that is very useful thanks!

dapps/marketplace/src/pages/nav/Nav.js
sparrowDom: That is a pretty cool optimisation @franckc. 

Did a quick performance test on my (not so flagship) iPhone6: 
```
const log = { method:'info', log: 'The purpose of this is to really test how fast various approaches to storing stuff in local storage are. This should be enough text for performance testing' }

//prepare 15 logs in storage
const largeLogEntry = []
for (let i = 0; i < 15; i++) {
  largeLogEntry.push(log)
}
localStorage.setItem(`singleCapturedLogsTest`, JSON.stringify(largeLogEntry))


const singleLogTest = () => {
  for (let i = 0; i < 15; i++) {
    localStorage.setItem(`largeCapturedLogsTest`, JSON.stringify(largeLogEntry))
  }
}

const moduloTest = () => {
  // writing only 1 log
  for (let i = 0; i < 15; i++) {
    localStorage.setItem('smallCapturedLogsTest', JSON.stringify(log))
  }
}

var t0 = Date.now();
singleLogTest();
var t1 = Date.now();
console.log("Single test took " + (t1 - t0) + " milliseconds.");

t0 = Date.now();
moduloTest();
t1 = Date.now();
console.log("Modulo test took " + (t1 - t0) + " milliseconds.");
```

Your method really works ~5times faster. The differences though for storing 15 logs was 0.2ms with optimised method and 1ms for non optimised one. I am leaning a bit towards not introducing more code complexity...

Will merge this for now. If you feel this should still be optimised let me know and I will optimise it according to your suggestion.

franckc: Thanks for benchmarking ! Yes totally agreed it is not worth the extra complexity given your benchmark result. Good call.

PR #3637, opened 2019-10-04
Defer messaging event loading

shahthepro:
Should fix the following
- [x] Loading offer event fails when offer/listing is on different contracts
- [x] Defer messaging event loading with HOC. (Otherwise it blocks the UI for sometime, since it eventSource reads from blockchain)
- [ ] ???

### Checklist:

### DIFF
--- a/packages/messaging-client/src/Messaging.js
+++ b/packages/messaging-client/src/Messaging.js
@@ -890,10 +890,14 @@ class Messaging {
    * @returns {[Object]} A sorted array of conversations by time in descending order
    */
   async getMyConvs({ limit, offset } = {}) {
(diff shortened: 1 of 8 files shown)


packages/messaging-client/src/Messaging.js
```diff
@@ -914,7 +918,7 @@ class Messaging {
 
         return conv2LastMessage.msg.created - conv1LastMessage.msg.created
       })
-      .slice(Number(offset) || 0, Number(limit) || 10)
+      .slice(startIndex, endIndex)
```
sparrowDom: nice catch, seems that previous slice had an error where end index didn't account for non 0 start index

PR #3694, opened 2019-10-10
origin-scripts temporary deployment

mikeshultz:
### Description:

To make @franckc's life easier for running rewards payout scripts.  And hopefully I won't ruin his efforts by deploying again in the future.

I think we might add similar docs and deployment config files for things like network diagnostics and volume management as we go.

### Checklist:


devops/kubernetes/misc/origin-scripts.yaml
```diff
@@ -0,0 +1,9 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  name: origin-scripts
+spec:
+  containers:
+    - name: origin-scripts
+      image: "gcr.io/origin-214503/dev/origin-cron:latest"
+      command: ['sleep', '3600']
```
tomlinton: The container will exit when this command completes, so `sleep 3600` will only keep it alive for 10 minutes. I think we probably want longer here to handle rewards payouts etc, might need replacing with an infinite loop.

devops/kubernetes/misc/origin-scripts.yaml
```diff
@@ -0,0 +1,9 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  name: origin-scripts
+spec:
+  containers:
+    - name: origin-scripts
+      image: "gcr.io/origin-214503/dev/origin-cron:latest"
```
tomlinton: Is this meant to be `origin-scripts`?

devops/kubernetes/misc/origin-scripts.yaml
mikeshultz: Does it matter if someone's connected to it?  Guess it could be an infinite bash loop?

    while true; do foo; sleep 3600 done

devops/kubernetes/misc/origin-scripts.yaml
mikeshultz: Nope.  Didn't want to setup a build process for another image when the `origin-cron` image would do just fine for this purpose.

devops/kubernetes/misc/origin-scripts.yaml
tomlinton: It doesn't care about connections, it'll just boot them off if it has finished doing its business. Yep that should work fine.

PR #3697, opened 2019-10-10
Admin: DeployMarketplace account selection fix

mikeshultz:
### Description:

The way account selection is done in the constructor prevents the state from being updated when the prop is updated.  Added handling of this case in a `componentDidUpdate` method.

This also handles some cases(geth?) where a `Proxy.owner()` call may return a zero-address.  This was specifically seen on dev and I imagine other geth JSON-RPC nodes as well.

This also sets `autoWhiltelist` to false because it's only useful in the case of the deployer account actually being the token owner, which I think is rare.

### Checklist:

### DIFF
--- a/dapps/admin/src/pages/contracts/_DeployContract.js
+++ b/dapps/admin/src/pages/contracts/_DeployContract.js
@@ -20,6 +20,16 @@ class DeployContract extends Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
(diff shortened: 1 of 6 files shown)


mikeshultz:
Going to update the other components as well before merging...

mikeshultz:
Updated the other deploy components as well, and fixed a couple of other minor things.

Issue #3360, opened 2019-09-12
Pull to refresh scroll detection broken on Android


tomlinton:
Following the merge of #3297 the pull to refresh function is disabled on Android. A JavaScript snippet gets injected that sends a request through window.postMessage to report how far down the page the user has scrolled, and only enables pull to refresh if the user is at the top of the page. Unfortunately `window.onscroll` doesn't fire in the mobile modal component  (see #2884).

The scroll detection needs an alternate approach, maybe a polling function will need to be used to detect scroll in either mobile modal or on the document generally.

sparrowDom:
@tomlinton like you predicted a pretty ugly workaround was required here. Pull request here: 
https://github.com/OriginProtocol/origin/pull/3715

tomlinton:
Fixed by the above PR!

PR #3715, opened 2019-10-14
Android pull to refresh fix

sparrowDom:
### Description:

The problem is that on Android we need to find within webView if user swiping down should allow of the trigger of refresh. The documentElement.scrollTop does not report the scroll position of modal elements. 

For this reason every 500ms a function is executed that finds all scrollable elements and checks their scrollTop value. The maximum of those numbers is reported, and only once all are 0 the refresh on swipe down is enabled.

Works in all cases tested except when: 
 - user opens `all messages` and has them scrolled a bit down
 - then opens a specific message conversation and scrolls around. The scroll down when on top here doesn't work, because the background `all messages` element is not at `scrollTop` position `0`.

Since the above case prevents the reload, it should be an ok compromise.

**Related issue:** https://github.com/OriginProtocol/origin/issues/3360#event-2633586472
### Checklist:


tomlinton:
This looks great @sparrowDom, thanks for doing it, it looks like it was a headache. Will spin it up on Android and test it out today or tomorrow!

tomlinton:
Works great thanks @sparrowDom. I ran the injected JavaScript through a linter, just to make it consistent. Also changed the react Component to a PureComponent to reduce the number of renders. It was causing a lot of renders for some reason.

This does introduce some weirdness with the onboarding modal, if you refresh and then hit the back arrow, it'll display again when you try and navigate somewhere. Maybe because of the date we inject into the URL to force refresh on Android. I'll make an issue and we can deal with this in a separate PR though.

tomlinton:
Ref https://github.com/OriginProtocol/origin/issues/3741

sparrowDom:
Nice thanks for the linting and testing Tom.

PR #3718, opened 2019-10-15
Fix group type

shahthepro:
Seems like the group type of Korean Telegram group is "supergroup" instead of "group". This makes the events from that group to be dropped


sparrowDom:
lgtm

Issue #3305, opened 2019-09-08
Error withdrawing offer


sparrowDom:
On mainnet when withdrawing offer Id: 1-000-3126-4 I got a failed transaction with this error: 
**Fix for this issue:** https://github.com/OriginProtocol/origin/pull/3725
![Screenshot_20190908-233006](https://user-images.githubusercontent.com/579910/64494761-33a43080-d291-11e9-8dab-434cd98fb701.jpg)

sparrowDom:
Not sure if this error is related: https://github.com/OriginProtocol/origin/issues/3283

mikeshultz:
I don't think it's related.  Curious error.  

CC @tomlinton

tomlinton:
This'll happen if you do a web3.eth.sendTransaction from an account that is not "unlocked" in web3, i.e. web3.eth.accounts.wallet.add or web3.eth.accounts.privateKeyToAccount has not been called with an account with the some from address.

Not sure what/why was trying to do that though, maybe in _txHelper.js.

mikeshultz:
I tested editing a listing on catcher 0.22 on Rinkeby.  Worked okay for me.  Mine went over the relayer.

micahalcorn:
Not likely specific to withdrawing 🐘

franckc:
I just tested editing a listing, which was consistently failing for me before, and it worked perfectly for me. I'm ok closing this issue if @micahalcorn can't reproduce it.

micahalcorn:
Sure 🤞

sparrowDom:
Franck could you see if you can reproduce the error using the default mnemonic: 
`candy maple cake sugar pudding cream honey rich smooth crumble sweet treat`

then going to staging dapp: 
- and navigating to active listings
- and trying to [edit the listing with id 4-000-127](https://dapp.staging.originprotocol.com/#/listing/4-000-127/edit/details)

sparrowDom:
I think I have gotten closer to the root of the problem. First of all this can only be reproduced with mobile device and rinkeby config combination - local config on mobile and rinkeby on desktop work fine.

Editing the listing above causes the `tx_helper` to send the transaction without proxy enabled with justification [no proxy specified](https://github.com/OriginProtocol/origin/blob/master/packages/graphql/src/mutations/_txHelper.js#L133) This code path doesn't execute if you create a listing on staging and edit it, since the proxy option is on.

[...]

PR #3725, opened 2019-10-15
Sparrow dom/contracts fixes

sparrowDom:
### Description:
- Fix how marketplace contracts are modified with the mobile bridge
- Refactor the `graphql/contracts.js`
- Add stability when determining marketplace version.
- fixes a bug where wallet was required for screen console to log data
- adds configuration for Android environment so when in dev mode any ip can be loaded into browser from non `https` source
- adds `localhost-rinkeby` configuration to react native in dev mode. So you can debug stuff on rinkeby using local dapp

**Fixes issue:** https://github.com/OriginProtocol/origin/issues/3305
### Checklist:


packages/graphql/src/contracts.js
```diff
@@ -275,18 +275,24 @@ export function setNetwork(net, customConfig) {
     )
   }
 
-  let marketplaceVersion
+  let largestMarketVersion
```
franckc: nit: I'd suggest calling this "latestMarketplaceVersion"

packages/graphql/src/contracts.js
```diff
@@ -449,6 +455,11 @@ function setMobileBridge() {
     context.marketplace._address
   )
 
+  Object.values(context.marketplaces).forEach(marketplace => {
+    marketplace.contract = context.marketplaceExec
```
franckc: I'm not familiar at all with this code...
Is it ok to use the same context.marketplaceExec (which is constructed on line 455 above with the same context.marketplace._address) for all the marketplaces ?
If yes, maybe add a comment so that we don't wonder again when we look at this code in the future :)

nick: `contract` should be used when making queries whereas `contractExec` should only be used when making a transaction. This is so we only use the wallet's provider when making a transaction and use our own provider otherwise.

It's so we don't just use, eg, Infura for all requests which can have data inconsistencies.

It also looks like this code is using the same marketplace contract for all marketplace versions? That would mean a tx intended for marketplace 000 would go to marketplace 001... not sure what was intended?

mikeshultz: Maybe we can document this intent in `contracts.js` somewhere.

packages/graphql/src/contracts.js
sparrowDom: agreed 👍

sparrowDom:
@mikeshultz @nick @franckc would you mind taking another look? The biggest thing is I have refactored the `contracts.js` with intent to un-spaghettize the code a bit. 

Definitely when landing this a lot of testing on staging would be required.

packages/graphql/src/contracts.js
```diff
@@ -245,82 +268,79 @@ export function setNetwork(net, customConfig) {
...
   if (isBrowser) {
     const MessagingConfig = config.messaging || DefaultMessagingConfig
     MessagingConfig.personalSign = metaMask && metaMaskEnabled ? true : false
-    if (isWebView) {
-      context.mobileBridge = OriginMobileBridge({ web3 })
-    }
     context.messaging = OriginMessaging({
       ...MessagingConfig,
-      web3,
+      web3: web3,
```
mikeshultz: nit: not really necessary

packages/graphql/src/contracts.js
```diff
@@ -336,71 +356,46 @@ export function setNetwork(net, customConfig) {
       decimals: '18',
       supply: '1000000000'
     })
+
+    if (shouldUseMobileBridge()) {
+      context.ognExec = new context.web3Exec.eth.Contract(
+        OriginTokenContract.abi,
+        context.ogn._address
+      )
+    }
```
mikeshultz: Do we need `context.ognExec` without mobile bridge?

packages/graphql/src/contracts.js
```diff
@@ -245,88 +268,79 @@ export function setNetwork(net, customConfig) {
       ethGasStation: ['mainnet', 'rinkeby'].includes(net)
     })
   }
+}
+
+function setupWeb3(config, net) {
+  const provider = process.env.PROVIDER_URL
```
franckc: nit: how about defining getEnvVar as a function outside of setupOriginProviders so that it can also be used here and in other functions in the future ?

packages/graphql/src/contracts.js
```diff
@@ -336,71 +359,46 @@ export function setNetwork(net, customConfig) {
...
       token.type === 'OriginToken' ? OriginTokenContract : TokenContract
-    const contract = new web3.eth.Contract(contractDef.abi, token.id)
+    let contract
+    if (shouldUseMobileBridge()) {
+      contract = new context.web3Exec.eth.Contract(contractDef.abi, token.id)
+    } else {
+      contract = new web3.eth.Contract(contractDef.abi, token.id)
+    }
+    //contract is used for calls (queries) whereas contractExec is used for transactions
     token.contract = contract
```
franckc: I know you did not change this. But still I thought I would point this out:
I don't understand why in case mobile bridge is use we use context.web3Exec rather than context.web as value for token.contract. I thought context.web3Exec was for supposed to be used only for transactions and not for queries...

packages/graphql/src/contracts.js
```diff
@@ -413,7 +411,7 @@ function setMetaMask() {
     context.ognExec = context.ognMM
     context.tokens.forEach(token => (token.contractExec = token.contractMM))
     context.daiExchangeExec = context.daiExchangeMM
-  } else {
+  } else if (!shouldUseMobileBridge()) {
```
franckc: nit:
 - might be cleaner check for shouldUseMobileBridge() at the top of setMetaMask method and return right away
 - add a comment explaining (if I understand correctly) that this is the case for being in an environment that is not origin mobile and not metamask.

packages/graphql/src/contracts.js
sparrowDom: Thanks for these comments. After the refactor each marketplace contract is initialised with the `mobile-bridge` provider  when in right environment and not modified after the initialisation. So this issue is fixed. 

Also added the comments to document the `[contract/web3]Exec` intent as per request of @mikeshultz .

packages/graphql/src/contracts.js
sparrowDom: yup good point reverting

packages/graphql/src/contracts.js
sparrowDom: Yeah this part is where I feel I could still do a better job with refactoring, but haven't tackled that part. This still gets set in the `setMetaMask` function: https://github.com/OriginProtocol/origin/blob/master/packages/graphql/src/contracts.js#L403-L431 
I didn't refactor it out into separate methods because `setMetaMask` gets called by exported `toggleMetaMask`. Meaning that besides working as part of the setup, it also needs to be toggle-able for on the fly enable/disable of Metamask.

sparrowDom: I am very open to suggestions how this could be more neatly solved.

packages/graphql/src/contracts.js
sparrowDom: Yup good point, early exits are nicer, also adding the comment.

PR #3726, opened 2019-10-15
Add code to not massively overestimate proxied gas prices.

### LINKED ISSUE #3393: Gas is overestimated when using a proxy
Account 0x5b2A5d1AB8a5B83C0f22cB1Df372d23946aA7d8F
Proxy 0xa95D201E84d2B50c075Fd28ffD3C1DBb0bb00C5B
Purchased: https://shoporigin.com/#/listing/1-000-3161 priced at $1

The MetaMask transaction confirmation showed me a gas estimate of $3.51 because we set the gasLimit to 1M [here](https://github.com/OriginProtocol/origin/blob/master/packages/graphql/src/mutations/_txHelper.js#L220).

Here is the [txHash](https://etherscan.io/tx/0x98cb0a3d54fe98d0b5b42f71c0932b1726bb5a5347993abc7ff62501c22256f4).

That particular makeOffer tx ended up using 201,879 gas, way lower than 1M.

DanielVF:
## Problem

Anytime we use a proxy, either creating one or just executing through, we were using a hardcoded gas limit of 1,000,000. This is ludicrously high and scares users. 

## Solution

This PR adds math to actually calculate the gas overhead needed for using proxies. Closes #3393.

I've used two data sources for checking this.

First, I instrumented the origin marketplace tests to record the proxy-less gas price estimates coming into the TX helper, and compare these with the final transaction gas amounts actually used.

Secondly, I looked at traces of both 'execute' and 'create' transactions, and manually calucated a formula to be used.

Execute trace:
http://tracer.originprotocol.com/#0xe2f935e6b55a64d9550b664221103d3714137af569dc9f2f494c81261a75f094

Create trace:
http://tracer.originprotocol.com/#0x4a45cae8c7d4f75981f0bde19ae5293528f970bca9bf0980b902a4059d5cf53e

## Further Work

This PR only adjusts how the tx helper calculates the overhead gas needed for proxying. It does not adjust the base estimates. Some of our transactions, such as finalizeOffer and withdrawOffer, are currently hardcoded a bit higher than they appear to cost in practice.

## Risks & deploying

This PR massively reduces the gas limits we send - as much as 1/20 the previous amount for some of cheaper transactions. We need to monitor the our transactions for increased gas price failures after deploying, in case I have anything wrong.

I've included a extra padding gas in each to ensure that if I'm off in my calculations or data, we should still be okay for now. If we have no problems, we can remove or reduce it.


DanielVF:
@mikeshultz I think the cost of swapAndMakeOffer is already rolled into our price -  our fixed price for makeOffer is already high, and explicitly mentions swapAndMakeOffer:

https://github.com/OriginProtocol/origin/blob/cb88a9d6b33bfdb789628cf82a9c6f75e98225e6/packages/graphql/src/mutations/_gasCost.js#L7

We've also got a unit test in the the graphql package that performs a

franckc:
CC @nick in case he has any comment...

nick:
Looks good to me.

Wonder if we can look at all past transactions (especially those via Proxies) to verify these new limits work OK?

DanielVF:
I'm taking a look at pulling actual values from live txs.

Here's a chart of "inner" method call prices to the marketplace to proxied contracts. (EG, how much gas is used only the the marketplace contract, including things the marketplace contract calls)

image

Here's a table of the max inner gas used.


method | maxGasUsed
-- | -- 
makeOffer | 198482
finalize | 118900
executeRuling | 94134
withdrawOffer | 83309
createListing | 53918
updateListingWithSender | 32475
withdrawListing | 20525
acceptOffer | 20490
dispute | 9530
updateListing | 3348
addData | 2887

I've not checked our TX's vs this data yet.

DanielVF:
I've checked our gas prices vs what I've seen proxied, and we should be good. Although I don't see a swapAndMakeOffer live in the wild here, the uniswap portion uses 70,000 gas which should fit comfortably into our current price.

https://github.com/OriginProtocol/origin/blob/master/packages/graphql/src/mutations/_gasCost.js

DanielVF:
This PR should be good to go.

franckc:
Let me merge this in....

PR #3729, opened 2019-10-16
Fix active campaign resolver

shahthepro:
Seems like `rewardStatus` in the GrowthCampaign model is not a reliable way to query or filter the active campaign. So changing that to use `CampaignRules.getStatus()` method

### DIFF
--- a/infra/growth/src/resources/campaign.js
+++ b/infra/growth/src/resources/campaign.js
@@ -42,7 +42,7 @@ class GrowthCampaign {
     const campaigns = await db.GrowthCampaign.findAll({})
     return campaigns
       .map(campaign => new CampaignRules(campaign, JSON.parse(campaign.rules)))
-      .find(campaign => campaign.status === enums.GrowthCampaignStatuses.Active)
+      .find(campaign => campaign.getStatus() === enums.GrowthCampaignStatuses.Active)
   }
 
   /**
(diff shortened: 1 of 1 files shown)


infra/growth/src/resources/campaign.js
```diff
@@ -39,18 +39,10 @@ class GrowthCampaign {
...
-    })
-
-    if (!campaign) {
-      return null
-    }
-
-    return new CampaignRules(campaign, JSON.parse(campaign.rules))
+    const campaigns = await db.GrowthCampaign.findAll({})
+    return campaigns
+      .map(campaign => new CampaignRules(campaign, JSON.parse(campaign.rules)))
```
franckc: One caveat is that this be CPU intensive, since creating a new campaign involves parsing the rules and creating a bunch of objects. And the number of campaigns in the DB will keep increasing.
If getActive is a method that's going to get called frequently then you might want to benchmark it to see how long it takes. If it takes tens of msec then I'd suggest to optimize by either:
 1. caching the result (using memoize)
 2. tweaking the DB query to filter the campaign rows to load only if campaign.startDate < now < campaign.endDate

shahthepro: @franckc As of now, this query is called only once when the bridge server is started. So I don't think it'd have much impact on the performance. However, I added the where constraint for startDate and endDate to be more precise. 

If we somehow end up using this query in the DApp (at present we don't, we get all the campaigns for the DApp) or somewhere else, we could consider storing that in cache.

PR #3734, opened 2019-10-16
Remove completely outdated API section from docs.

DanielVF:
Sniff... Yanks the entire API documentation out of the docs, since the docs were 100% out of date.


micahalcorn:
Thx @DanielVF 🙏

PR #3744, opened 2019-10-18
[Bridge] New routes for centralized identity

franckc:
First pull request? Read our [guide to contributing](https://docs.originprotocol.com/guides/getting_started/contributing.html)

### Description:

This PR adds 2 new routes to the bridge server:
 - GET /api/identity: reads an identity
 - POST /api/identity: writes an identity

Note: There is some duplicate code in event-listener and bridge for writing identity. We can't delete the event-listener code because we need to keep it running since some user may be running old version of the DApp code that stores data on the blockchain. I'll refactor/cleanup the event-listener code as a separate PR in order to keep this PR's size manageable (easier to review/reason about this way).

### Checklist:


infra/bridge/src/controllers/identity.js
```diff
@@ -0,0 +1,167 @@
...
+ *
+ *  @args {string} req.ethAddress Eth address of the user.
+ *  @returns {{identity: Object, ipfsHash: string}}
+ */
+router.get('/', identityReadVerify, async (req, res) => {
+  const ethAddress = req.query.ethAddress.toLowerCase()
+  logger.debug(`identity/read called for addr ${ethAddress}`)
+
+  // Lookup the identity associated with the eth address.
+  const identity = await Identity.findOne({ where: { ethAddress } })
```
shahthepro: Just asking for a clarification, is `ethAddress` always the owner address? If yes, shouldn't we check if it is proxy and find its owner?

infra/bridge/src/controllers/identity.js
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * TODO(franck): Authenticate the request.
+ *
+ * @args {string} req.query.ethAddress: Eth address of the user.
+ * @args {Object} req.body.identity: Identity JSON blob.
+ * @args {string} req.body.ipfsHash: IPFS hash of the identity JSON blob.
+ * @returns {{id: string}} Returns the eth addresses passed as input.
+ */
+router.post('/', identityWriteVerify, async (req, res) => {
+  const ethAddress = req.query.ethAddress.toLowerCase()
```
shahthepro: Same comment as above, What happens if `ethAddress` is a proxy address?

infra/bridge/src/controllers/identity.js
franckc: I plan on having the graphQL code always passing the owner's address... but you are right, better to make this more resilient by checking for proxy address here. Good thinking... will do !

infra/bridge/src/utils/identity.js
```diff
@@ -0,0 +1,126 @@
...
+  const postOptions = {
+    host: url.hostname,
+    port: url.port,
+    path: url.pathname,
+    method: 'POST',
+    headers: {
+      'Content-Type': contentType,
+      'Content-Length': Buffer.byteLength(data)
+    }
+  }
```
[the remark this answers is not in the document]
franckc: Yes good catch. We should have authentication for those requests. The webhook is implemented on the website at https://www.originprotocol.com/mailing-list/join.
We could either:
1. pass thu the aut token from the DApp, then have the webhook check that.
2. use a secret that we store in EnvKey and that only the bridge and website know about.
I'm thinking #2 might be simpler and gives the flexibility to have other clients that may not have the ability to generate an auth token to call the webhook (I'm thinking for ex. if in the future we need to run a backfill script).
For now I'll add a TODO and let's handle this as a separate PR...

infra/bridge/src/controllers/identity.js
franckc: Done

infra/bridge/src/controllers/identity.js
franckc: Done

infra/bridge/src/utils/identity.js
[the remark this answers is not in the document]
franckc: Added TODO

PR #3773, opened 2019-10-23
[Listener] Changes to support fast-identity data format

franckc:
### Description:

Update the listener to write data in the Identity DB table using the same format as the bridge endpoint. The goal is to be able to have both the listener and the bridge coexisting for some time.
More specifically, it changes the format of the data stored in the identity.data JSONB column.
The other columns are unchanged.

Also refactor a bit so that both the listener and the bridge use common utility methods in the identity package to avoid code duplication.

### Checklist:


infra/bridge/src/controllers/identity.js
```diff
@@ -84,26 +85,19 @@ router.post('/', identityWriteVerify, async (req, res) => {
...
-      data.identity
-    )
-    validator(
-      'https://schema.originprotocol.com/profile_2.0.0.json',
-      data.identity.profile
-    )
-    data.identity.attestations.forEach(a => {
-      validator('https://schema.originprotocol.com/attestation_1.0.0.json', a)
-    })
+    validateIdentityIpfsData(ipfsData)
```
mikeshultz: Out of curiosity, why the change from @origin/validator to this?  Shouldn't the JSON schema be the same?

infra/identity/src/utils.js
```diff
@@ -216,6 +253,81 @@ async function loadIdentityAttestationsMetadata(addresses, attestations) {
...
+ */
+async function saveIdentity(owner, ipfsHash, ipfsData, attestationMetadata) {
+  // Create an object representing the updated identity.
+  // Notes:
+  //  - By convention, the identity is stored under the owner's address in the DB.
+  //  - We omit a few fields coming from the attestation metadata such as
+  //    twitterProfile and telegramProfilesince. Those are large JSON objects
+  //    that would clutter the identity and they can be retrieved if needed
+  //    from the attestation table.
+  const blacklistedFields = ['twitterProfile', 'telegramProfile']
```
mikeshultz: Clutter?  Curious what lead to this.

infra/bridge/src/controllers/identity.js
franckc: Just a refactor. validateIdentityIpfsData still uses @origin/validator to validate the data.

infra/identity/src/utils.js
franckc: The twitterProfile for example is very verbose. It includes lot of data that I don't think we have ever queried via the identity table (for example last tweet from the user). And it makes it hard to look at the data when querying it via postgres cli. Since we were just duplicating the data we already store in the attestation table, for now I decided to not continue doing that.

PR #3784, opened 2019-10-24
Zippy identity implementation at GraphQL and DApp layers

franckc:
### Description:

- Added new DApp env var ENABLE_CENTRALIZED_IDENTITY that controls whether identity read/write queries should go against the bridge server or the blockchain
- Added new graphQL config variables centralizedIdentityEnabled and identityServer. I defined identityServer rather than using the existing bridge config variables since it is handy when running the dapp locally: bridge can point to dev.bridge.originprotocol.com which is configured to do all the attestations, while identityServer can point to a local bridge server for reading/saving identities.
- Updated packages/service to include an option to start a bridge server. This is used when starting the DApp in centralized mode via "yarn run start:centralized" in dapps/marketplace
- Updated the DApp to not show any blockchain waiting modal when using centralized identity.
- Updated graphql identity and identities resolvers to support both centralized and blockchain identities, depending on config.
 - Did some refactor at the graphQL layer and also exposed some new helper methods for getting owner/proxy addresses.
 - Added a new bridge endpoint api/identity/list to power the identities resolver in centralized mode. This not used by the DApp, only by admin. In retrospect I should have implemented this as a separate PR since it is fairly separate. Happy to create a separate PR if reviewers feel like it. Just lmk.

TODO:
 - Add some tests for centralized mode

### Checklist:


franckc:
I haven't committed unit/pupeteer tests yet (I'm working on it), but the logic code for this PR is done. I have tested it on my local host. I'm sending it out to get some early review feedback... Thanks !

dapps/marketplace/package.json
```diff
@@ -14,7 +14,7 @@
     "test:browser": "NODE_ENV=test HEADLESS=false mocha -r @babel/register --file test/setup test",
     "test:watch": "NODE_ENV=test HEADLESS=false mocha -r @babel/register --file test/setup test -w -b",
     "start": "node -r @babel/register index",
-    "start:performance": "PERFORMANCE=true NETWORK=localhost node -r @babel/register index",
+    "start:centralized": "PERFORMANCE=true ENABLE_CENTRALIZED_IDENTITY=true NETWORK=localhost node -r @babel/register index",
```
mikeshultz: env var does not match what's in the pull description.

packages/graphql/src/configs/rinkeby.js
```diff
@@ -14,6 +14,7 @@ export default {
   discovery: 'https://discovery.staging.originprotocol.com',
   growth: 'https://growth.staging.originprotocol.com',
   bridge: 'https://bridge.staging.originprotocol.com',
+  idenbtityServer: 'https://bridge.staging.originprotocol.com',
```
mikeshultz: why separate from bridge?  Seems redundant?

packages/graphql/src/resolvers/IdentityEvents.js
```diff
@@ -160,170 +198,225 @@ function getAttestations(account, attestations) {
...
+    throw new Error('identityServer server not configured')
+  }
+  const url = `${identityServer}/api/identity?ethAddress=${id}`
 
-    return filtered
-  }, [])
+  // Query the identity server.
+  const response = await fetch(url, { credentials: 'include' })
+  if (response.status === 204) {
+    // No identity found for this eth address.
```
mikeshultz: 204?  Should this not be a 404?

packages/graphql/src/configs/rinkeby.js
franckc: It is mostly for convenience when testing on local host. It allows to point bridge to https://bridge.dev.originprotocol.com and use it as an attestation server that is already configured, while pointing identityServer to a local bridge server that stores identity in a local db.

packages/graphql/src/resolvers/IdentityEvents.js
franckc: 204 is No Content
404 is Not Found
I picked 204 because the route and the request are valid but there is just no identity data yet associated with the eth address. Rather than 404 which would imply the route is invalid.

packages/services/index.js
```diff
@@ -210,10 +213,29 @@ const startRelayer = () =>
...
+  new Promise(resolve => {
+    console.log('Starting bridge server...')
+    const cwd = path.resolve(__dirname, '../../infra/bridge')
+    const startServer = spawn(`node`, ['src/app.js'], {
+      cwd,
+      stdio: 'inherit',
+      env: {
+        ...process.env,
+        DATABASE_URL: 'postgres://origin:origin@localhost/origin',
+        LOG_LEVEL: process.env.LOG_LEVEL || 'NONE'
```
shahthepro: Shouldn't the `EMAIL_ATTESTATION_AUTOPASS` env be set here? Also should we set some default values for `ATTESTATION_ACCOUNT` and `ATTESTATION_SIGNING_KEY`? Attestations may not work locally without setting these env variables.

packages/services/index.js
```diff
@@ -344,6 +367,16 @@ module.exports = async function start(opts = {}) {
     }
   }
 
+  if (opts.bridge) {
+    if (!(await portInUse(5432))) {
+      console.log('Bridge server requires Postgres to be running on port 5432')
+    } else if (await portInUse(5000)) {
+      console.log('Bridge server already started')
```
shahthepro: nit: Telegram attestation and Exchange rate APIs may not work without redis  But could add a env variable to override the requirement of redis when running locally.

packages/graphql/src/resolvers/IdentityEvents.js
mikeshultz: If we're not using status codes to refer to the URI(since it's a GET arg), then I'd argue it should still return 200.  But it's not worth holding this up or continuing to argue about it, so fine with me.

packages/services/index.js
franckc: For now I was thinking keeping this bridge setup lightweight and mostly for identity purposes.
For verification purposes, it seems easier to user bridge.dev.originprotocol.com since it is already fully configured for all attestation services.

packages/services/index.js
franckc: Let me add a warning. For now I was only thinking having a lightweight bridge setup that requires only the DB, and to be used for identity.

franckc:
Added some pupeteer tests and verified they pass.
Going to merge this in and deploy. Here is the plan:
1. deploy to staging and make sure no regression when centralized identity is disabled.
2. enable on staging and test
3. deploy to prod disabled
4. enable in prod once authentication is ready

PR #3786, opened 2019-10-25
Fix messaging status subscriptions

shahthepro:
Suppressing the event (when it is already in the same state) caused some unexpected behaviour on mobile randomly.

https://github.com/OriginProtocol/origin/compare/shahthepro/messaging-sub-fix?expand=1#diff-64fc362de5e424bf915ace27f6898007L122-L125

Also, `withMessagingStatus` HOC was not subscribed to MessagingStatusChange.


sparrowDom:
lgtm

PR #3807, opened 2019-10-29
[UniversalLogin] Add authentication to notifications server

shahthepro:
Should add authentication to notification server package.

### Checklist:


shahthepro:
@mikeshultz 

> Use of JWT with RDBMS while also storing the JWT in that DB is a bit redundant. One of the main benefits of JWT is that the necessary user data can be stored in the token itself and provided by the user, requiring no backend lookup at all. If the token decrypts with the server's key, it's valid. You appear to be doing that but doesn't that make the rest redundant?

I don't store all tokens to DB. Only the tokens that are blacklisted or revoked by user are stored to DB. While we could just remove the token on the client side while logging out, sometimes the user may want to terminate sessions on other devices. The `/revoke` endpoint can be useful in those cases. 

Also, while generating a token, I do log the user's IP and the signature data to DB (not the token, which however can be generated with the JWT secret :/). However this data is **not** read anywhere in the code. We have no purpose to this entries in DB as of now. But if we decide to restrict login if we notice suspicious behaviour (like logging in from a different IP than usual) or allowing the user look at the "Activity" of his account on Origin, this could be useful. But this is not in use at present, could be removed, if needed.

> Also, unless I missed something, Redis doesn't appear to actually be used? The data is set, but never actually fetched that I saw. Is that a dep that can be removed, or is that to be used later?

Redis is also used only for blacklisted tokens. The data is set in `revoke-token.js` and read in `verify-token.js`

mikeshultz:
> I don't store all tokens to DB. Only the tokens that are blacklisted or revoked by user are stored to DB.

I'd recommend storing a hash of the token, or a hash of the unique contents of the JWT to reduce redundancy and crypto analysis(paranoia).  Maybe even choosing one data store for revokes instead of the redundancy.  I think redis would be fine for this.  But that's just my opinion, and not something I really want to push on.

> Also, while generating a token, I do log the user's IP and the signature data to DB (not the token, which however can be generated with the JWT secret :/). However this data is not read anywhere in the code. We have no purpose to these entries in DB as of now. But if we decide to restrict login if we notice suspicious behaviour (like logging in from a different IP than usual) or allowing the user look at the "Activity" of his account on Origin, this could be useful. But this is not in use at present and could be removed, if needed.

@franckc would probably disagree with me on this, but it looks like overkill to me.  DB writes are expensive and we're getting nothing from it right now except *maybe* future analysis.  :man_shrugging:

Might make more sense if redis was used as a short term cache(<24hrs) and DB was fallback.

> The data is set in revoke-token.js and read in verify-token.js

Indeed, I missed that, thanks.

infra/notifications/package.json
```diff
@@ -26,6 +26,7 @@
   "author": "Origin Protocol Inc",
   "license": "MIT",
   "dependencies": {
+    "@origin/auth-server": "^0.1.0",
```
DanielVF: Why does this depend on the auth-server, rather than the auth-client?

packages/auth-client/test/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+    const client = new AuthClient({
+      authServer: AUTH_SERVER_HOST,
+      activeWallet: USER_ADDRESS,
+      disablePersistence: true
+    })
+
+    const { signature, payload } = signAuthMessage()
+
+    const { authToken } = await client.getTokenWithSignature(signature, payload)
```
DanielVF: This test doesn't seem to validate that the client actually sent a well formed request, just that it sent some kind of request to the right URL.

infra/auth-server/src/utils/verify-sign.js
```diff
@@ -0,0 +1,44 @@
...
+      logger.debug('Cannot use older sign', address, payload, signature)
+      return false
+    }
+
+    const recoveredAddress = Web3Eth.accounts.recover(
+      stringify(payload),
+      signature
+    )
+
+    return recoveredAddress === Web3Utils.toChecksumAddress(address)
```
DanielVF: Since this is one of two most vital checks in the code, would be nice to have a test case verifying that if our payload signature doesn't match our payload or address, that this fails correctly.

mobile/src/constants.js
```diff
@@ -154,3 +154,5 @@ export const LANGUAGES = [
   ['zh_TW', '繁體中文'],
   ['en_US', 'English']
 ].filter(l => Object.keys(TRANSLATIONS).includes(l[0]))
+
+export const AUTH_MESSAGE = 'I intent to sign in to Origin Marketplace app'
```
DanielVF: "I intent" should be "I intend"

mikeshultz:
@DanielVF  If an attacker gets a hold of a user's token, they do not get access to the contents of that token without the server's secret key, but same as a session key stored in cookies or local storage, it can be used to impersonate.  Even if they got access to the data within(impossible without server's secret key), the data isn't valid for re-use(after expiration, anyway) since the signature includes the timestamp(if I read that right).

If a secret key is compromised, all bets are off.

infra/auth-server/src/utils/verify-sign.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {String} address ETH address of the signer
+ * @param {String} signature The actual signature
+ * @param {Object} payload The data that was signed
+ * @returns {Boolean} true if `signature` is signed by `address` on `payload` in the last 30 days; false otherwise
+ */
+const verifySign = ({ address, signature, payload }) => {
+  try {
+    // TODO: Should we also check if timestamp > Date.now()?
+    const minTimestamp = Date.now() - 1000 * 60 * 60 * 24 * 30
+    if (payload.timestamp < minTimestamp) {
```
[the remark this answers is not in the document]
mikeshultz: Including the timestamp in the signature(assuming it's verified) helps prevent a signature from being re-used.

DanielVF:
@mikeshultz JWT tokens payloads are "plain-text". Three sections, a header, a payload, and a signature, each base64 encoded, each separated by a ".".

Pop open a node console and try it out:

```javascript
const jwt = require('jsonwebtoken')
const x =  jwt.sign(
            {'address': 'foo', 'signature': 'bob', payload: 'foo'},
            'BOB',
            {expiresIn: '30 days'}
)
//x is now 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhZGRyZXNzIjoiZm9vIiwic2lnbmF0dXJlIjoiYm9iIiwicGF5bG9hZCI6ImZvbyIsImlhdCI6MTU3MjM3MTI5MCwiZXhwIjoxNTc0OTYzMjkwfQ.rdveqCYYBfCdwayhRlkllF-kZqZzBpoNiT7MRbuT-TY' 

// We can see the payload by base 64 decoding, no secret needed
Buffer.from("eyJhZGRyZXNzIjoiZm9vIiwic2lnbmF0dXJlIjoiYm9iIiwicGF5bG9hZCI6ImZvbyIsImlhdCI6MTU3MjM3MTI5MCwiZXhwIjoxNTc0OTYzMjkwfQ", 'base64').toString('utf8')
// Results in '{"address":"foo","signature":"bob","payload":"foo","iat":1572371290,"exp":1574963290}'
```

infra/auth-server/src/utils/verify-sign.js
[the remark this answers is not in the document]
DanielVF: We should drastically reduce the period this eth signing timestamp is good for, perhaps to 20 minutes? Since there is no nonce, this allows generating an unlimited number of auth tokens during this period.

mikeshultz:
This is something I really should have known before, thank you @DanielVF.  Might need to go double-check my previous implementations...

JWE might be a use case here, but I'll leave that to Shah.

infra/auth-server/README.md
```diff
@@ -0,0 +1,11 @@
+# `@origin/auth-server`
+
+> TODO: description
```
tomlinton: 😁

infra/auth-server/src/utils/verify-token.js
```diff
@@ -0,0 +1,73 @@
...
+    return true
+  }
+
+  const monthAgo = Date.now() - 1000 * 60 * 60 * 24 * 30
+
+  const entry = await AuthTokenBlacklist.findOne({
+    where: {
+      authToken,
+      createdAt: {
+        // Only search the tokens blacklisted in the last month
```
tomlinton: Is there a reason for this? Not immediately clear why tokens that are blacklisted aren't blacklisted forever. Maybe worth writing why only the last month is checked in the comment.

mobile/src/PushNotifications.js
```diff
@@ -290,6 +297,62 @@ class PushNotifications extends Component {
...
+    const { wallet } = this.props
+
+    // No active account, can't proceed
+    if (!wallet || !wallet.activeAccount) {
+      console.debug('Cannot generate auth token, no active account')
+      return
+    }
+
+    // No private key (Samsung BKS account), can't proceed
+    if (wallet.activeAccount.hdPath) {
```
tomlinton: Do we need to implement something for Samsung BKS users to get an auth token?

infra/auth-server/src/models/authtoken-blacklist.js
```diff
@@ -0,0 +1,49 @@
...
+  const AuthTokenBlacklist = sequelize.define(
+    'AuthTokenBlacklist',
+    {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: DataTypes.INTEGER
+      },
+      ethAddress: {
```
tomlinton: Do we need to store any of this additional data? It would seem like the only thing we need is the `authToken` column, and a reason column might be nice.

infra/notifications/package.json
shahthepro: I could probably move the middleware to a different package. But that would still need DB and other dependencies it has now to verify the token. The other way I could think for this would be to do a HTTP request. I'm not sure how good it is to do that in a middleware though

packages/auth-client/test/index.js
shahthepro: Thanks for catching this. Will update them

packages/auth-client/src/cookie-helpers.js
```diff
@@ -0,0 +1,33 @@
...
+
+  debug('Found token data in cookies', key, tokenData)
+
+  return tokenData
+}
+
+function saveTokenCookie(walletAddress, tokenData) {
+  const key = 'authtoken:' + walletAddress
+  debug('Saving token data to cookies', key, tokenData)
+  Cookies.set(key, tokenData, {
```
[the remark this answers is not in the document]
shahthepro: Sure. Will do. As of now, this file isn't used. But, it will be used once I start with the changes on the DApp side.

infra/auth-server/src/utils/verify-sign.js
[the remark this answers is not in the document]
shahthepro: > Perhaps somewhere we should have a comment about why we have one timestamp in the payload, and another roughly identical timestamp, also for 30 days, in the, in the JWT token itself?

Will do so.

> We should drastically reduce the period this eth signing timestamp is good for, perhaps to 20 minutes?

We could do that on Origin Marketplace app. But on other wallets and DApp browsers, I don't think we want the users to have the "Sign" prompt every hour or so.

But I agree, 30 days is a long time. We should perhaps reduce it.

infra/auth-server/src/utils/verify-sign.js
shahthepro: Will add a test for this.

infra/auth-server/README.md
shahthepro: Oh, I missed the Readme files. Sorry, will add it 😅

infra/auth-server/src/utils/verify-token.js
shahthepro: As of now, we have an expiration time of 30 days. So irrespective of the status of the token, it will be invalid after 30 days from date of issuance. So, we don't have to run it against all the tokens in the DB. Kind of a short-circuit.

mobile/src/PushNotifications.js
shahthepro: Yeah, we should. I'm yet to look into that.

infra/auth-server/src/models/authtoken-blacklist.js
shahthepro: Yep, just the authToken and another status/reason column would be enough. Will update it.

infra/auth-server/migrations/20191023070742-generation-logs.js
```diff
@@ -0,0 +1,54 @@
...
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable(tableName, {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER
+      },
+      eth_address: {
```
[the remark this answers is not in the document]
shahthepro: Will add it.

infra/auth-server/migrations/20191023110802-authtoken-blacklist.js
```diff
@@ -0,0 +1,42 @@
+'use strict'
+
+const tableName = 'auth_tokens_blacklist'
```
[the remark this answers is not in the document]
shahthepro: We could have a status column on `auth_tokens_gen_logs` model too. Should work fine. Will look into that.

infra/auth-server/src/utils/revoke-token.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+const revokeToken = async (authToken, user) => {
+  try {
+    await AuthTokenBlacklist.create({
+      ethAddress: user.address,
+      authToken,
+      ipAddress: user.ip
+    })
+
+    const key = 'authtokens/' + authToken + '/revoked'
```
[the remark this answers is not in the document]
shahthepro: > I actually wonder if this is really necessary. Isn't that premature optimization ?

Yeah, might not be necessary now. I just overthought this. Should this be removed?

shahthepro:
@DanielVF 

> I'd strongly recommend reducing the amount of time a signature can be valid for generating a new token to something like 20 minutes (or whatever we need to handle time offset trickery). Certainly not days.

Thinking about this, We could have signature valid for only a few hours for generating a token but the generated token can be valid longer than that. Is that what you mean to say? 

If both the signature and the token are short-lived, it can have a very bad effect on the UX. We certainly don't want the user sign a message often. So, may have to come up with something else. 

> I think we should use public/private key signing for the JWT tokens, so that only the auth server has the signing secret.

Yep, asymmetrical encryption makes sense. Will look into that.

All of these are spot on. Thanks for bringing these up.

infra/auth-server/src/utils/verify-sign.js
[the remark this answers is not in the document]
DanielVF: I'm just talking about how long the timestamp for generating new tokens is good for - not about the length of time that the token itself is good for.

shahthepro:
@franckc @DanielVF @mikeshultz @tomlinton I have addressed most of your comments. Also, got rid of some early optimization that I did and switched from JWT to asymmetric encryption with RSA. If you want to take another pass at this, please do so.

infra/auth-server/src/controllers/tokens.js
```diff
@@ -0,0 +1,124 @@
...
+        `${user.address} trying to revoke token of ${tokenData.address}`
+      )
+      return res.status(403).send({
+        success: false,
+        errors: ['Unauthorized to revoke token']
+      })
+    }
+
+    // TODO: What if req.body.token === req.headers.authorization?
+    // Can the user revoke the token he is using for this request?
```
DanielVF: It would make sense that a user could revoke a token his is using for the revoke request, since you would want to do just that to logout.

infra/auth-server/src/utils/verify-sign.js
```diff
@@ -0,0 +1,52 @@
+'use strict'
+
+const Web3Utils = require('web3-utils')
+
+const Eth = require('web3-eth')
+const Web3Eth = new Eth()
```
DanielVF: Super picky:

Is there any reason these two lines aren't just:

```
const Web3Eth = require('web3-eth')
```

infra/auth-server/src/utils/verify-sign.js
```diff
@@ -0,0 +1,52 @@
...
+ * @param {Object} payload The data that was signed
+ * @returns {Boolean} `true` if `signature` is signed by `address`
+ *                    on `payload` in the last 15 minutes;
+ *                    `false` otherwise
+ */
+const verifySign = ({ address, signature, payload }) => {
+  try {
+    const now = Date.now()
+    const signExpLimit =
+      (parseInt(process.env.SIGN_EXPIRES_IN) || 15) * 1000 * 60
```
DanielVF: Yay! I like this time much better. We'll just have to see how accurately people's clocks are set in the wild.

We should be okay on mobile though, since those are set automaticly.

infra/auth-server/src/utils/verify-sign.js
```diff
@@ -0,0 +1,52 @@
...
+    const minTimestamp = now - signExpLimit
+
+    if (payload.timestamp < minTimestamp || payload.timestamp > now) {
```
DanielVF: Because of client side clock differences, we should probably also allow perhaps one minute or five minutes into the future.

infra/auth-server/test/index.js
```diff
@@ -0,0 +1,249 @@
...
+
+const { generateToken, verifyToken } = require('@origin/auth-utils/src/utils')
+
+const app = require('../src/app')
+
+const Eth = require('web3-eth')
+
+const stringify = require('json-stable-stringify')
+
+const Web3Eth = new Eth()
```
DanielVF: Again, could this just be `const Web3Eth = require('web3-eth')` ?

packages/auth-client/src/auth-client.js
```diff
@@ -0,0 +1,95 @@
...
+        address: this.activeWallet,
+        signature,
+        payload: data
+      })
+    })
+
+    const response = await rawResponse.json()
+
+    if (rawResponse.status !== 201 || !response.success) {
+      throw new Error(get(response, 'errors[0]', 'Something went wrong'))
```
DanielVF: Is "something went wrong" shown to the user? Does it need to be something different?

packages/auth-client/src/cookie-helpers.js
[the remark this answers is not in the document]
DanielVF: If the intention behind using cookie storage is to avoid cookies being stolen when hosted on IPFS, we should also set the path options so that the cookie is valid only on the current page and below. Otherwise, by default, js-cookie cookies are available to the entire domain.

I think it should be set to `path: ''`

packages/auth-client/test/index.js
DanielVF: Did this get added? Still looks like it doesn't actually test one of the format of the message to the server, or an actual request to the server.

packages/auth-utils/README.md
```diff
@@ -0,0 +1,46 @@
+# `@origin/auth-utils`
```
DanielVF: Nice README

packages/auth-utils/test/middleware.js
```diff
@@ -0,0 +1,172 @@
+'use strict'
```
DanielVF: Nice tests, @shahthepro

packages/auth-client/src/auth-client.js
shahthepro: At present no. But once I'm done with DApp changes, we may or may not show this.

infra/auth-server/test/index.js
shahthepro: Correct me if I'm wrong. But you could only use `web3-utils` without creating a new instance. You have to create an instance for `web3-eth`.  [Ref.](https://web3js.readthedocs.io/en/v1.2.2/)

infra/auth-server/src/utils/verify-sign.js
shahthepro: Makes sense. Added.

infra/auth-server/src/controllers/tokens.js
shahthepro: Yeah, but from what we decided, logout will be the equivalent of "forget such a token exists". When you log out, the token won't be revoked. Revoke token is to be used when you want to terminate sessions on other devices.  

However this should not affect the user from logging back in though.

packages/auth-client/test/index.js
shahthepro: Yeah, I added the body contents to the `nock` server above. This should verify that the exact contents are sent by the client. If the `auth-client` doesn't send a request to that URL, nock wouldn't capture it and the test would fail.

https://github.com/OriginProtocol/origin/pull/3807/files#diff-6b05ef2644bfa77cb6cb5803238636aaR64-R73

Seems good to me. Or did I overlook something?

packages/auth-client/src/cookie-helpers.js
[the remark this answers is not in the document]
shahthepro: Have done this. But will have ensure that `path: ''` works

shahthepro:
@DanielVF Thanks for all your comments. :)
Have made the necessary changes.

packages/auth-utils/migrations/20191103130800-authtoken-blacklist.js
```diff
@@ -0,0 +1,36 @@
...
+      },
+      reason: {
+        type: Sequelize.STRING
+      },
+      created_at: {
+        allowNull: false,
+        type: Sequelize.DATE,
+        defaultValue: Sequelize.fn('now')
+      },
+      updated_at: {
```
franckc: Isn't this an insert only table? If yes do we need the updated_at column?
Maybe in the associated model file add a comment explaining its purpose.

shahthepro: Yeah, this is more like a append-only log. We don't need the updated_at column. Can remove it.

infra/notifications/src/app.js
```diff
@@ -178,9 +180,9 @@ app.post('/', async (req, res) => {
  *  - On iOS devices, the app may call this endpoint before the user accepts
  *    to turn on notifications. So the deviceToken argument may be empty.
  */
-app.post('/mobile/register', async (req, res) => {
+app.post('/mobile/register', authMiddleware, async (req, res) => {
```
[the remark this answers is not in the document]
shahthepro: Thanks for bringing this up. Have added a non-strict middleware that does the same job as the other one (but doesn't fail on missing authorization header).

dapps/marketplace/src/pages/growth/WithEnrolmentModal.js
```diff
@@ -545,8 +545,10 @@ function withEnrolmentModal(WrappedComponent) {
 
       return (
         <Query query={profileQuery} notifyOnNetworkStatusChange={true}>
-          {({ error, data }) => {
-            if (error) {
+          {({ error, data, networkStatus, loading }) => {
+            if (networkStatus === 1 || loading) {
+              return null
+            } else if (error) {
```
shahthepro: @nick @sparrowDom Just a note: It was throwing some error randomly (i.e. when data is still loading), Had to add this to fix those tests. But I have no change in this PR that could have caused this. Either way, we should probably change all queries to check the networkStatus before accessing the data.

PR #3810, opened 2019-10-29
T3 investor types and revised schedule terms handling

tomlinton:
- Adds column for investor types
- Adds column for users who have rejected revised schedule and skips displaying the revised schedule agreement if they have rejected it or if they have previously accepted it

### Checklist:

### DIFF
--- a/infra/token-transfer-server/src/enums.js
+++ b/infra/token-transfer-server/src/enums.js
@@ -32,6 +32,9 @@ const TransferStatuses = new Enum(
   'Expired'
 )
 
+const InvestorTypes = new Enum('Advisor', 'Strategic', 'CoinList')
+
 module.exports = {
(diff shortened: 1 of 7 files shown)


infra/token-transfer-server/src/enums.js
```diff
@@ -32,6 +32,9 @@ const TransferStatuses = new Enum(
   'Expired'
 )
 
+const InvestorTypes = new Enum('Advisor', 'Strategic', 'CoinList')
```
franckc: We'll need to update [that dude](https://github.com/OriginProtocol/origin/blob/master/infra/token-transfer-server/src/scripts/import_data.js#L155) but it can be done in a separate PR since anyway that guy probably needs a refresh.

PR #3825, opened 2019-10-31
Docker build tests

mikeshultz:
### Description:

Sets up docker build tests for non-main branches to make sure changes to `package.json` will build properly before merging. To save test time, these builds will only run if the "watched" files are changed.

Currently, this only looks for differences in the packages' `package.json` but this can be adapted so other files can be included(e.g. `webpack.config.js`) if any of you think it would be useful.  lmk

There's a couple trivial `package.json` changes here just for testing.  If you feel strongly about it, I can revert them.

### Checklist:


.circleci/docker_test_builds.sh
```diff
@@ -0,0 +1,77 @@
...
+## -----
+## do_build package-name path/to/file/to/check
+################################################################################
+
+echo "Testing builds..."
+
+DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null 2>&1 && pwd)"
+REPO_COMMIT=$(git rev-parse master)
+
+if [[ $ENVKEY = "" ]]; then
```
franckc: Why is ENVKEY necessary ?
And how does it get set ?

mikeshultz: ENVKEY is necessary for at least some builds(e.g. dapp).  That's what that project I asked you to create the other day was about.  It has no data.  The env var is set in CircleCI.  Should be relatively secure, but I didn't want to take any chances.

mikeshultz:
Performance is pretty rough, which is why I created the bash script to only run the builds when they're useful.  Most PRs won't have to wait for it.  This PR has trivial changes that triggered the `origin-cron` and `origin-dapp` image builds for testing.  They last took ~10m.  You can see for yourself on CircleCI for this PR:

https://app.circleci.com/jobs/github/OriginProtocol/origin/78665

.circleci/docker_test_builds.sh
franckc: makes sense. ty for the details !

PR #3845, opened 2019-11-04
Add auth-server service

shahthepro:
### TODO:
- [x] Add EnvKey keys to secrets file


mikeshultz:
I think this looks okay.  Would like to add prom to it at some point.  Mind if I alter your branch to add a build test, @shahthepro?  

It appears origin-notifications no depends on `@origin/auth-utils`.  That Dockerfile will need to be updated to copy over the new dependency.

Also, before we merge, let me apply these when you're ready before merge.

devops/dockerfiles/origin-auth-server
```diff
@@ -0,0 +1,23 @@
...
+# Install envkey-source to make environment available for sequelize migration
+RUN curl -s -L -o envkey-source.tar.gz https://github.com/envkey/envkey-source/releases/download/v1.2.5/envkey-source_1.2.5_linux_amd64.tar.gz
+RUN tar -zxf envkey-source.tar.gz 2> /dev/null
+RUN rm envkey-source.tar.gz
+RUN mv envkey-source /usr/local/bin
+
+COPY package.json ./
+COPY yarn.lock ./
+COPY lerna.json ./
+COPY ./infra/auth-utils ./infra/auth-utils
```
franckc: Have you considered putting this under packages/auth-utils rather than in infra ? I think overall we are trying to use infra for servers and packages for any common library. Not a huge deal though and this should not be addressed in this PR but rather in #3807

tomlinton:
Looks good to me, do you want me to add the EnvKey to the encrypted secrets files?

devops/dockerfiles/origin-auth-server
shahthepro: Actually, it's under packages. auth-server is the one under infra. A mistake from my side here.

shahthepro:
@mikeshultz 
> I think this looks okay. Would like to add prom to it at some point. Mind if I alter your branch to add a build test, @shahthepro?

Sure.

> It appears origin-notifications no depends on @origin/auth-utils. That Dockerfile will need to be updated to copy over the new dependency.

Have updated it.

@tomlinton 

> Looks good to me, do you want me to add the EnvKey to the encrypted secrets files?

Thanks but Franck just gave me permissions to EnvKey and I have added them to the secret files. :)

shahthepro:
Thanks @mikeshultz

mikeshultz:
@shahthepro yeah, good catch, thanks.

PR #3849, opened 2019-11-04
origincatcher -> originmarketplace

mikeshultz:
### Description:

Made these changes last week.  I think the build.gradle one was causing me issues, and the other was just done to match but not sure if it was strictly necessary.  The play store link might actual cause an issue if we aren't able to rename the package.  lmk, @tomlinton 

### Checklist:

### DIFF
--- a/mobile/android/app/build.gradle
+++ b/mobile/android/app/build.gradle
@@ -128,7 +128,7 @@ android {
     }
 
     defaultConfig {
-        applicationId "com.origincatcher"
+        applicationId "com.originmarketplace"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode versionMajor * 10000 + versionMinor * 100 + versionPatch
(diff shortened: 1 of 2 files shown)


tomlinton:
We can't change this since it's been published on the play store with the other applicationId. It shouldn't be necessary since I can build it as is. Let me know the errors and I should be able to help out.

You can also always keep these changes locally if it helps.

mikeshultz:
You build with xcode or something, I imagine?  The `build.gradle` change was necessary for my local.  Will test out building later and come back with the results if you want.

I suppose the change to `update-prompt.js` might be garbage, though.  That just came from my grepping around.

tomlinton:
Android studio or CLI. I think @sparrowDom does the same. It may be you need to clear up cache or project settings since the last time you built it might have had a different name?

sparrowDom:
I build from x-code and AndroidStudio, so not much experience from CLI from my end. Didn't have package name problems so far...

PR #3859, opened 2019-11-05
Fix Dev EnvKey

shahthepro:
Made a minor error in the last PR #3858. This should fix it.


shahthepro:
Seems like I messed up by reusing a deleted branch. Closing this in favor of #3860

PR #3882, opened 2019-11-06
Fix max depth reached error during onboarding

sudheerDev:
### Description:
  * The condition for setState in CDU is wrong causing repeated calls

Condition should be if `finished` props is `true` then set `finished` state to `true`. 

### Checklist:


dapps/marketplace/src/pages/onboard/Email.js
```diff
@@ -40,7 +40,7 @@ class OnboardEmail extends Component {
     if (
       this.props.identityLoaded &&
       this.props.identity &&
-      !this.props.finished
+      this.props.finished
     ) {
       this.setState({
         finished: true,
```
shahthepro: It doesn't make sense. Why would we want to check if `finished` is true and then set `finished: true`?

shahthepro:
@sudheerDev Sorry for hijacking your branch. I did a quick change that fixes it for me.

dapps/marketplace/src/pages/onboard/Email.js
sudheerDev: I didn't go though it but the naming didn't felt right based on the logic. Sorry the earlier code didn't make much sense for me. i did this quick change which fixed it for me locally to work so made a PR. It could be completely wrong and i don't mind closing this even.

sudheerDev: Aah now i get it but this rather should be based on `getDerivedStateFromProps` than setting state in CDU. Adding another prop later can cause issue because it can update the component. Feel free to close and make another PR.

sudheerDev: And BTW
> It doesn't make sense.

Does not seem polite IMO.

shahthepro: > Does not seem polite IMO.

Sorry, if it offended you. Didn't mean it in an impolite way.

Thanks for reporting the issue and opening the PR though. :)

sudheerDev: @shahthepro Ya, No problem. Feel free to go ahead with the changes you made. I can give this another shot over weekend but i am still confused on why there is a state needed if it is directly based on props.

Why can't it just be inside render with 

`const isFinishedFromProps = this.props.identityLoaded && this.props.identity &&this.props.finished`. 
along with the other `finished ` state in the component dependent on `onCompleted` so in render

```
    if (isFinishedFromProps || isFinishedFromState) {
      let link = '/onboard/profile'
      const onboardCompleted = localStore.get(`${wallet}-onboarding-completed`)
```

For me the state `finished` seems confusing. I don't see a reason for having 
It seems like it is only needed for the `onCompleted`. 

One more question how is isFinished being fed to the component?

shahthepro: Okay, now I get it. I don't think there is a `finished` prop that is being passed down. It should have been `!this.state.finished` in the if constraint. Perhaps, that caused it in the first place.

shahthepro:
@sudheerDev Closing this, I did another fix in one of other PRs and it works fine on staging. Thanks for looking into this. :)

PR #3888, opened 2019-11-07
T3 configuration refactor

tomlinton:
- Switching to a single EnvKey instead of a separate one for the client/server because it has multiple values in common (unlock dates, lockup bonuses and durations) and the potential for errors in configuring is high (forgot to change both values)
- Introduce separate configuration for employee and investor unlock dates
- Added more tests for unlock dates for lockups and withdrawals
- Renamed some configuration variables for clarity
- Moved shared configuration variables to shared.js in the server

### Checklist:


infra/token-transfer-client/src/constants/index.js
```diff
@@ -1,9 +1,9 @@
-import moment from 'moment'
+import {
+  employeeUnlockDate,
+  investorUnlockDate,
+  lockupBonusRate
+} from '@origin/token-transfer-server/src/shared'
```
franckc: The client having a dependency on the server will cause it to pull back-end packages such as sequelize. I don't think that's a huge deal. Just pointing this out.

infra/token-transfer-client/src/components/pages/WithdrawalHistory.js
```diff
@@ -44,6 +44,9 @@ const WithdrawalHistory = props => {
     )
   }
 
+  const unlockDate = props.user.employee
```
franckc: Super nit: in order to avoid code duplication, might be worth putting this code in a util function that takes a user as arg and returns the unlock date.

infra/token-transfer-server/src/shared.js
```diff
@@ -113,6 +114,20 @@ function calculateWithdrawn(transfers) {
   }, BigNumber(0))
 }
 
+const employeeUnlockDate = process.env.EMPLOYEE_UNLOCK_DATE
+  ? moment.utc(process.env.EMPLOYEE_UNLOCK_DATE)
+  : moment.utc('2020-01-01')
```
franckc: Paranoia: would it be worth adding a check that if NODE_ENV === 'production' then we crash rather than using that default value? Same question for the other constants below.

infra/token-transfer-client/src/constants/index.js
tomlinton: I have a [separate file](https://github.com/OriginProtocol/origin/blob/d6fe10694d3067b56c175cd63e51e1afc51f5497/infra/token-transfer-server/src/shared.js) in the server that contains all the shared code.  I have to be very careful with the imports for the reason you mentioned (leave sequelize out, etc). This dependency exists elsewhere in the client for sharing the vesting logic between client and server. It works well enough, but can cause problems if you add any extra dependencies in that file. Open to alternatives for sharing the code though, perhaps there is a cleaner way.

infra/token-transfer-client/src/components/pages/WithdrawalHistory.js
tomlinton: Good idea 👍

infra/token-transfer-server/src/shared.js
tomlinton: This is the shared file I mentioned in the comment above. Because this is imported by the client you can't do things like exit early because its not a node target. The HTTP bundle has no concept of things like `process.exit()`.

PR #3921, opened 2019-11-09
Bug fixes

### LINKED ISSUE #3914: Docker auth_server broken
Somthing is wrong when [this part of code](https://github.com/OriginProtocol/origin/blob/master/packages/auth-utils/src/helpers/index.js#L21-L24) is trying to encrypt the payload. 
Screenshot 2019-11-08 11 53 08

Have generated a fresh set of certificates, but couldn't figure out why the signature fails.

shahthepro:
- Revert the change that caused #3912 - Will add GraphQL subscriptions remove polling and then readd this check
- Fix new line characters in dockerfile - Fix #3914 
- If user doesn't have sufficient balance in any of accepted tokens, don't default to anything.


dapps/marketplace/src/components/WithPrices.js
```diff
@@ -127,7 +127,7 @@ const WithPrices = ({
       } else {
         // User doesn't have sufficient balance in any of the accepted tokens
         // Fallback to ETH or any accepted token
-        suggestedToken = get(listing, 'acceptedTokens[0].id', 'token-ETH')
+        suggestedToken = get(listing, 'acceptedTokens[0].id', null)
```
franckc: nit: update comment above to not mention anymore the fallback to ETH.

PR #3934, opened 2019-11-11
Update T3 vesting logic

tomlinton:
- Add separate vesting logic for investors vs employees
- Change vesting schedule diagrams
- Remove vesting interval column from grant table and convert all employee vesting intervals to months (I don't think we need optionality here any more and it makes things more complex). 

### Checklist:


infra/token-transfer-server/src/lib/vesting.js
```diff
@@ -21,93 +21,123 @@ function momentizeGrant(grant) {
...
+  // Add initial vest
+  vestingSchedule.push({
+    amount: BigNumber(grant.amount)
+      .times(initialVestPercentage)
+      .div(100),
+    date: grant.start.clone(),
+    vested: grant.start <= now
+  })
+
+  // First quarterly vest
```
franckc: nit: why doing this outside of the loop below?
could we have a single loop
```
const vestingDate = firstQuarterlyVestDate
for (let i=1; i<=8; i++) {
  vestingSchedule.push({
    amount: quarterlyVestingAmount,
    date: vestingDate.clone(),
    vested: vestingDate <= now
  })
  vestingDate.add(3, 'months')
}
```

infra/token-transfer-server/src/lib/vesting.js
```diff
@@ -21,93 +21,123 @@ function momentizeGrant(grant) {
...
-    const vested = vestingEvents(grant).slice(0, threshold + 1)
-    return vested.length ? Math.round(BigNumber.sum(...vested)) : BigNumber(0)
+  // Calculate initial vest percentage granted on grant start date
+  const initialVestPercentage = 6
+  // Time after which regular quarterly vesting begins
+  const quarterlyVestDelayMonths = 4
+  const firstQuarterlyVestDate = grant.start
+    .clone()
+    .add(quarterlyVestDelayMonths, 'months')
+  const quarterlyVestingPercentage = (100 - initialVestPercentage) / 8
```
franckc: Regarding the calculation of quarterlyVestingPercentage  and quarterlyVestingAmount, since I think we had decided to not support decimal, should this be rounded down so that it is an integer? Then the last vesting amount would need to be adjusted so that the total vested amounts equals the grant amount

tomlinton: Right you are sir, I need to add another round of tests for employee/investor vesting that doesn't round nicely like the values I've chose

infra/token-transfer-server/src/lib/vesting.js
tomlinton: 👍

Issue #3940, opened 2019-11-11
Listener fails to process OfferCreated events


franckc:
The listener is crashing when processing OfferCreated events.
The following stacktrace is found in the logs:
```
[INFO]  event-listener: Processing event: blockNumber=8878583     transactionIndex=147     eventName=OfferCreated
[ERROR] event-listener: { Error: GraphQL error: fetch is not defined
GraphQL error: fetch is not defined
    at new ApolloError (/Users/franck/src/origin/node_modules/apollo-client/bundle.umd.js:92:26)
    at /Users/franck/src/origin/node_modules/apollo-client/bundle.umd.js:1587:34
    at /Users/franck/src/origin/node_modules/apollo-client/bundle.umd.js:2007:15
    at Set.forEach (<anonymous>)
    at /Users/franck/src/origin/node_modules/apollo-client/bundle.umd.js:2005:26
    at Map.forEach (<anonymous>)
    at QueryManager.broadcastQueries (/Users/franck/src/origin/node_modules/apollo-client/bundle.umd.js:2003:20)
    at /Users/franck/src/origin/node_modules/apollo-client/bundle.umd.js:2130:19
    at Object.next (/Users/franck/src/origin/node_modules/zen-observable/lib/Observable.js:308:23)
    at notifySubscription (/Users/franck/src/origin/node_modules/zen-observable/lib/Observable.js:130:18)
    at onNotify (/Users/franck/src/origin/node_modules/zen-observable/lib/Observable.js:165:3)
    at SubscriptionObserver.next (/Users/franck/src/origin/node_modules/zen-observable/lib/Observable.js:219:7)
    at /Users/franck/src/origin/node_modules/apollo-client/bundle.umd.js:1101:36
    at Set.forEach (<anonymous>)
    at Object.next (/Users/franck/src/origin/node_modules/apollo-client/bundle.umd.js:1100:21)
    at notifySubscription (/Users/franck/src/origin/node_modules/zen-observable/lib/Observable.js:130:18)
    at onNotify (/Users/franck/src/origin/node_modules/zen-observable/lib/Observable.js:165:3)
    at SubscriptionObserver.next (/Users/franck/src/origin/node_modules/zen-observable/lib/Observable.js:219:7)
    at /Users/franck/src/origin/node_modules/apollo-link-schema/lib/index.js:24:30
```

Note: I haven't confirmed whether this is specific to OfferCreated events or also happens for other type of marketplace event.

[...]

PR #3942, opened 2019-11-11
Use node-fetch rather than fetch in graphql identity code

franckc:
### Description:

Fix for #3940 

The listener process imports and runs the graphql package.
Some of the queries issues by the listener to get the details about an offer involve getting data about a the buyer and seller. This was triggering the execution of the graphql identity resolver which makes a call to fetch() for getting the identity from the bridge server.
fetch() is not a supported call in Node.js which was causing an error.

### Checklist:


shahthepro:
LGTM

PR #3975, opened 2019-11-14
Add T3 column for users who abstained from voting on the revised vesting schedule

tomlinton:
We needed a way to represent abstinence from voting on the revised vesting schedule.  These users should not be shown the revised vesting schedule to accept.

### DIFF
--- a/infra/token-transfer-client/src/utils/index.js
+++ b/infra/token-transfer-client/src/utils/index.js
@@ -7,7 +7,14 @@ export const getNextOnboardingPage = user => {
     return null
   } else if (!user.termsAgreedAt) {
     return '/terms'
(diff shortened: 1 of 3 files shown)


infra/token-transfer-client/src/utils/index.js
```diff
@@ -7,7 +7,14 @@ export const getNextOnboardingPage = user => {
     return null
   } else if (!user.termsAgreedAt) {
     return '/terms'
-  } else if (!user.revisedScheduleAgreedAt && !user.revisedScheduleRejected) {
+  } else if (
+    // Already agreed, don't display revised schedule
+    !user.revisedScheduleAgreedAt &&
```
franckc: We could also have an enum with 3 values: Agreed, Rejected, Abstained.
And a separate agreedAt column with a date.

Probably not worth the extra complexity.

PR #3989, opened 2019-11-15
Allow unknown unlock date in T3

tomlinton:
Allows setting of unlock date env vars to `true` with a generic unlock message rather than a countdown for situations where the unlock date is not known.

### Checklist:


infra/token-transfer-server/src/shared.js
```diff
@@ -134,13 +134,19 @@ function lockupHasExpired(lockup) {
   )
 }
 
-const employeeUnlockDate = process.env.EMPLOYEE_UNLOCK_DATE
+const employeeUnlockDate = moment(
+  process.env.EMPLOYEE_UNLOCK_DATE,
+  'YYYY-MM-DD'
+).isValid()
```
franckc: If we set the env var by mistake to an invalid date it will default to true.
So I wonder if we should explicitly test for process.env.EMPLOYEE_UNLOCK_DATE === 'true'?

nit: also the value 'true' is slightly confusing. wonder if it would make sense to use the value 'soon' to improve readability.

tomlinton: I think that is the correct behaviour, if we set it badly it'll default to a locked state.

I think I'll change it to use undefined to represent locked with an unknown unlock date which is probably the most readable.

PR #3996, opened 2019-11-15
Partner referral

mikeshultz:
### Description:

Partner referral funnel growth and mobile app changes.  

1) originprotocol.com stores a referral code in the clipboard
1) the app checks for it and sends it during device registration
1) growth pulls config from originprotocol.com to verify code and reward

Example rule I used for testing with the November campaign:

        {
          id: 'PartnerReferral',
          class: 'PartnerReferralEvent',
          config: {
            eventType: 'PartnerReferral',
            reward: {
              amount: tokenToNaturalUnits(1000),
              currency: 'OGN'
            },
            limit: 1,
            visible: true,
            nextLevelCondition: false,
            scope: 'campaign',
            statusScope: 'user'
          }
        },

The amount doesn't matter since it's derived from the config on originprotocol.com.

### Checklist:


mikeshultz:
I'm going to follow this PR up with the pages/screens to force users into rewards signup.

mikeshultz:
https://github.com/OriginProtocol/origin-website/pull/445 is a dependency for this PR.  **DO NOT MERGE**

mobile/src/AppContainer.js
```diff
@@ -49,6 +50,17 @@ class MarketplaceApp extends React.Component {
...
   }
 
+  /* Check the clipboard for a referral code
+   */
+  checkReferral = async () => {
+    const clipData = await Clipboard.getString()
+    if (clipData && clipData.startsWith(REFERRAL_PREFIX)) {
+      const referralCode = clipData.slice(REFERRAL_PREFIX.length)
+      console.debug(`referral code found: ${referralCode}`)
+      this.props.setReferralCode(referralCode)
```
tomlinton: Looks good, in some other similar code we have we clear the clipboard, but probably not any reason that is necessary

infra/growth/src/resources/rules/partnerReferralRule.js
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  /**
+   * Retrieve all campaign configuration
+   */
+  async _getConfig() {
+    const url = `${PARTNER_CONF_URL}/campaigns.json`
+    const res = await fetch(url)
+    if (res.status !== 200) {
+      throw new Error('Failed to fetch campaigns JSON')
```
franckc: wonder if we should log the error but not crash?
This way if for some reason originprotocol.com is down, it won't impact rewards (it'll use whatever last version of the config was fetched).

infra/growth/src/resources/rules/partnerReferralRule.js
```diff
@@ -0,0 +1,194 @@
...
+      return
+    }
+
+    this.validCodes = Object.keys(config)
+
+    if (this.validCodes.length === 0) {
+      return
+    }
+
+    // Get the rewards for the code
```
franckc: Should PARTNER_REWARDS be cleared ? So that in case we decide to remove a code in the config, it does not stick around.

infra/growth/src/resources/rules/partnerReferralRule.js
```diff
@@ -0,0 +1,194 @@
...
+
+  /**
+   * Calculate potential reward
+   * @param {string} ethAddress
+   * @param {Object} identityForTest - For testing only.
+   * @returns {Promise<Reward>}
+   */
+  async getReward() {
+    /**
+     * TODO: This isn't exactly a fixed amount, but most likely one user won't
```
franckc: Maybe explain in the comment that the "potential" reward will not be shown to the user on the UI (since the reward only shows up AFTER the user completed the action), so the value does not matter. Maybe the value should actually be zero, just in case a user queries the data via graphql, so that they don't get upset if not getting 1k OGN?

infra/growth/src/resources/rules/partnerReferralRule.js
```diff
@@ -0,0 +1,194 @@
...
+      return zero
+    }
+
+    logger.debug(`Found ${events.length} matching events for ${ethAddress}`)
+
+    const rewards = []
+    const seenCodes = []
+
+    /**
+     * User can ear multiple partner rewards, but only one from an individual
```
franckc: typo: earn

infra/growth/src/resources/rules/partnerReferralRule.js
```diff
@@ -0,0 +1,194 @@
...
+     * partner.  Make sure to de-dupe as this could happen multiple times.
+     */
+    events.forEach(ev => {
+      if (ev.type !== 'PartnerReferral' || !PARTNER_REWARDS[ev.customId]) return
+
+      if (PARTNER_REWARDS[ev.customId].currency === 'ogn') {
```
franckc: nit: will OGN be lowercased in the config? usually currency symbols are upper case. wonder if we should lowercase the test in case community team does not pay attention to the casing in the config.

mikeshultz:
Just had to add a bunch of extraneous linting from other parts of the repo for standard push hook verify to pass... so there's some pollution.

> Would be good to add a couple of unit test to make sure the calculation works as expected.

I was looking at that, it looks like all the tests are for the monthly campaigns?

infra/growth/src/resources/rules/partnerReferralRule.js
mikeshultz: :ear:

tomlinton:
I put a PR up for that linting stuff, we need to make sure everyone updates prettier

infra/growth/src/resources/rules/partnerReferralRule.js
mikeshultz: My weak knowledge of growth means I don't know exactly how this is used, so I'm just going to quote you in a comment, then.  Will also set it to zero. :man_shrugging:

infra/growth/src/resources/rules/partnerReferralRule.js
mikeshultz: Sure, if you think it's a possiblity.

infra/growth/src/resources/rules/partnerReferralRule.js
mikeshultz: Do you think anyone will see the logs?  If it can't load the config, it won't work.

mikeshultz: Made this change though...

franckc:
Regarding tests, yes currently the only tests are for monthly campaigns...
Either feel free to update the November tests or to create December tests.
Alternatively create a separate non-monthly test.

PR #3998, opened 2019-11-15
Final 𝕿𝟥 Copy Scrub

micahalcorn:
My gift to you 🦠

### DIFF
--- a/infra/token-transfer-client/src/components/pages/LockupConfirm.js
+++ b/infra/token-transfer-client/src/components/pages/LockupConfirm.js
@@ -52,7 +52,8 @@ const LockupConfirm = props => {
           rel="noopener noreferrer"
         >
           contact support
-        </a>.
+        </a>
+        .
       </h1>
     </div>
   )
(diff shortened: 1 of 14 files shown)


infra/token-transfer-client/src/components/pages/LockupConfirm.js
```diff
@@ -45,7 +45,15 @@ const LockupConfirm = props => {
   return (
     <div className="text-center p-5 text-muted">
       <h1>
-        An error occurred confirming your token lockup, has the token expired?
+        An error occurred confirming your token lockup. Please try again or{' '}
```
tomlinton: 99% of the time this will be an expiry, you sure we don't want to mention something about expiry?

micahalcorn: How about something like the following:

> An error occurred with your token lockup. Please try again and be sure to confirm the email within 10 minutes of receiving it.
>
> Contact Support

⌛️

tomlinton: Yea that sounds good 👍

PR #4018, opened 2019-11-19
DApp Bug Fixes

shahthepro:
[Link to test build](https://originprotocol.github.io/test-builds/shah/#/)

Should fix and close the following issues:
- #3433
- #3620 
- #3761 
- #3928 

### Checklist:

- [ ] Test your work and double-check to confirm that you didn't break anything
- [ ] [Wrap any new text/strings with `fbt`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#wrapping-text) for translation
- [ ] If there are any new/changed strings to translate, [`npm run translate`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#integrating-translations)
- [ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)

PR #4054, opened 2019-11-25
Defer telegram status check on mobile

### LINKED ISSUE #3984: Error: Objects are not valid as a React child (found: Error: GraphQL error: Failed to fetch).
Sentry Issue: [ORIGIN-DAPP-NE](https://sentry.io/organizations/origin-protocol/issues/1317411767/?referrer=github_integration)

```
Error: Objects are not valid as a React child (found: Error: GraphQL error: Failed to fetch).
  at Error: Minified React error #31; visit  (/docs/error-decoder.html)
  at Pa (/app.ff110214.js:93:53042)
  at None (/app.ff110214.js:93:57836)
  at Hs (/app.ff110214.js:93:66663)
  at Mo (/app.ff110214.js:93:106584)
...
(15 additional frame(s) were not displayed)
```

shahthepro:
[Link to test build](https://originprotocol.github.io/test-builds/telegram)

I added back the "Verify" button in PR #4018 to temporarily fix #3984. However, it seems like the verify button caused more confusion. This PR should remove that and not reintroduce #3984. 

Queries running actively, when the app is being suspended, seem to be terminated. However, when defering the query execution, it works just fine when app is returned to the active state.

Also, I had to disable Telegram Attestation on mobile wallets (other than Origin Marketplace app), since most wallets don't seem to open the telegram app or open the link in native browser. This causes the user to lose the state when swiping back, which makes it impossible to do the attestation. Telegram doesn't have a redirect or callback, so we don't have much choice here.

Alternatively, for Origin Marketplace app, we could use the [AppState](https://facebook.github.io/react-native/docs/appstate) to query when the user has returned from the app. But it will need a new version to be released to the store and we would end up with more complex code.

### Checklist:


micahalcorn:
For the record, I'm very much in favor of abandoning the Telegram attestation on other mobile wallets if it means that it will work better on our app. 👍

PR #4069, opened 2019-12-01
Update to T3 transfer logic

tomlinton:
- Moves T3 transfers out of a script and into a function.
- Adds tests for transfer logic
- Adds delay for large transfers


infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -169,21 +168,20 @@ async function confirmTransfer(transfer, user) {
...
-  // Setup token library. tokenMock is used for testing.
-  const token = tokenMock || new Token(networkId)
+  await transfer.update({ status: enums.TransferStatuses.Processing })
 
+  // Setup token library
+  const token = new Token(networkId)
   // Send transaction to transfer the tokens and record txHash in the DB.
   const naturalAmount = token.toNaturalUnit(transfer.amount)
   const supplier = await token.defaultAccount()
   const txHash = await token.credit(transfer.toAddress, naturalAmount)
```
franckc: token.credit may throw an exception for example if the hotwallet does not have enough funds.
should we wrap this call in a try/catch statement and set the transfer status to failed?

infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -204,7 +202,6 @@ async function executeTransfer(transfer, opts) {
     case 'failed':
       transferStatus = enums.TransferStatuses.Failed
       eventAction = TRANSFER_FAILED
-      failureReason = 'Tx failed'
```
franckc: why removing this? is it because it is redundant with the eventAction and does not add much more info?

infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -229,12 +222,13 @@ async function executeTransfer(transfer, opts) {
     const event = {
       userId: user.id,
       action: eventAction,
-      data: JSON.stringify({
-        transferId: transfer.id
-      })
+      data: {
+        transferId: transfer.id,
+        failureReason: failureReason
```
franckc: this is conditionally set at line 23 below so remove?

infra/token-transfer-server/src/tasks/transfer.js
```diff
@@ -0,0 +1,113 @@
...
+  clearWatchdog()
+
+  checkWalletBalance()
+}
+
+const checkWalletBalance = async () => {
+  const token = new Token(networkId)
+  const balance = token.toTokenUnit(token.balance())
+
+  if (balance < 10) {
```
franckc: Threshold of 10 OGN seems very low. Should this be an env var and we set it to several hundred thousand OGN at least?

infra/token-transfer-server/src/tasks/transfer.js
```diff
@@ -0,0 +1,113 @@
...
+    const webhookData = {
+      embeds: [
+        {
+          title: `Wallet balance is \`${balance}\`, should be \`${balance}\` OGN`
```
franckc: replace balance with configurable threshold

infra/token-transfer-server/src/tasks/transfer.js
```diff
@@ -0,0 +1,113 @@
...
+
+  for (const transfer of transfers) {
+    logger.info(`Processing transfer ${transfer.id}`)
+    const result = await executeTransfer(transfer)
+    logger.info(
+      `Processed transfer ${transfer.id}. Status: ${result.txStatus} TxHash: ${result.txHash}`
+    )
+  }
+
+  clearWatchdog()
```
franckc: What is the purpose of the watchdog specifically? Is it for:
A. Preventing more than 1 instance of the process to get started
B. Preventing the in-process cron scheduler to somehow start more than 1 job

If it's for A. maybe the watchdog should get initialized when the process starts and deleted at shutdown.

BTW, not sure if we have decided under which environment we will run the server.
 - If it's kubernetes, I'm not sure exactly about the detail of the lifecycle of a container. Does kube bring a new container and check its health before terminating the old one? If yes that could potentially cause 2 cron jobs to run in parallel.
 - Not sure about heroku
Overall I wonder if it would be simpler to use the DB as a central persistent storage for the watchdog. Like a new "watchdog" table.

infra/token-transfer-server/src/lib/transfer.js
tomlinton: Yea

infra/token-transfer-server/src/tasks/transfer.js
tomlinton: For B. More than 1 instance of the process is actually fine, but we really don't want more than one instance of the transfer processing job.

Re: kubernetes, it depends on the resource type. A StatefulSet will terminate one before starting the other, but a standard deployment will run into the problem you described.

Nothing decided yet, but a DB is probably more robust. You raise good points. I'm going to firm this up more later I think, but I want to test this out of on dev/staging before making any decisions.

infra/token-transfer-server/src/lib/transfer.js
tomlinton: 👍

infra/token-transfer-server/src/tasks/transfer.js
tomlinton: Will remove this, will add it back in later, it will be part of the analytics stuff

infra/token-transfer-server/src/tasks/transfer.js
franckc: Sgtm!

infra/token-transfer-server/src/lib/transfer.js
tomlinton: Done!

infra/token-transfer-server/src/tasks/transfer.js
tomlinton: I moved the watchdog stuff into a separate SQL table which records task start and end times and links the transfers via a foreign key. Would you mind giving it a look? Then will merge this :tada:

franckc: sure... will look a bit later tonite...

infra/token-transfer-server/src/tasks/transfer.js
```diff
@@ -0,0 +1,82 @@
...
+      end: null
+    }
+  })
+  if (outstandingTasks.length > 0) {
+    throw new Error(
+      `Found incomplete transfer task(s), wait for completion or clean up manually.`
+    )
+  }
+
+  const transferTask = await TransferTask.create({
```
franckc: should this row insertion get moved to line 47, after the test on no unconfirmed transfer?

infra/token-transfer-server/src/tasks/transfer.js
```diff
@@ -0,0 +1,82 @@
...
+} = require('../config')
+const { Transfer, TransferTask, Sequelize } = require('../models')
+const { executeTransfer } = require('../lib/transfer')
+const logger = require('../logger')
+const enums = require('../enums')
+
+const executeTransfers = async () => {
+  logger.info('Running execute transfers job...')
+
+  const outstandingTasks = await TransferTask.findAll({
```
franckc: If we really want to be paranoid about race conditions, we should hold a lock on the TransferTask table, do the required validation, insert a new row into TransferTable and the finally release the lock... What are the chances that with Heroku 2 instances of the server get started at the same time?

franckc: Thinking a bit more about all this, it might be simpler to take a lock on each Transfer row before starting its processing. This way, no need for a TransferTask table at all. And it may be less prone to situations where there is TransferTask row in a dangling state because the app got terminated (for a code push or other operational reason) in the middle of a task being executed...

Something like this:
```
 - get all Transfer rows ready for processing
 - for each row, do:
     - In a transaction statement:
         - Lock the row (using SELECT FOR UPDATE)
         - Make sure the row is in status Enqueue, if not error out
         - Update row status to Processing
         - Commit transaction
     - Execute the transfer
     - Update row status to Processed (or Failed)
````

tomlinton: Good idea on the table wide lock. That'd only happen if we scaled up the dynos, but can never be too careful!

The row level locking was the other way I thought about doing this. I think the disadvantages are:

1. Doesn't actually prevent two transfer tasks running side by side and processing different transactions at the same time and causing nonce problems.
2. This extra table records how long it takes to run the job and what transactions ran with which job which may be useful.

I didn't fully flesh out the idea though so might be worth revisiting. Any thoughts on 1?

infra/token-transfer-server/src/tasks/transfer.js
tomlinton: Yep my bad 👍

infra/token-transfer-server/src/tasks/transfer.js
franckc: Oh indeed, excellent point about the nonce... Yeah I think that's a big disadvantage of this solution. Darn blockchain always making things more complicated ;)
A TransferTask table is then probably the best way to go. Can't think of anything really better...

We'll have to be a bit careful about not deploying while there are payments being processed.
One potential way to help with that would be to have a way to indicate to the cron scheduler to not schedule any new task so that we can deploy new code. Maybe scheduler could check a value in the DB that indicates if payments are enabled or not before it executes any transfer. Maybe like a "config" table where we can store a flag. And that we would set manually via a DB query before a deployment. But perhaps not worth over optimizing. Let's see if we run into that problem before implementing something like this.

PR #4072, opened 2019-12-02
Adds PartnerReferral to december rules

mikeshultz:
### Description:

Adds PartnerReferral to december so it can be properly tested.

### Checklist:


infra/growth/campaigns/december.js
```diff
@@ -512,6 +512,22 @@ const novemberConfig = {
...
+          config: {
+            eventType: 'PartnerReferral',
+            reward: {
+              amount: '0',
+              currency: 'OGN'
+            },
+            limit: 1,
+            visible: true,
+            nextLevelCondition: false,
+            scope: 'campaign',
```
franckc: should the scope be set to 'user' so that this reward is only available once for a given user?
otherwise by setting it to 'campaign', the user could potentially get the reward every campaign.

mikeshultz: I copied the other referral rule, I think.  Had no idea what that was used for.  Will change.

PR #3992, opened 2019-11-15
Bug Fixes

shahthepro:
[Link to test build](https://originprotocol.github.io/test-builds/shah)

- [x] Fix #3568 
- [x] Fix #3749 
- [x] Fix #3351 
- [x] Fix #3958 
- [x] Potential fix for #3970 

### Checklist:

- [ ] Test your work and double-check to confirm that you didn't break anything
- [ ] [Wrap any new text/strings with `fbt`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#wrapping-text) for translation
- [ ] If there are any new/changed strings to translate, [`npm run translate`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#integrating-translations)
- [ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/origin/tree/master/docs)

PR #4078, opened 2019-12-03
Fix messaging encryption key

shahthepro:
[This changeset](https://github.com/OriginProtocol/origin/pull/3992/commits/5e0becc3df15a068f515483ca9abe2b2ead3354b) from me broke messaging encryption keys. This PR should fix that


shahthepro:
FYI Merging this and deploying it now. Let me know if you have any comments on this.

PR #4079, opened 2019-12-03
Shop updates

nick:


dapps/shop/backend/routes/stripe.js
```diff
@@ -0,0 +1,96 @@
...
+const Web3 = require('web3')
+const bodyParser = require('body-parser')
+const stripe = require('stripe')(process.env.STRIPE_BACKEND)
+
+const abi = require('../utils/_abi')
+
+const ZeroAddress = '0x0000000000000000000000000000000000000000'
+
+const web3 = new Web3(config.provider)
+const account = web3.eth.accounts.wallet.add(process.env.WEB3_PK)
```
mikeshultz: if we intend for this to be run by users in the futuretime, maybe we should move to Ethereum SecretStore files and use env vars maybe for decryption password.

PR #4113, opened 2019-12-11
[Growth signup] Do not show scary popup to US IPs

franckc:
### Description:

For US IPs, as opposed to a scary modal asking the user to confirm they are not US citizens, we show a checkbox on the terms acceptance page, asking the user to confirm they are not US citizen.

### Checklist:


dapps/marketplace/src/pages/growth/WithEnrolmentModal.js
```diff
@@ -124,8 +126,26 @@ function withEnrolmentModal(WrappedComponent) {
...
+    }
+
     handleAcceptTermsCheck(e) {
-      this.setState({ termsAccepted: e.target.checked })
+      const termsAccepted = e.target.checked
+      this.setState({ termsAccepted })
+      const { notCitizenChecked } = this.state
+      // If we detected country as US based on IP, user must certify they
+      // are not US citizen/resident before we enable the signup button.
+      if (this.countryCode === 'US') {
```
sparrowDom: we have this [locationInfo](https://github.com/OriginProtocol/origin/blob/master/infra/growth/src/util/locationInfo.js#L3-L4) in the growth service where we "restrict" countries - which currently only holds US country. To avoid having it configured in 2 places this check could be `if (eligiblity === 'Restricted')`. Eligibility info is fetched here: 
https://github.com/OriginProtocol/origin/blob/master/dapps/marketplace/src/pages/growth/WithEnrolmentModal.js#L461

franckc: Thanks for the review. Yes good point. It is not ideal. But I think Matt only wanted this for the US and not for any other future restricted countries. I added a comment to clarify this in the code...

PR #4123, opened 2019-12-12
Reward should come from campaigns.json

mikeshultz:
### Description:

Pull reward amount from `campaigns.json` for partner funnel dapp pages as the current method from PR #4115 will always return 0.

This is untested and depends on https://github.com/OriginProtocol/origin-website/pull/458

### Checklist:

### DIFF
--- a/dapps/marketplace/src/hoc/withPartnerCampaignConfig.js
+++ b/dapps/marketplace/src/hoc/withPartnerCampaignConfig.js
@@ -4,8 +4,12 @@ import { PARNTER_REFERRAL_CONFIG_URL } from 'constants/config'
 
 let cachedCampaigns = {}
 
-async function fetchConfig() {
-  const resp = await fetch(PARNTER_REFERRAL_CONFIG_URL)
+async function fetchConfig(signal) {
+  if (Object.keys(cachedCampaigns).length) {
+    return cachedCampaigns
+  }
+
+  const resp = await fetch(PARNTER_REFERRAL_CONFIG_URL, { signal })
   if (resp.status !== 200) {
     console.error('Failed to fetch campaign config')
     return
@@ -19,18 +23,15 @@ function withPartnerCampaignConfig(WrappedComponent) {
     const [partnerConfig, setPartnerConfig] = useState(cachedCampaigns)
 
     useEffect(() => {
-      let timeout
-      if (cachedCampaigns && Object.keys(cachedCampaigns).length) {
-        return
-      } else {
-        timeout = setTimeout(async () => {
-          setPartnerConfig(await fetchConfig())
-        }, 3000)
-      }
+      const abortController = new AbortController()
 
-      return () => {
-        clearTimeout(timeout)
-      }
+      fetchConfig(abortController.signal)
+        .then(data => setPartnerConfig(data))
+        .catch(() => {
+          // Probably aborted, do nothing
+        })
+
+      return () => abortController.abort()
     }, [])
 
     return (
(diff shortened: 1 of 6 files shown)


dapps/marketplace/src/hoc/withPartnerCampaignConfig.js
```diff
@@ -0,0 +1,42 @@
...
+      let timeout
+      if (cachedCampaigns) {
+        return
+      } else {
+        setTimeout(async () => {
+          const resp = await fetch(PARNTER_REFERRAL_CONFIG_URL)
+          if (resp !== 200) {
+            console.error('Failed to fetch campaign config')
+            return
+          }
```
shahthepro: @mikeshultz Just a thought. We seem to already have the code to fetch the referral campaigns in mobile and growth packages. Instead of having the similar code across packages, perhaps we can extract this into a separate GraphQL query and use that in all 3 (mobile, growth, DApp).

shahthepro: Similar fetch in mobile: https://github.com/OriginProtocol/origin/blob/master/infra/growth/src/resources/rules/partnerReferralRule.js#L82

And in growth: https://github.com/OriginProtocol/origin/blob/master/mobile/src/screens/partnerWelcome.js#L62

mikeshultz: Adding a whole graphql resolver and fetch just to pull config JSON from originprotocol.com really irks me for some reason.  That said, I'm having some issues with `useEffect()` working properly so I'm amenable...

shahthepro: Yeah, using fetch inside useEffect doesn't always go well (I learnt it the hard way with Split). 

For the DApp alone, you could move that to a GraphQL resolver (local schema, not the remote one) and handle the memory cache (the 60-minutes cache you do) in the resolver or set `fetchPolicy` to `cache-only` (since it is not bound to change often). In the HOC, you could use `useQuery` hook. Just my opinion though.

shahthepro: Or you could just create a React Context for fetching that. Both should work fine.

mikeshultz: eh, I got the hook to work okay, I'm not going to refactor everything to integrate into graphql for the sake of speed.  Feel free to do that after if you feel strongly about it.

mikeshultz:
@shahthepro The enrolled check always seems to be `false` in the `Finished` component.  Know what that's about?  Is this check necessary here for some reason?

    const enrolled = growthEnrollmentStatus === 'Enrolled'

shahthepro:
@mikeshultz We don't want to show the users that they earned "x OGN" when they clicked on "Skip" in the previous step. That's the reason that check is there.

mikeshultz:
https://github.com/OriginProtocol/origin-website/pull/458 was merged and deployed

dapps/marketplace/src/utils/growthTools.js
```diff
@@ -245,7 +257,17 @@ export function getReferralReward(campaign) {
       (referralCode.startsWith('op') ? 'PartnerReferral' : 'Referral')
   )
 
-  const reward = get(action, 'reward.amount')
+  // if there's no campaign action, nor config for the referral code, there's
+  // no point in continuing
+  if (
+    !action ||
+    !Object.prototype.hasOwnProperty.call(campaignConfig, codeParts[1]) ||
+    campaignConfig[codeParts[1]].reward.currency !== 'ogn'
```
shahthepro: @mikeshultz This might throw if `campaignConfig[codeParts[1]]` is undefined

PR #4150, opened 2019-12-17
Tomlinton/t3 withdrawal testing

tomlinton:
Fixes stuff I ran into testing withdrawals on Origin testnet.

- Minor UI tweaks
- Move block confirmation checking to separate function so the transfer execution doesn't need to wait for it (to avoid problems where the server is restarted while waiting for confirmation)
- Added some missing columns to TransferTask model via a new migration (created_at and updated_at never got created)
- Updated @origin/token to use the correct RPC address for testnet
- Updated @origin/token to have a txIsConfirmed method that contains the guts of waitForTxConfirmation but is useful on its own

### DIFF
--- a/infra/token-transfer-server/src/lib/transfer.js
+++ b/infra/token-transfer-server/src/lib/transfer.js
@@ -1,5 +1,6 @@
 const get = require('lodash.get')
 const jwt = require('jsonwebtoken')
+const moment = require('moment')
 
 const Token = require('@origin/token/src/token')
 
@@ -12,7 +13,7 @@ const {
   TRANSFER_REQUEST,
   TRANSFER_CONFIRMED
 } = require('../constants/events')
-const { Event, Transfer, sequelize } = require('../models')
+const { Event, Transfer, User, sequelize } = require('../models')
 const { hasBalance } = require('./balance')
 const { transferConfirmationTimeout, transferHasExpired } = require('../shared')
 const { clientUrl, encryptionSecret, networkId } = require('../config')
@@ -20,7 +21,7 @@ const enums = require('../enums')
 const logger = require('../logger')
 
 // Number of block confirmations required for a transfer to be consider completed.
-const NumBlockConfirmation = 8
+const NumBlockConfirmation = 3
 
 // Wait up to 20 min for a transaction to get confirmed
 const ConfirmationTimeoutSec = 20 * 60 * 60
@@ -170,9 +171,9 @@ async function confirmTransfer(transfer, user) {
 }
 
 /**
- * Sends a blockchain transaction to transfer tokens and waits for the transaction to get confirmed.
- * @param {Transfer} transfer: DB model Transfer object
- * @returns {Promise<{txHash: string, txStatus: string}>}
+ * Sends a blockchain transaction to transfer tokens.
+ * @param {Transfer} transfer: Db model transfer object
+ * @returns {Promise<String>} Hash of the transaction
  */
 async function executeTransfer(transfer, transferTaskId) {
   const user = await hasBalance(transfer.userId, transfer.amount, transfer)
@@ -184,6 +185,7 @@ async function executeTransfer(transfer, transferTaskId) {
 
   // Setup token library
   const token = new Token(networkId)
+
   // Send transaction to transfer the tokens and record txHash in the DB.
   const naturalAmount = token.toNaturalUnit(transfer.amount)
   const supplier = await token.defaultAccount()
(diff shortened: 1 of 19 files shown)


infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -190,47 +195,81 @@ async function executeTransfer(transfer, transferTaskId) {
...
     status: enums.TransferStatuses.WaitingConfirmation,
     fromAddress: supplier.toLowerCase(),
     txHash
   })
 
+  return txHash
+}
+
+/**
+ * Sends a blockchain transaction to transfer tokens.
```
franckc: nit: update comment

infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -190,47 +195,81 @@ async function executeTransfer(transfer, transferTaskId) {
...
+    if (timedOut) {
       transferStatus = enums.TransferStatuses.Failed
       eventAction = TRANSFER_FAILED
       failureReason = 'Confirmation timeout'
-      break
-    default:
-      throw new Error(`Unexpected status ${status} for txHash ${txHash}`)
+      logger.warn(`Transaction timed out ${transfer.txHash}`)
+    }
+    return null
```
franckc: Should there be a else statement here so that null is only return if it is not a timeout?

infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -190,47 +195,81 @@ async function executeTransfer(transfer, transferTaskId) {
...
-    case 'timeout':
+  if (!result) {
+    // Check if too long elapsed
+    const timedOut =
+      moment.utc(transfer.updatedAt).diff(moment.utc(), 'seconds') >
+      ConfirmationTimeoutSec
+    if (timedOut) {
       transferStatus = enums.TransferStatuses.Failed
       eventAction = TRANSFER_FAILED
       failureReason = 'Confirmation timeout'
```
franckc: In case of a confirmation timeout, the tx may still get executed later on.
This may happen for ex if there is a surge in gas price right after the tx was submitted, causing the tx to not get mined for a while... If we consider the tx as failed the user may be able to double spend...
I wonder if it would be safer to halt all transfers until we get a chance to manually look at what happened?

infra/token-transfer-server/src/tasks/transfer.js
```diff
@@ -27,43 +40,39 @@ const executeTransfers = async () => {
...
         { transaction: txn }
       )
     }
   )
 
+  if (!transferTask) return
+
   const cutoffTime = moment.utc().subtract(largeTransferDelayMinutes, 'minutes')
-  const transfers = await Transfer.findAll({
+  const transfer = await Transfer.findOne({
```
franckc: should we add an order by created_at asc so that we process in fifo order?

packages/token/src/token.js
```diff
@@ -141,6 +141,48 @@ class Token {
     return txHash
   }
 
+  /**
+   * Checks if a transaction is confirmed by checking if a number of blocks has
+   * been mined since the transaction was added.
+   * @param {string} txHash: the transaction hash.
+   * @param {number} numBlocks: the number of block confirmation to wait for
+   * @returns {Promise<{status:string, receipt:Object}>}
```
franckc: nit: update to reflect it returns undefined in case of an error
```
@returns {Promise<{status:string, receipt:Object} || undefined>}
```

infra/token-transfer-server/src/tasks/transfer.js
tomlinton: 👍

packages/token/src/token.js
tomlinton: 👍

infra/token-transfer-server/src/lib/transfer.js
tomlinton: What about we just remove the timeout? In the case you described it'd be good if the script would eventually pick it up, even if it took a long time. And if there is a transaction that is still waiting for confirmation nothing else will process anyway so its effectively the same as halting all transfers.

infra/token-transfer-server/src/lib/transfer.js
tomlinton: My bad!

infra/token-transfer-server/src/lib/transfer.js
franckc: Yes I like your idea of removing the timeout. I'd vote for that too.

tomlinton:
I don't think we gain much from doing that, except a warning in the error logs that there were multiple? I might be missing something. 😀

franckc:
I think the thing we gain is that in case there would be more than 1 transfer waiting for confirmation, with the current logic, we may send a new tx without waiting for all the pending transfers to get confirmed first. Agreed it's paranoia. Feel free to ignore... :)

tomlinton:
Ahh I see the advantage now, sounds like a good idea, will do!

PR #4213, opened 2020-01-07
Change Network Displayed Name

micahalcorn:
Lots of people are changing this to `Origin` for some reason. This is an untested, blind change that I hope isn't misguided. 🤞


tomlinton:
That would work fine @micahalcorn, but I'll tweak it so the Origin network is only available in dev builds.

PR #4219, opened 2020-01-08
Notify users to go through Telegram Attestation on DApp

shahthepro:
When an user sends `/start` directly to the bot, reply to them with a message to go through the flow on the DApp again.


shahthepro:
@micahalcorn Have updated the message.

PR #4223, opened 2020-01-09
Increases ingress rate limits for graphql and growth

mikeshultz:
### Description:

Ref: #4215

### Checklist:


franckc:
LGTM

I forgot why we had the limit at 50 for growth. It should only get called to fetch campaign data so I can't really imagine a user making more than 1 qps. Anyway, I think it's fine to increase. If we get DDOS we can always reduce...

Thanks for looking into this @mikeshultz !

mikeshultz:
Don't we `POST` a bunch of user events to growth?

mikeshultz:
This stuff will get rolled during the next deployments.

PR #4244, opened 2020-01-10
Get time for login from auth server

shahthepro:
User are not able to login if they have incorrect date/time on their machine. This PR adds an endpoint to auth server that will return the current timestamp of auth server.

It shouldn't be a problem security wise either, we already trust the client to send the timestamp. Also, we don't generate auth token if the time is in future or older than a few minutes.

This requires a mobile app update.

### Checklist:

### DIFF
--- a/packages/auth-client/src/auth-client.js
+++ b/packages/auth-client/src/auth-client.js
@@ -97,7 +97,7 @@ class AuthClient {
 
     const payload = {
       message: message || 'I intend to sign in to Origin Marketplace',
-      timestamp: Date.now()
+      timestamp: await this.getServerTime()
     }
 
     let signature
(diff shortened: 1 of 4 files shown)


packages/auth-client/src/auth-client.js
```diff
@@ -380,6 +380,28 @@ class AuthClient {
...
+    try {
+      let url = new URL(this.authServer)
+      url.pathname = '/api/now'
+      url = url.toString()
+
+      const rawResponse = await fetch(url)
+
+      const response = await rawResponse.json()
+
+      return response.body.timestamp || Date.now()
```
sparrowDom: If body contents of the request can be retrieved but timestamp field is missing, I would rather see this throwing an exception and at least logging the error, and then returning client side Date.now(). Instead of silently failing

shahthepro: Thanks @sparrowDom. Have made that change

Issue #4259, opened 2020-01-13
Campaigns home page crash when there is no active campaign


franckc:
We just announced we'll be ending the January campaign sooner than anticipated (tomorrow).
That means that until the new February campaign starts, there won't be any active campaign.

I updated staging to end the January campaign and the [rewards home page](https://dapp.staging.originprotocol.com/#/campaigns) crashes. The console has the following error:
``` TypeError: Cannot read property 'actions' of undefined```

When there is no active campaign we should show a message along the lines of "There is currently no active campaign, please come back later and check".

franckc:
@sparrowDom Would you mind taking a look at this when you wake up tomorrow?

@micahalcorn Please provide feedback regarding what the UI should say when there is no active campaign.

sparrowDom:
On it 👍

sparrowDom:
Fix for this here: https://github.com/OriginProtocol/origin/pull/4263

PR #4263, opened 2020-01-13
Sparrow dom/rewards active campaign

sparrowDom:
- display a message when there is not active rewards campaign
- fix docker environment setup

Fixes: https://github.com/OriginProtocol/origin/issues/4259

### Checklist:


docker-compose.yml
```diff
@@ -210,7 +210,7 @@ services:
     command:
       >
       /bin/bash -c "wait-for.sh -t 0 -q elasticsearch:9200 --
-      wait-for.sh -t 0 -q services:1111 --
+      wait-for.sh -t 0 -q services:8545 --
```
sparrowDom: @mikeshultz have changed this back from port 1111 to 8454. It was preventing some of the services in docker to start up.

sparrowDom:
@micahalcorn @franckc currently I have it set to this:
Screenshot 2020-01-13 15 31 53

Happy to change it if needed.

micahalcorn:
Thanks @sparrowDom 👍 Should we link to the blog post there?

dapps/marketplace/src/pages/growth/Campaigns.js
```diff
@@ -338,14 +338,32 @@ class GrowthCampaign extends Component {
...
+            <div className="mt-3">
+              Budget allocated for the last campaign has been reached earlier
+              than expected.
+            </div>
+            <div className="mt-3">
+              Check our{' '}
+              <a href="https://discordapp.com/invite/jyxpUSe">
+                #announcements Discord
+              </a>{' '}
+              channel for information about future campaigns.
```
shahthepro: @sparrowDom Shouldn't these messages be translated?

sparrowDom: yup thanks fixed

sparrowDom:
@micahalcorn was thinking about linking to the blog post, but we will show this screen anytime where there is no active campaigns. The block post link might be outdated with the next occurrence when we close a campaign early.

docker-compose.yml
mikeshultz: Perhaps because something with services isn't working?  This is preventing other services from coming up because services (probably) isn't coming up.  It specifically brings up an empty service on port `1111` to indicate that everything is good to go:

https://github.com/OriginProtocol/origin/blob/2334d4e15a8026671445a9cf1587428ce284ae7d/packages/services/index.js#L461

By checking 8545(ganache), you're skipping basically every other service that it brings up.  I'd recommend looking into the cause of your issue instead of making this change.

franckc:
Chatted with @joshfraser - he recommends we put a link to the blogpost and to our telegram english channel. We'll have to update the blog post link in the future... but this is still worth doing to in order to minimize the amount of users inquiries in the short-term.

franckc:
@micahalcorn Could you please check the copies before I merge this in? Thanks:)

docker-compose.yml
sparrowDom: Ah thanks @mikeshultz The problem was with identity contract not deploying. Fix here: 
https://github.com/OriginProtocol/origin/pull/4274

PR #4264, opened 2020-01-13
Don't include username in attestation body if empty

### LINKED ISSUE #4217: Invalid Attestion(s)
We're getting quite a few reports of this with Telegram. I haven't been able to find a way to reproduce it.

invalid

-----

shahthepro:
In Telegram the username is optional. So when an user without an username tries to do the attestation, the attestation body contains `{ username: { raw: null } }` which is invalid as per the attestation_v1.0.0.json schema. It has to be string value or not exist at all.

This fixes #4217 

### Checklist:


shahthepro:
@franckc We fallback to `id` field if `username` is not set in other places. And `id` will always be present.

franckc:
cool. sounds good!

PR #4276, opened 2020-01-14
Update fallback value in bridge tests

shahthepro:
PR #4248 had changed the fallback values in [exchange-rate.js](https://github.com/OriginProtocol/origin/blob/shahthepro/fix-exchange-rate-tests/infra/bridge/src/utils/exchange-rate.js#L18) file.

This PR updates that values so that the tests don't fail


sparrowDom:
Screenshot 2020-01-14 17 28 32

was also fixing this today  :)

yours is better though undoing mine

Issue #4250, opened 2020-01-10
Self referrals should not be allowed


franckc:
In the prod database, I'm seeing ~300 entries in the growth_referral table where referrer and referee are identical.

This query shows the accounts:
select * from growth_referral where referrer_eth_address = referee_eth_address;

I'm not sure how this could happen since the only place we write to growth_referral is in the method makeReferralConnection and it does have [a check](https://github.com/OriginProtocol/origin/blob/master/infra/growth/src/resources/invite.js#L81) to prevent exactly this from happening!

franckc:
@sparrowDom I have been staring at this code for 10 minutes and I don't see how this could be happening... any idea? :)

sparrowDom:
This is a weird one. After debugging I've found 2 issues with referral codes: 
- we store the referral code in [navigations package](https://github.com/OriginProtocol/origin/blob/288423f3d582fd0746096988799bbe4c13114f17/infra/notifications/src/app.js#L282) and there are no checks there. This must be the reason for the bug
- after debugging in local docker environment, the whole `enrol` endpoint in growth  server is never called. Since the authentication server does the signing-in part. And from my testing the referral link on desktop is never made. 

Preparing a fix for these 2 issues

franckc:
oh... yes I totally missed that referrals are also inserted from the notification server. that must be it.
excellent catch sir!

sparrowDom:
Fix here: https://github.com/OriginProtocol/origin/pull/4277

sparrowDom:
fixed via https://github.com/OriginProtocol/origin/pull/4277

PR #4277, opened 2020-01-14
Referral code refactor/fixes

sparrowDom:
### Description:

We weren't doing much checks for the referral code / link generation on the notification server side. This PR separates that into a utils package that `growth` and `notifications` both use. 

Also fixes issue where inviteCode generation can be 10 characters long instead of 11.  On production database there are 6.8k of 110k cases like that. 

Fixes issue: https://github.com/OriginProtocol/origin/issues/4250
### Checklist:


packages/utils/src/growth.js
```diff
@@ -0,0 +1,65 @@
...
+  if (!inviteCode) {
+    throw new Error(`Invalid invite code ${code}`)
+  }
+  return inviteCode.ethAddress
+}
+
+async function makeReferralConnection(code, walletAddress) {
+  const referralLink = await GrowthReferral.findOne({
+    where: {
+      refereeEthAddress: walletAddress.toLowerCase()
```
franckc: nit: since walletAddress.toLowerCase() is used several times, maybe declare a variable at the top of the method. like
```const referee = walletAddress.toLowerCase()```

**Comment by franckc** [comment]
Ah also are there any unit tests that need to be updated?

**Comment by sparrowDom** [comment]
@franckc renamed `growth-event` to `growth-shared`. Thanks for the suggestion, I think it is the most elegant solution. 

Have also found a bug where invite code creation can create a code 10 characters longs instead of 11. Desktop welcome page ignores those invite codes. (While mobile has no such restrictions). There are 6.8k invite codes like that out of 110k total on production.

To fix that this brach:
- fixes the invite code bug, to always generate 11 characters long codes
- removed the desktop welcome page restrictions of invite code length. -> to cause as little as possible problems on production with existing invite codes 

Do you think this is an ok solution? Also PTAL

**Comment by sparrowDom** [comment]
hmm also @franckc does the rename require any changes for production deployment scripts maybe?

File: infra/growth/src/resources/authentication.js [hunk 2]
```diff
@@ -201,10 +201,11 @@ async function createInviteCode(accountId) {
    * IMPORTANT: Frontend expects 11 letter invite code. Be mindful of that
    * if changing invite code creation logic
    */
-  const code =
+  const code = (
     `${accountId.substring(2, 5)}` +
     `${accountId.substring(accountId.length - 3, accountId.length)}` +
     `${Math.round(Math.random() * 1000000).toString(16)}`
+  ).padEnd('1')
```
franckc: Where does the Front-End check for 11 characters invite code?

The logic above may generate codes that are anywhere between 7 and 11 characters, depending on the value returned by Math.random().

For example, if Math.random() return 0.000001, Math.round(Math.random() * 1000000).toString(16) returns '1' and the code will be 7 chars long.

The prod database reflects this:
```
import=# select length(code), count(*) from growth_invite_code group by length(code);
 length | count  
--------+--------
      7 |      1
      8 |     26
      9 |    413
     10 |   6818
     11 | 103391
```

So I don't think we need to add the padEnd

infra/growth-shared/src/resources/referral.js
```diff
@@ -0,0 +1,66 @@
...
+    throw new Error(`Referrer ${referrer} can't use own referral code`)
+  }
+
+  const returnObj = {
+    referrer
+  }
+
+  if (referralLink) {
+    if (
+      referralLink.referrerEthAddress.toLowerCase() !== referrer.toLowerCase()
```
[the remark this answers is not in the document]
sparrowDom: Good point... removing

infra/growth/src/resources/authentication.js
sparrowDom: On master we still have this check, have removed it in this branch. See Welcome.js page changes in this PR. And yes agreed, with changes in Welcome page this change is no longer needed. Removing it....

PR #4290, opened 2020-01-17
[Growth] Mass payout contract and scripts

franckc:
### Description:

This PR adds a new TokenDistributor contract that allows to send tokens to multiple recipients in a single transaction. The caller must first approve the contract to send tokens, then call the transfer method and pass a list of addresses and amounts. Only callers that have been whitelisted can call the transfer method. The contract is Ownable so that after deployment it can be transferred to a multi-sig wallet in production.

The token package is modified to add two new methods (approveMulti and creditMulti) for integrating with the TokenDistributor contract.

The growth distribution script is modified to support a new command line argument for processing payout in batches. When set, the script will break up the payouts in chunks and process each chunk by calling the TokenDistributor contract to transfer tokens.

I did some preliminary testing on Rinkeby and things seem to work as expected. I will conduct further testing on Mainnet before using this for the January payout. But I wanted to share this PR early so that folks can poke any hole at the contract and overall logic surrounding it...

### Checklist:


nick:
All looks good to me. Seems like there's a lint test error in the contracts repo... nothing I could see apart from that 👍

PR #4294, opened 2020-01-20
Default supressErrors to true on growth HOCs

nick:
Default to failing silently if growth server is unavailable. Since growth server seems to sometimes be blocked, prevents a hard error in the dapp.


sparrowDom:
@nick perhaps only test what happens on Welcome page, if growth is not available and errors are suppressed.  Might need to turn suppression off theere

PR #4306, opened 2020-01-28
Fix USDT payment

### LINKED ISSUE #4304: USDT incorrectly configured
Example GraphQL query result for a USDT balance:

```
{"__typename": "TokenHolder", "balance": "18755256", "id": "USDT_0xd15f105E2667Fd52313877585D4a0b316C7D8a20", "token": {"__typename": "Token", "decimals": 18, "id": "0xdac17f958d2ee523a2206206994597c13d831ec7"}}
```

USDT should be 6 decimals and not 18. It is most likely the settings in the network files in `packages/graphql/src/configs/`.

shahthepro:
Should fix #4304

[Link to test build](https://originprotocol.github.io/test-builds/shah/)

### Checklist:


tomlinton:
Thanks @shahthepro, looks like a lot of test failures that are probably related to these changes. I'll update the branch to run them again but would you mind checking them out if it still fails?

shahthepro:
@tomlinton Thanks. PR is not yet ready for review. Yeah, will be fixing the tests

tomlinton:
Sorry @shahthepro, someone (cough @micahalcorn) requested review 👺 😂

packages/graphql/src/utils/unitTokenValue.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+const getUnitWithDecimals = memoize(decimals => {
+  const web3 = contracts.web3
+
+  const unit = Object.keys(web3.utils.unitMap).find(unit => {
+    const valueInWei = web3.utils.unitMap[unit]
+
+    const zeroCount = valueInWei.length - valueInWei.replace(/0/g, '').length
+
+    return zeroCount === decimals
```
sparrowDom: AFAICT web3 units for each zero count [don't exist](https://web3js.readthedocs.io/en/v1.2.0/web3-utils.html#id88). Sometimes there are 3 zeros between the 2 closest units. So the above function might miss the correct unit in case its' decimal count is between 2 units. (for example between `szabo` -> 12 zeros and `nano` -> 9 zeros)

Maybe one suggestion as how to amend the above implementation: 
```
let biggestUnit = 'wei'
let c = 0
const unit = Object.keys(web3.utils.unitMap).forEach(unit => {
    const valueInWei = web3.utils.unitMap[unit]

    const zeroCount = valueInWei.length - valueInWei.replace(/0/g, '').length
    if (zeroCount <= decimals && zeroCount > biggestUnitZeros) {
        biggestUnit = unit
        biggestUnitZeros = zeroCount
    }
```

mikeshultz:
Sorry, it's early and I may have missed something but I don't get why we're doing string operations here instead of using BN.js?  Here's a conversion from "human units" to an 8-decimal token value:

    const ONE_TOKEN = new BN('100000000') // 1e8
    function toReal(v) {
      return v.mul(ONE_TOKEN)
    }

Seems a bit more accurate and straight forward than trying to do math this way?

tomlinton:
Ethersjs has great utility functions for this kind of thing. Maybe we should add it as a dependency and slowly try to move away from web3js. Tough job though 😞

packages/graphql/src/utils/unitTokenValue.js
shahthepro: Thanks for catching this @sparrowDom, Will fix this

shahthepro:
@mikeshultz 
> Seems a bit more accurate and straight forward than trying to do math this way?
Agreed. This seems simple. Will update it. Thanks

@franckc 
> Agreed with @sparrowDom that adding some simple unit tests would be nice.
Thanks. Will do it

@tomlinton 
> Ethersjs has great utility functions for this kind of thing. Maybe we should add it as a dependency and slowly try to move away from web3js. Tough job though 😞
We could but Mike's solution should reduce the complexity of the code right now. I'm not sure if we should add a dependency right now. But thanks, let me check this

shahthepro:
@mikeshultz 
I tried that but it doesn't work well. `new web3.utils.BN('123.12345').toString()` results in `122812345` which is way off.

However, I did some other change that will simply it more (more like treating it as a string).

@sparrowDom @franckc The issue with decimal places (that are not multiples of 3) is fixed now and have also added a couple of simple unit tests

@tomlinton Tried ether.js, but the potential that Domen reported was still reproducible (i.e the function not working with decimals which are not multiples of 3). I managed to fix this without ether.js or even using web3. So skipping adding ether.js for now

mikeshultz:
@shahthepro you can't feed BN a decimal.  In fact, I'm surprised that didn't error.  Here's a slightly more comprehensive example using your given numbers starting with a string:

    > x = new BN(parseInt(parseFloat('123.12345') * 1e6)) // for 6-decimal, use 1e8 for 8-decimal, etc
    <BN: 756b6fa>
    > x.toString()
    '123123450'

shahthepro:
> @shahthepro you can't feed BN a decimal. In fact, I'm surprised that didn't error. Here's a slightly more comprehensive example using your given numbers starting with a string:

> EDIT: This also assumes relatively small numbers as user input that won't overflow in JS. Might want to add some checks if you expect extremely large numbers(`N > Number.MAX_SAFE_INTEGER / 1e8`).

Yep, this is what I'm concerned about. We had code to truncate anything longer than 18 decimals (before passing that to `toWei`). So, there is a possibility, albeit rare, that it can fail. Also, even with 18 decimals, JS seems to round off digits.

I'm probably gonna merge this change in and will fix forward, if we ever any issue with the present implementation

Thanks @mikeshultz

PR #4312, opened 2020-01-29
Catch clicks exactly on icon in dropdowns

wanderingstan:
The ::after pseudo-element with carrot icon for dropdowns (in footer at least) was capturing mouse events. Meaning that if you clicked *exactly* on the carrot, the dropdown would not open. Adding `pointer-events: none` prevents this capturing of mouse events.

Image showing region of un-clickable pseudo-element in dapp footer dropdown:

image


franckc:
Cool! Danke shoën herr @wanderingstan

PR #4337, opened 2020-02-13
[Growth] Add an endpoint for logging events into the growth_event table

franckc:
### Description:
This new endpoint '/log_event' on growth.originprotocol.com can be called by the Chrome Extension to register an install so that we can then give credit to the user as part of Origin Rewards.

Can be tested with:
curl -X POST "http://localhost:4008/log_event?type=BrowserExtensionInstalled&eth_address=0xD85A569F3C26f81070544451131C742283360400&uid=123"

### Checklist:


infra/growth/src/resources/eventLogger.js
```diff
@@ -0,0 +1,56 @@
...
+  // Check presence and validity of the uid parameter
+  // TODO(franck): check validity of the uid.
+  const uid = req.query.uid
+  if (!uid) {
+    return res.status(400).end('Missing uid')
+  }
+
+  const now = new Date()
+  await GrowthEvent.insert(
+    logger,
```
tomlinton: These fields look like they might be in the wrong order or am I going crazy? Why is `logger` being passed here?

franckc: Ah yeah, this stuff is confusing. This is actually not calling the DB model for GrowthEvent but rather a wrapper method defined here: https://github.com/OriginProtocol/origin/blob/master/infra/growth-shared/src/resources/event.js#L35

infra/growth/src/resources/eventLogger.js
```diff
@@ -0,0 +1,56 @@
...
+ *  - type: "BrowserExtensionInstalled"
+ *  - eth_address: eth address the user used when signing up for origin rewwards.
+ *  - uid: unique browser extension id
+ * @param req
+ * @param res
+ * @returns {Promise<*>}
+ */
+async function logEvent(req, res) {
+  // Check presence and validity of the type parameter.
+  const type = req.query.type
```
shahthepro: @franckc Are we going to blindly trust the client? There isn't much we can do but I'm wondering if there is something we can do to make sure that the API call isn't called by the user manually to fool the system?

franckc: There won't be any login required after installing the browser extension. So we can't authenticate the user in the backend.

We could perhaps come up with some sort of encoding scheme that uses the eth address to generate a token that then gets verified by the backend. For ex: token = hash(secret+eth address), with secret being hardcoded in the browser extension code (in some obfuscated way) and also being known of the server. It's not bullet proof since a  savvy user could extract the secret from the extension source code.

Now, the payout for the reward not only rely on the presence of this event. It will also be contingent on having recorded browsing activity via the extension throughout the campaign.

So not sure this verification token is necessary. I'd prefer keeping things simple?

franckc: CC @tomlinton in case he has any thoughts on this topic...

franckc:
I'm going to merge this in as is so that I can keep making progress. But please do let me know any feedback and I'll address it as a separate PR. Thanks! :)

infra/growth/src/resources/eventLogger.js
tomlinton: Yea I'm not worried about it since they need browsing activity anyway

infra/growth/src/resources/eventLogger.js
```diff
@@ -0,0 +1,56 @@
...
+  if (!ethAddress) {
+    return res.status(400).end('Missing eth_address')
+  }
+  if (!Web3.utils.isAddress(ethAddress)) {
+    return res.status(400).end('Invalid eth_address')
+  }
+
+  // Check presence and validity of the uid parameter
+  // TODO(franck): check validity of the uid.
+  const uid = req.query.uid
```
tomlinton: Just realised this uses query strings instead of the POST body, is that a growth standard?

franckc: Not at all. If easier for the client to pass args in the body, I'm happy to change this endpoint. Lmk...

tomlinton: That'd be great, it might be easier. If I pass everything in the query string express errors with a body required error so body is probably mandatory.

FYI I've submitted an initial build of the extension with this growth call with the parameters in the body.

franckc: No problem. I'll make this change first thing on Monday...

PR #4361, opened 2020-02-21
Alternative to polling for balance updates

mikeshultz:
### Description:

This uses "strategic" events to update the balance instead of an interval to poll for balance updates.  Currently this includes sending transactions and changing networks.

Anything I'm missing?

### Checklist:

### DIFF
--- a/mobile/src/components/origin-web3view.js
+++ b/mobile/src/components/origin-web3view.js
@@ -6,7 +6,13 @@
 
 import React, { useState } from 'react'
 import { connect } from 'react-redux'
-import { Modal, StyleSheet, PermissionsAndroid, Platform } from 'react-native'
+import {
+  Modal,
+  StyleSheet,
+  PermissionsAndroid,
+  Platform,
+  DeviceEventEmitter
+} from 'react-native'
 import { ethers } from 'ethers'
 import { WebView } from 'react-native-webview'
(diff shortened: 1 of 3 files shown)


tomlinton:
Is there any reason to ditch polling? As far as I can tell this doesn't cover the case where someone is staring at the Origin wallet and tries to send OGN to it from MetaMask, it'll never update without some intervention.

mikeshultz:
The reason is a reduction in JSON-RPC requests. Having it infinitely ask for balance updates is expensive(especially ERC20 balance checks).

That's an interesting use case.  Wonder if a refresh button(or pull down) is reasonable addition for this.  Or maybe just add polling when the wallet page is displayed.  Though I don't know if the latter is possible since the nav seems to not unmount pages after their initial display.

Suggestions?

tomlinton:
This should work for polling on focus only https://reactnavigation.org/docs/en/3.x/with-navigation-focus.html

mikeshultz:
Awesome, that seems to do the trick.  Updated so it'll poll every 10s when a user is eyeballing their balances.  

I'm kind of wondering if this is the only place a balance update should occur(when a user is looking at the wallet screen).  Though it'll take 10s for the first interval to hit so I think I'll leave the other event emitters.

PR #4367, opened 2020-02-25
Create ogn script for a Docker quality of life buff

rembrandtreyes:
Created a script that would "shorten" a lot of the docker commands we use daily e.g. `docker-compose up` becomes `ogn up` or `docker exec -it notifications /bin/bash` becomes `ogn shell notifications` etc...

Added [lazydocker](https://github.com/jesseduffield/lazydocker) to be even lazier with docker 😏. So running lazydocker in a terminal will bring up a cool GUI to manage and maintain docker. Have to incorporate conditionals to properly install for windows

I created a nonexhaustive list of ogn commands to be used.
```
Usage: ogn [command]
    Commands:
    - help                         Show this message.
    - init                         Perform initial container setup, generate alias and install lazydocker (Run once).
    - (up | start)                 Spin up containers and follow logs.
        - (up | start) detach      Optional detach to not follow docker logs.
    - build                        Rebuild the Origin Services container.
    - list                         Display a list of container names
    - log <container name>         Follow a single container logs e.g. ogn log services
    - logs                         Follow ALL container log outputs.
    - stop                         Stop ALL running containers.
    - restart <container name>     Restart a single container.
        - restart all              Restart ALL containers and follow logs
        - restart all detach       Restart ALL containers and detach logs
    - shell <container name>       Start a shell for a specific container e.g. ogn shell notifications
    - prune                        Removes dangling docker images, containers, volumes, networks, and build cache.
```

### Description:

### Checklist:

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -157,6 +157,50 @@ consuming. To install new dependencies, get a shell in the container and run
 ⚠️ Don't run `docker-compose down` when stopping containers! Any changes made
 since the initial Docker build will be lost. Instead use `docker-compose stop`.
 
+#### Would you like a docker-compose shortcut?
+In case you need to go fast, we have created a script that will convert some of the
+`docker-compose` commands to `ogn [command]`. Note: you still can use 
+`docker-compose` and all other docker commands. This is just an added layer to your 
+docker user-experience here at Origin Protocol. 
+
+ 
+🏎️ Follow these steps to get setup today:
+1. Make sure you have [homebrew](https://brew.sh/) installed
+2. `cd` to your origin directory
+3. Run `./ogn init` - This will create an alias in your shell file and installs [lazydocker](https://github.com/jesseduffield/lazydocker).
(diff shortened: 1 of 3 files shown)


franckc:
Cool! Thanks for taking the initiative and submitting this daft PR for feedback:)

Now, I have to confess that I haven't been using our docker setup so I don't have a strong opinion on this.

CCing @shahthepro @sparrowDom @mikeshultz @tomlinton since they might have some opinion/feedback

sparrowDom:
I haven't tested it but looks cool and friendlier for the user. 
If this doesn't interfere with the current way docker is operated (so it can be used both ways) and is another layer with easier UX I am all for it.

shahthepro:
@rembrandtreyes This is cool. But I agree with Domen that users should still be able to Docker, if they need/want to. I don't see anything that will break anything when using docker though.

Regarding the README, I'd prefer to have both docker related commands as well as this CLI tool's commands.

rembrandtreyes:
Thanks for the feedback and yes this will be an added benefit to the development experience. Any developer/user will still be able to use the native Docker commands.

I'll adjust the README so that this is an optional step for the docker environment.

PR #4368, opened 2020-02-25
Nick/shop backend

nick:
Various dshop backend changes


mikeshultz:
Not sure I follow the whole network ID relationship you seem to want here.  I'm wondering if you have some endgame in mind and want to be on the same page.

- Why add an extra call to get the network ID from the provdier?
- Why store it in the DB, when the DB is explicitly unique to the network?
- Why parameterize it if the backend instance is specific to a network?

Do you envision a future with a multi-network unified backend or something?

nick:
Yea the original idea was that the backend can handle multiple networks at the same time. The DB being explicitly unique to the network was only for SQLite, which was only ever intended for dev mode anyway.

I moved removed the NETWORK_ID env var as it can be inferred directly from the provider.

PR #4416, opened 2020-04-01
Early lockups for T3 server

tomlinton:
- Implements a /config endpoint so we don't have to set so many vars on both the server and the client. The client queries the server and uses those config variables.
- Some refactoring of the way variables are stored to support the config endpoint
- Implements early lockups against the next vest. This requires the logic to understand that there are two "balances", one standard balance and the balance of the next vest. Early lockups store the vest that they were for in the data column of the lockup table.

@franckc this is ready for you to start taking a look at if you have some cycles. 🕶


tomlinton:
I should mention this is server only, the changes for the client will be in a separate PR to keep everything manageable. The server is the important bit though.

franckc:
ack. I will review this beauty tomorrow morning first thing...

infra/token-transfer-server/src/shared.js
```diff
@@ -91,6 +105,45 @@ function calculateLocked(lockups) {
   }, BigNumber(0))
 }
 
+/* Determine if a lockup is a nearly lockup based
```
franckc, line 108: nearly a typo?

infra/token-transfer-server/src/lib/lockup.js
```diff
@@ -19,14 +25,50 @@ const { encryptionSecret, clientUrl } = require('../config')
...
+
+    // Augment lockup data field with additional information about the future
+    // vesting event that this lockup is being created for
+    const user = await User.findOne({
+      where: {
+        id: userId
+      },
+      include: [{ model: Grant }]
+    })
+    data = {
```
franckc, line 52: I wonder if we should also store an "early" boolean flag in the data, to make it easy to query/look at early lockups.

infra/token-transfer-server/src/shared.js
```diff
@@ -91,6 +105,45 @@ function calculateLocked(lockups) {
...
+ */
+const isEarlyLockup = lockup => {
+  return !!(lockup.data && lockup.data.vest && lockup.data.vest.grantId)
```
franckc, line 111: per my earlier comment, it might be simpler to test on lockup.data.early

infra/token-transfer-server/src/lib/balance.js
```diff
@@ -44,39 +39,80 @@ async function hasBalance(userId, amount, currentTransferId = null) {
...
+    `User ${user.email} tokens in early lockup`,
+    nextVestLocked.toString()
+  )
+
   // Calculate total available tokens
   const available = vested
     .plus(lockupEarnings)
     .minus(transferWithdrawnAmount)
     .minus(lockedAmount)
+    .plus(nextVestLocked)
```
franckc, line 64: should this be .plus(earlyLockupEarnings) ?

infra/token-transfer-server/src/config.js
```diff
@@ -53,10 +58,41 @@ const unlockDate = moment(process.env.UNLOCK_DATE, 'YYYY-MM-DD').isValid()
...
+  'YYYY-MM-DD'
+).isValid()
+  ? moment.utc(earlyLockupsEnabledUntil) > moment.utc()
+  : false
+
+// Lockup duration in months
+const lockupDuration = process.env.LOCKUP_DURATION || 12
+
+// Whether lockups are enabled
+const lockupsEnabled = process.env.LOCKUPS_ENABLED || true
```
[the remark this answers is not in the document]
tomlinton: I do that here https://github.com/OriginProtocol/origin/blob/55c8f5803e3d9708a7a826a558cab712c3b5b008/infra/token-transfer-server/src/utils.js#L27

The getter methods in that utils file are used throughout

infra/token-transfer-server/src/lib/lockup.js
```diff
@@ -19,14 +25,50 @@ const { encryptionSecret, clientUrl } = require('../config')
...
  * @param userId - user id of the user adding the lockup
  * @param amount - the amount to be locked
+ * @param early - whether this is an early lockup of the next vest
+ * @param data - additional data to be stored with the lockup, e.g. fingerprint
  * @returns {Promise<Lockup>} Lockup object.
  */
-async function addLockup(userId, amount, data = {}) {
-  const user = await hasBalance(userId, amount)
+async function addLockup(userId, amount, early, data = {}) {
+  // Check if user has sufficient balance
```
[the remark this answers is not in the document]
tomlinton: There is a lock in the controller so this can't get called concurrently, did you mean in addition to that? Not sure if you saw that

Ref https://github.com/OriginProtocol/origin/blob/55c8f5803e3d9708a7a826a558cab712c3b5b008/infra/token-transfer-server/src/controllers/lockup.js#L84

infra/token-transfer-server/src/lib/lockup.js
```diff
@@ -46,15 +88,15 @@ async function addLockup(userId, amount, data = {}) {
   const txn = await sequelize.transaction()
   try {
     lockup = await Lockup.create({
-      userId: user.id,
-      start: moment.utc(),
+      userId: userId,
+      start: moment.utc(), // Note lockup starts immediately at not at time of next vest
```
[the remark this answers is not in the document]
tomlinton: 👍

infra/token-transfer-server/src/lib/lockup.js
```diff
@@ -19,14 +25,50 @@ const { encryptionSecret, clientUrl } = require('../config')
...
+    const user = await User.findOne({
+      where: {
+        id: userId
+      },
+      include: [{ model: Grant }]
+    })
+    data = {
+      ...data,
+      vest: getNextVest(
+        user.Grants.map(g => g.get({ plain: true })),
```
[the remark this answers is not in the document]
tomlinton: Yea it is, the vestingSchedule function it calls requires plain objects or it doesn't return the vesting amounts correctly. Typescript was something I should have done 😦

infra/token-transfer-server/src/shared.js
tomlinton: 👍

infra/token-transfer-server/src/lib/balance.js
tomlinton: When an early lockup is matured (i.e. current date is passed the date stored in "end" column) it doesn't get treated differently to a non-early lockup, so the lockupEarnings includes earnings from early lockups

Ref https://github.com/OriginProtocol/origin/blob/55c8f5803e3d9708a7a826a558cab712c3b5b008/infra/token-transfer-server/src/shared.js#L62-L73

infra/token-transfer-server/src/lib/balance.js
```diff
@@ -1,28 +1,23 @@
...
- * or a lockup.
- *
- * Throws an exception in case the request is invalid.
+ * Helper method to check the available balance for a user.
  *
  * @param userId
- * @param amount
- * @param currentTransferId
- * @returns Promise<User>
+ * @returns Promise<Integer> - available balance
```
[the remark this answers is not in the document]
tomlinton: 👍

infra/token-transfer-server/src/lib/transfer.js
```diff
@@ -175,7 +184,14 @@ async function confirmTransfer(transfer, user) {
  * @returns {Promise<String>} Hash of the transaction
  */
 async function executeTransfer(transfer, transferTaskId, token) {
-  const user = await hasBalance(transfer.userId, transfer.amount, transfer.id)
+  const balance = await getBalance(transfer.userId)
+  // Subtract the current transfer amount from the available balance because
+  // it is what we are transferring
+  if (BigNumber(transfer.amount).gt(balance.minus(transfer.amount))) {
```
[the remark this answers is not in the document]
tomlinton: Ergh good spot, thanks!

infra/token-transfer-server/src/lib/balance.js
franckc: Ah thanks. that makes sense re calculateUnlockedEarnings.

Now sorry to be super slow, but why do we need to add nextVestLocked to the available balance? aren't the nextVestLocked tokens locked and by definition not available?

infra/token-transfer-server/src/lib/lockup.js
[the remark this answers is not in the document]
franckc: Ah sorry I had missed that... yes, AyncLock achieves the same thing as locking for update the user row in the DB, as long as we are certain there will never be more than 1 instance of the T3 server.

infra/token-transfer-server/src/lib/lockup.js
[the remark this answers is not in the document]
franckc: got it.

infra/token-transfer-server/src/lib/balance.js
tomlinton: That is because `lockedAmount` is subtracted, and it includes early lockups as well. Early lockups are against a future vest and don't change the current balance, so that number needs to be added back in.

I'm not very happy with the naming, I'll see if I can perhaps separate `lockedAmount` into early and standard to make it clearer.

infra/token-transfer-server/src/utils.js
```diff
@@ -16,13 +24,18 @@ const getUnlockDate = () => {
 }
 
 // Use a function so that this value can be mocked in tests
-const getEarnOgnEnabled = () => {
-  return earnOgnEnabled
+const getLockupsEnabled = () => {
+  return lockupsEnabled === true || lockupsEnabled == 'true'
```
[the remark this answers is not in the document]
tomlinton: 👍

infra/token-transfer-server/src/lib/lockup.js
tomlinton: I'll probably leave this for now, I don't think we'll be doing manual queries and its a pretty small optimisation, this is actually less code than adding the migration, thanks for suggesting it though

infra/token-transfer-server/src/lib/balance.js
franckc: Ah I get it now. Thanks for explaining... Maybe add a comment in the code for dummies like me.

tomlinton: No it is way too confusing, I've changed it to be clearer. It doesn't include it in the lockedAmount anymore and I'll handle it separately.

tomlinton:
Think I've resolved everything and I cleaned up some of the jsdoc stuff. I'll merge this if it passes.

franckc:
Sounds good! Great work on this sir.

PR #4419, opened 2020-04-02
Add support for targetting language/country

shahthepro:
Adds methods to send push notification based country/language on codes and adds an endpoint to return all supported language and country codes.

### DIFF
--- a/infra/notifications/src/app.js
+++ b/infra/notifications/src/app.js
@@ -16,7 +16,8 @@ const _ = require('lodash')
 // const { browserPush } = require('./browserPush')
 const MobilePush = require('./mobilePush')
 const EmailSender = require('./emailSend')
-const MobileRegistry = require('./models').MobileRegistry
+const db = require('./models')
+const MobileRegistry = db.MobileRegistry
 const { GrowthEventTypes } = require('@origin/growth-shared/src/enums')
 const { GrowthEvent } = require('@origin/growth-shared/src/resources/event')
 const {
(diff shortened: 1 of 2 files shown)


infra/notifications/src/app.js
```diff
@@ -471,6 +472,35 @@ app.post('/send_pn', pushAppAuth, async (req, res) => {
   res.send(data)
 })
 
+const cachedCountryCodes = []
+const cachcedLangCodes = []
+app.get('/push-targets', pushAppAuth, async (req, res) => {
+  if (!cachcedLangCodes.length) {
+    const [langCodes] = await db.sequelize.query(`
+      select distinct data->>'language' as langcode
```
franckc, line 480: That's going to be a pretty slow query. An alternative would be to query the identity table on the 'country' column and add an index to it.

infra/notifications/src/app.js
```diff
@@ -471,6 +472,35 @@ app.post('/send_pn', pushAppAuth, async (req, res) => {
   res.send(data)
 })
 
+const cachedCountryCodes = []
+const cachcedLangCodes = []
+app.get('/push-targets', pushAppAuth, async (req, res) => {
+  if (!cachcedLangCodes.length) {
```
franckc, line 478: typo: cachedLangCodes

infra/notifications/src/app.js
```diff
@@ -471,6 +472,35 @@ app.post('/send_pn', pushAppAuth, async (req, res) => {
...
+    const [langCodes] = await db.sequelize.query(`
+      select distinct data->>'language' as langcode 
+        from growth_participant order by data->>'language' ASC;
+    `)
+
+    cachcedLangCodes.push(...langCodes.map(l => l.langcode).filter(l => l))
+  }
+
+  if (!cachedCountryCodes.length) {
+    const [countryCodes] = await db.sequelize.query(`
```
franckc, line 488: Pretty slow query. If performance becomes an issue, consider adding a DB index on the country column.

infra/notifications/src/mobilePush.js
```diff
@@ -555,10 +455,86 @@ class MobilePush {
     await this._sendInChunks(data, deviceType, mobileRegisters)
   }
 
+  async _sendToCountryCode(data, deviceType, countryCode) {
```
franckc, line 458: No need to fix now but maybe add a comment that if the server would get restarted in the middle of this method being called (for ex. due to a deploy), some notifications would not get sent.

infra/notifications/src/mobilePush.js
```diff
@@ -555,10 +455,86 @@ class MobilePush {
...
+    `,
+      {
+        replacements: {
+          countryCode
+        }
+      }
+    )
+
+    if (!addresses.length) {
+      console.log('There are no addresses from', countryCode)
```
franckc, line 475: nit: use logger

shahthepro:
@franckc Thanks for the comments. yes regarding the queries, I did see the slowness on prod.  So, I just exported the values from there to static files: https://github.com/oplabs/pn-admin/tree/master/api/data

infra/notifications/src/mobilePush.js
shahthepro: Will add that.

infra/notifications/src/mobilePush.js
```diff
@@ -555,10 +455,86 @@ class MobilePush {
...
+        deleted: false,
+        'permissions.alert': true
+      },
+      order: [['updatedAt', 'DESC']] // Most recently updated records first.
+    })
+
+    await this._sendInChunks(data, deviceType, mobileRegisters)
+  }
+
   async multicastMessage(data) {
```
franckc, line 532: nit: in order to know what's going on on the server side, I would add a bit more logging throughout the code. at least something to track the progress of the notifications being sent. at least something like "Sending chunk X out of Y"

PR #4432, opened 2020-04-08
Execute transfer logic fixes

tomlinton:
- Fix T3 balance check when executing transfers
- Fix issue when LARGE_TRANSFER_DELAY_MINUTES set to "0" via environment variable

### DIFF
--- a/infra/token-transfer-server/src/config.js
+++ b/infra/token-transfer-server/src/config.js
@@ -45,8 +45,7 @@ if (!sessionSecret) {
 const largeTransferThreshold =
   Number(process.env.LARGE_TRANSFER_THRESHOLD) || 100000
 
-const largeTransferDelayMinutes =
-  Number(process.env.LARGE_TRANSFER_DELAY_MINUTES) || 60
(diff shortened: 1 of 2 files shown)


infra/token-transfer-server/src/config.js
```diff
@@ -45,8 +45,7 @@ if (!sessionSecret) {
 const largeTransferThreshold =
   Number(process.env.LARGE_TRANSFER_THRESHOLD) || 100000
 
-const largeTransferDelayMinutes =
-  Number(process.env.LARGE_TRANSFER_DELAY_MINUTES) || 60
+const largeTransferDelayMinutes = process.env.LARGE_TRANSFER_DELAY_MINUTES || 60
```
franckc, line 48: Why removing the cat to Number?

tomlinton: If you set the environment variable to 0 it appears as a string here, then it does `Number("0") || 60` which is always 60 because it is doing 0 || 60. The value gets passed to moment.subtract which is happy with a number or a string anyway.

franckc: sg. ty

PR #4436, opened 2020-04-10
IPFS headers

nick:
Move IPFS security headers to nginx.config


mikeshultz:
Where did `more_set_headers` come from?  We sure that's going to work here?  Only reference I can find is this, which appears to require install: https://github.com/openresty/headers-more-nginx-module

PR #4439, opened 2020-04-13
Fix Push Notifications

shahthepro:
- Fixes push notifications on Android
- Fixes notification URL opening on iOS and Android
- Opens any external URL (from notifications) in native browser on both iOS and Android.


mobile/android/build.gradle
```diff
@@ -34,14 +34,15 @@ buildscript {
...
+        buildToolsVersion = "28.0.3"
     }
     repositories {
         google()
         jcenter()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.5.3'
-        classpath 'com.google.gms:google-services:4.0.2'
+        classpath 'com.google.gms:google-services:+'
```
mikeshultz, line 45: Is this okay?  I remember all these java libs requiring very specific versions for compat between eachother.

shahthepro: I'm not sure if anything changed recently, but the version that we had previously didn't work. So I had to change this.

shahthepro:
@mikeshultz yes that `+` does work on Android

PR #4453, opened 2020-05-05
Dshop: adds /affiliate/products endpoint

mikeshultz:
### Description:

Adds the `/afiliate/products` endpoint.

This expects the BQ dataset to be updated to use a fully qualified listing ID.  There's some other things with the BQ data that leave me wanting but this will do what we need.

### Checklist:


dapps/shop/backend/routes/affiliate.js
```diff
@@ -24,6 +27,48 @@ function authAffiliate(req, res, next) {
   next()
 }
 
+/**
+ * Formats a BigQuery product row into a dshop product obj
+ */
+function bqProductFormatter(product) {
+  // product ID is only in the IPFS path returned
```
franckc, line 34: Can't we use the new product_id col from BQ?

mikeshultz: No.  

Nick/FE is expecting the product ID that has already been used.  Here's an example I was given to emulate: https://affiliate.ogn.app/affiliate/products.json  

Though I do realize now that I forgot to prepend listing ID to it, which I'll fix shortly.

dapps/shop/backend/routes/affiliate.js
```diff
@@ -24,6 +27,68 @@ function authAffiliate(req, res, next) {
...
+  const id = makeId(product)
+  const {
+    price,
+    title,
+    image
+  } = product
+
+  return {
+    id,
+    data: `/ips/${product.ipfs_path}`,
```
nick, line 56: should this be /ipfs/?

dapps/shop/backend/routes/affiliate.js
```diff
@@ -24,6 +27,68 @@ function authAffiliate(req, res, next) {
...
+    price,
+    title,
+    image
+  }
+}
+
+/**
+ * Get the products from BigQuery for a specific shop
+ */
+async function fetchAffiliateProducts(listingId) {
```
nick, line 66: I think we were going to have the 50 most recent products from any store, instead of a specific store? Maybe we can add that as an option?

dapps/shop/backend/routes/affiliate.js
mikeshultz: yes, good catch

dapps/shop/backend/routes/affiliate.js
mikeshultz: Did not know it was not going to be shop specific.  Perhaps only limit it to a specific shop if an auth token is provided?

mikeshultz: Also, most recent 50?  I don't really have any idea about the use case of this endpoint, but that seems... unexpected.  It'll almost always be from one or two stores, and recency doesn't mean much when it comes to affiliates?

mikeshultz: Made filtering by shop optional.  If no auth token is provided, it won't filter on listing ID.

mikeshultz: limited to 50.

PR #4456, opened 2020-05-11
DShop admin: add a button to the console to emit a ListingUpdated event 

franckc:
### Description:

Add a button to the DShop admin > Settings > Console panel that allows to update the listing associated with a shop by calling the marketplace contract.

The listing data that is published includes a field "shopIpfsHash" that points to the IPFS hash of the dshop.

This feature enables any external system to listen for the ListingUpdated events to get notified when a dshop gets updated. For example, an ETL pipeline could listen to those events and trigger a re-indexing of the associated dshop's catalog.

### Checklist:

### DIFF
--- a/dapps/shop/src/pages/admin/settings/Console.js
+++ b/dapps/shop/src/pages/admin/settings/Console.js
@@ -1,151 +1,11 @@
 import React, { useState } from 'react'
-import ethers from 'ethers'
 
-import { get, post, getIpfsHashFromBytes32 } from '@origin/ipfs'
+import { get } from '@origin/ipfs'
 
 import useConfig from 'utils/useConfig'
 import useShopConfig from 'utils/useShopConfig'
+import { updateListing } from 'utils/listing'
 
-const abi = [
-  'function updateListing(uint256, bytes32, uint256)',
-  'event ListingCreated (address indexed party, uint indexed listingID, bytes32 ipfsHash)',
-  'event ListingUpdated (address indexed party, uint indexed listingID, bytes32 ipfsHash)'
-]
-
-/**
- * Gets a marketplace listing's most recent IPFS hash by querying the ethereum network
- * for relevant logs.
- * Note: In the future we may want to consider storing the listing's most recent
- * hash in the shop back-end DB.
- *
- * @param {Object} provider: ether.js provider.
- * @param {Object} contract: ether.js contract object for the marketplace.
- * @param {Integer} contractEpoch: block number at which contract was created.
- * @param {string} seller: ethereum address of the seller
- * @param {Integer} listingId: id of the listing in the marketplace contract.
- * @returns {Promise<string>} IPFS hash
- * @private
- */
-async function _getListingLatestIpfsHash(
-  provider,
-  contract,
-  contractEpoch,
-  seller,
-  listingId
-) {
-  // Create a filter to get the ListingCreated events for the listing.
-  const listingCreatedFilter = contract.filters.ListingCreated(
-    seller,
-    listingId,
-    null
-  )
-  listingCreatedFilter.fromBlock = contractEpoch || 0
-
-  // Create a filter to get the ListingUpdated events for the listing.
-  const listingUpdatedFilter = contract.filters.ListingCreated(
-    seller,
-    listingId,
-    null
-  )
-  listingUpdatedFilter.fromBlock = contractEpoch || 0
-
-  // Get all the ListingCreated|Updated events and keep only the most recent.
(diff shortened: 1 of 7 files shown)


dapps/shop/src/pages/admin/settings/Console.js
```diff
@@ -1,15 +1,159 @@
...
+ *
+ * @param {Object} provider: ether.js provider.
+ * @param {Object} contract: ether.js contract object for the marketplace.
+ * @param {Integer} contractEpoch: block number at which contract was created.
+ * @param {string} seller: ethereum address of the seller
+ * @param {Integer} listingId: id of the listing in the marketplace contract.
+ * @returns {Promise<string>} IPFS hash
+ * @private
+ */
+async function _getListingLatestIpfsHash(
```
nick, line 29: Do we need to bother with knowing the old hash? Can we just overwrite it with fresh listing data, regardless of what it was before?

franckc: Yes, we could. I was worried that in the future we may store more data in the listing's JSON and that it would get lost if we overwrite it.

But in the short-term we could overwrite with the generic base listing data. I'm happy to put a comment in there so that we remember in the future and to simplify it for the short-term.
Lmk if you have any preference...

mikeshultz: I like the idea of leaving the schema open and only impacting what this admin cares about.  Though functionally we probably don't need to, it's nice to have.

nick: Yea I think we can keep it simple in the short term so we're not requesting logs 😊

dapps/shop/src/pages/admin/settings/Console.js
```diff
@@ -1,15 +1,159 @@
...
+    return
+  }
+  if (shopIpfsHash.length !== 46) {
+    console.error('Invalid IPFS hash')
+    return
+  }
+
+  const enabled = await window.ethereum.enable()
+  if (!enabled) {
+    console.error('No web3 provider detected in the browser')
```
mikeshultz, line 91: I don't think this `enable()` API is universal.  web3.js doesn't support it and it's not in the Ethereum "standard", for instance.  Looks like it's probably part of EIP-2255, which is still draft, whatever that's worth.

dapps/shop/src/pages/admin/settings/Console.js
```diff
@@ -132,6 +276,32 @@ const AdminConsole = () => {
...
+      <form
+        className="d-flex"
+        onSubmit={async e => {
+          e.preventDefault()
+          if (!shopIpfsHash) {
+            return
+          }
+          console.log('Calling ListingUpdated...')
+          updateListing(config, shopIpfsHash)
+            .then(() => console.log('Listing updated successfully'))
```
mikeshultz, line 290: Should this maybe have more user feedback?

dapps/shop/src/pages/admin/settings/Console.js
franckc: I was just blatantly copying the same pattern as existing functionality for creating a listing. From this [code](https://github.com/OriginProtocol/origin/blob/master/dapps/shop/src/pages/super-admin/shops/CreateListing.js#L32)

Do you know what's the "standard" or better way to do this?

dapps/shop/src/pages/admin/settings/Console.js
franckc: Yes, all the functionalities exposed in the console are pretty bare bone and don't provide any UI feedback. Probably ok for the demo but in the future we if the business team starts using those, we should add some feedback.

For my information, do we have some toast library for showing a success/error message in the UI that we already use in the shop admin? I did not see any...

dapps/shop/src/pages/admin/settings/Console.js
franckc: Ok. I'll leave the _getListingLatestIpfsHash method as a util in the code but for the short-term I won't call it. We can wire it later if we decide to...

dapps/shop/src/pages/admin/settings/Console.js
mikeshultz: I don't know what the current defacto standard would be no.  I'm out of practice on multi-client web3 support.

franckc:
- Did a small refactor to move all listing utilities to utils/listing.js
- Incorporated review feedback to overwrite listing data in the short-term.

Ready for another round of review...

franckc:
I'm going to land this. Happy to iterate in a subsequent PR to address any further feedback...

PR #4461, opened 2020-05-13
CloudDNS support for dshop super admin

mikeshultz:
### Description:

Adds CloudDNS support for the super-admin

Ref: https://github.com/OriginProtocol/dshop/issues/20

### Checklist:


dapps/shop/backend/utils/dns/clouddns.js
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Looks for a record within a set of records matching name
+ *
+ * Ref: https://googleapis.dev/nodejs/dns/latest/Record.html
+ *
+ * @param {Array} of Record
+ * @param {string} name of DNS record
+ * @returns {boolean} if a record matching name was found
+ */
+function containsRecord(records, name) {
```
franckc, line 46: I think calls to this method can be replaced with ```records.includes(name)```.

dapps/shop/backend/utils/dns/clouddns.js
```diff
@@ -0,0 +1,177 @@
...
+  for (let i=0; i<records.length; i++) {
+    if (records[i].name == name) return true
+  }
+  return false
+}
+
+/**
+ * Get a speific Zone
```
franckc, line 54: nit: typo

dapps/shop/backend/utils/dns/clouddns.js
```diff
@@ -0,0 +1,177 @@
...
+ * @param {string} name of DNS record
+ * @returns {Zone}
+ */
+async function getZone(DNSName) {
+  const dns = getClient()
+  const [zones] = await dns.getZones({ maxResults: 50 })
+
+  for (let i=0; i<zones.length; i++) {
+    const dnsName = zones[i].metadata.dnsName
+    if (dnsName == DNSName) {
```
franckc, line 67: nit: as a convention, don't we use === ?

dapps/shop/backend/utils/dns/clouddns.js
```diff
@@ -0,0 +1,177 @@
...
+  getClient(credentials)
+
+  const zoneObj = await getZone(zone)
+
+  if (!zoneObj || !(await zoneObj.exists())) {
+    console.error(`Zone ${zone} not found.`)
+    return
+  }
+
+  const records = zoneObj.getRecords({ maxResults: 250 })
```
franckc, line 159: just checking. getRecords is not async and therefore does not need to use await when calling it?

dapps/shop/backend/utils/dns/clouddns.js
```diff
@@ -0,0 +1,177 @@
...
+
+  if (await containsRecord(records, fqSubdomain)) {
```
franckc, line 161: no need for await here. containsRecord is not asyn. also per prev comment, we can replace with ```if records.includes(fqSubdomain)```

packages/services/package.json
```diff
@@ -21,7 +21,7 @@
     "commander": "^4.0.0",
     "ganache-core": "2.4.0",
     "http-proxy": "1.18.0",
-    "ipfs": "^0.40.0",
+    "ipfs": "^0.43.3",
```
franckc, line 24: Not sure we should check this in yet since it breaks the DApp unit tests :(

dapps/shop/backend/utils/dns/clouddns.js
mikeshultz: How's that?  `records` is an array of objects of which we need to check against the `name` property.

dapps/shop/backend/utils/dns/clouddns.js
mikeshultz: Guess I got used to Python, thanks

dapps/shop/backend/utils/dns/clouddns.js
mikeshultz: Absolutely async.  Never would've noticed this in testing.  Thanks for catching that.

packages/services/package.json
mikeshultz: reverted

mikeshultz:
Thanks for the thorough review as always, @franckc

dapps/shop/backend/utils/dns/clouddns.js
franckc: doh. yes I read the code too fast!
I guess you could use the "some" function if you want ?
```records.some(r => r.name === name)```

Comment (mikeshultz) [reply dapps/shop/backend/utils/dns/clouddns.js:46]: probably no gain in refactoring at this point but will keep it in mind for the future

File: dapps/shop/src/pages/super-admin/Settings.js [hunk 7]
```diff
@@ -53,7 +54,7 @@ const SuperAdminSettings = () => {
     }).then(async res => {
       if (res.ok) {
         const json = await res.json()
-        setState(json)
+        setState(json.config)
```
nick, line 57: Hrm not sure this is necessary any more?  I made a recent change that lumps it all into one object

mikeshultz: Missed that.  Haven't tested with the recent master changes.  Let me do that before merge.

dapps/shop/backend/utils/dns/clouddns.js
mikeshultz: had to touch this code anyway, because of the getRecords result fix.  Refactored.

dapps/shop/src/pages/super-admin/Settings.js
mikeshultz: Reverted this.  Looks to be working.

PR #4465, opened 2020-05-16
Queue for dshop backend processes

DanielVF:
When a user makes a Stripe payment, we **now enqueue the work of creating an offer and submitting it to the blockchain**.

This PR uses the [bull queue](https://github.com/OptimalBits/bull) package, which is backed by redis. It seems to have a good API for this, doesn’t try to do too much, and has a lot of built in error handling.

It is great to be able to run our stack without external dependencies, so I’ve written a fallback queue class that is used if redis is not configured. In this case, we skip the queue, and instead process queue items as soon as they are added. 

I’ve added [bull-board](https://github.com/vcapretz/bull-board) UI to /super-admin/queue

In the future, we'll want to also queue up the actions that run on the listener side. I wanted to get this PR live first to verify our approach, and get our education before adding those. Those should be very quick.

- Part of issue https://github.com/OriginProtocol/dshop/issues/16


dapps/shop/backend/queues/makeOfferProcessor.js
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * job.data should have {shopId, amount, encryptedData}
+ * @param {*} job
+ */
+async function processor(job) {
+  job.log('Started processing')
+  const { shopId, amount, encryptedData } = job.data
+  const shop = await getShop(shopId)
+  job.log('Load encrypted shop config')
+  const shopConfig = await getShopConfig(shop)
```
nick, line 30: I don't think this needs to be async

dapps/shop/backend/queues/makeOfferProcessor.js
```diff
@@ -0,0 +1,127 @@
...
+}
+
+function createOfferJson(lid, amount, encryptedData) {
+  return {
+    schemaId: 'https://schema.originprotocol.com/offer_2.0.0.json',
+    listingId: lid.toString(),
+    listingType: 'unit',
+    unitsPurchased: 1,
+    totalPrice: {
+      amount: amount / 100,
```
nick, line 92: I'm thinking we set this to 0 for now so as not to expose too much data

DanielVF: I'm wondering if there's a clearer way to say here that the data is actually elsewhere, not just zero:

Perhaps:

    totalPrice: {
      amount: 0,
      currency: 'encrypted'
    }

dapps/shop/backend/queues/makeOfferProcessor.js
DanielVF: Both getting the shop from the db, and the shop config over a web request are async.

nick: There's nothing async inside the `getShopConfig` function though... do it doesn't need to be async from what I can see

dapps/shop/backend/queues/makeOfferProcessor.js
nick: Sure can do that. Or we could just omit `totalPrice` altogether.... though that might break things

dapps/shop/backend/package.json
```diff
@@ -22,6 +22,8 @@
     "body-parser": "^1.19.0",
     "bottleneck": "^2.19.5",
     "bs58": "^4.0.1",
+    "bull": "^3.14.0",
```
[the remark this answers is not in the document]
DanielVF: Whoops, fixed.

dapps/shop/backend/app.js
```diff
@@ -76,6 +78,7 @@ require('./routes/printful')(app)
 require('./routes/stripe')(app)
 require('./routes/discounts')(app)
 require('./routes/tx')(app)
+require('./queues/ui.js')(app)
```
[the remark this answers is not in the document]
DanielVF: Yup. Fixed.

dapps/shop/backend/queues/queues.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @file Queue objects for adding processing dshop jobs.
+ *
+ * If no REDIS_URL defined, will fallback to prossing jobs as the jobs are
+ * submitted, without using a queue.
+ */
+
+const { REDIS_URL } = require('../utils/const')
+
+const Queue = REDIS_URL ? require('bull') : require('./fallbackQueue')
```
nick, line 10: needs to be bullmq now we changed the package?

DanielVF: Fixed. I miss tests. :(

dapps/shop/backend/queues/makeOfferProcessor.js
DanielVF: Ah! I see, I was looking up the wrong get config when creating this method. I've removed the async from this one, since it was just decrypting the config it already had from the database.

dapps/shop/backend/queues/makeOfferProcessor.js
DanielVF: The totalPrice is required in the json schema for the offer, so I guess I'll keep it. Changed to zero the amount and use currency 'encrypted'.

dapps/shop/backend/queues/makeOfferProcessor.js
```diff
@@ -0,0 +1,127 @@
+const queues = require('./queues')
+
+const Web3 = require('web3')
```
franckc, line 3: nit: I'd love for us to get rid of Web3 and replace it with ether.js. We don't have to do this in this PR but would be awesome to clean up later... :)

franckc, line 3: I'd love for us to completely get rid of Web3 and replace it with ether.js. We don't have to do this in this PR. I'll create a separate github issue to track this and we can discuss there.

nick: Yea agree we should switch

dapps/shop/backend/routes/_makeOffer.js
```diff
@@ -1,86 +1,23 @@
...
-    )
-    .send({ from: walletAddress, gas: 350000 })
-    .then(tx => {
-      console.log('Make offer:')
-      console.log(tx)
-    })
-    .catch(err => {
-      console.log(err)
-    })
+    { attempts: 6 }
```
mikeshultz, line 19: Why 6?

dapps/shop/backend/queues/makeOfferProcessor.js
```diff
@@ -0,0 +1,127 @@
...
+  queue.resume() // Start if paused
+}
+
+/**
+ * Processes a credit card transaction and submit it to the blockchain.
+ *
+ * job.data should have {shopId, amount, encryptedData}
+ * @param {*} job
+ */
+async function processor(job) {
```
mikeshultz, line 25: Might recommend using `job.progress` to show completion.  It can sometimes help with troubleshooting, though perhaps `job.log` takes care of that here.

dapps/shop/backend/queues/queues.js
```diff
@@ -0,0 +1,30 @@
...
+  new Queue('makeOffer', backendUrl, queueOpts),
+  new Queue('download', backendUrl, queueOpts),
+  new Queue('discord', backendUrl, queueOpts),
+  new Queue(
+    'email',
+    backendUrl,
+    Object.assign(queueOpts, {
+      maxStalledCount: 0 // We don't want to risk sending an email twice
+    })
+  )
```
mikeshultz, line 24: Are these other queues just setup for later?  I don't see anything attaching to them or adding to them.

dapps/shop/backend/routes/_makeOffer.js
DanielVF: I wanted to give it a longer time window to try over. Other than that, it's arbitrary.

dapps/shop/backend/queues/queues.js
DanielVF: Yes. Going to be the next thing I code on this.

dapps/shop/backend/routes/_makeOffer.js
mikeshultz: Not a dealbreakr or antyhing, but you could configure `backoff` too.

nick:
It turns out that the Bull v4 API is different from v3. I think the UI also only works with v3 when you use the alpha build.

Probably easiest to just downgrade to v3 for now?

DanielVF:
Moved back to bull 3, and this thing is actually tested now! Give it a whir.

PR #4468, opened 2020-05-19
Record the IPFS hash of deployed shops

franckc:
### Description:

Add a new table "shop_deployments" to track the IPFS hash when deploying a shop.
In the future we may record other information relavant to a shop deployment in this table.

### Checklist:


dapps/shop/backend/models/shop_deployment.js
```diff
@@ -0,0 +1,19 @@
+module.exports = (sequelize, DataTypes) => {
+  const ShopDeployment = sequelize.define(
+    'ShopDeployment',
+    {
+      shopId: DataTypes.INTEGER,
+      ipfsHash: DataTypes.STRING
```
mikeshultz, line 6: These are defined with an underscore in the migration.  Just checking to make sure this has been tested as working?

dapps/shop/backend/data/migrations/20200519035456-addShopDeployments.js
```diff
@@ -0,0 +1,25 @@
...
+          autoIncrement: true,
+          primaryKey: true
+        },
+        shop_id: {
+          type: Sequelize.INTEGER,
+          references: { model: 'shops', key: 'id' }
+        },
+        ipfs_hash: Sequelize.STRING,
+        created_at: Sequelize.DATE,
+        updated_at: Sequelize.DATE
```
mikeshultz, line 16: Does it make sense to include the configured DNS URL here as well?  I'm not sure if that can/would change from deployment to deployment.  It's currently displayed on that deployment confirm page.

dapps/shop/backend/routes/shops.js
```diff
@@ -303,6 +319,20 @@ module.exports = function(app) {
         allFiles.push(file)
       }
       hash = String(allFiles[allFiles.length - 1].cid)
+    } else {
+      console.log(
+        'Shop not deployed to IPFS: Pinata not configured and not a dev environment.'
+      )
+    }
+    if (hash) {
```
mikeshultz, line 327: Might be worth creating this record even if we don't have an IPFS hash?  Maye to record an error?

dapps/shop/backend/data/migrations/20200519035456-addShopDeployments.js
```diff
@@ -0,0 +1,25 @@
...
+      })
```
nick, line 17: Can we also include user_id so we know who deployed it?

dapps/shop/backend/data/migrations/20200519035456-addShopDeployments.js
nick: Yea I think DNS url makes sense to add. Perhaps also the IPFS gateway the hash is pinned to (eg pinata)

dapps/shop/backend/models/shop_deployment.js
franckc: Yes, sequelize wants you to use camel case in the model. So that in the code, when referring to those columns their name is camel case. Sequelize does the translation when it generates the SQL statement.

And yes I tested it (I actually had a bug initially because I had this column as snake case!).

dapps/shop/backend/routes/shops.js
franckc: Yes, good point. Will be useful anyway to be able to audit the deployments. Will do.

dapps/shop/backend/data/migrations/20200519035456-addShopDeployments.js
franckc: Good idea. Added domain and ipfs_gateway to the shop_deployment table.

dapps/shop/backend/routes/shops.js
franckc: Done

franckc:
Merged in mono repo as PR https://github.com/OriginProtocol/dshop/pull/32

PR #4480, opened 2020-06-03
Add an user stats endpoint

shahthepro:
Part of public facing OGN dashboard that I have been working on. Need the number of investors and team members who have locked up tokens to be displayed in that dashboard. This PR adds an endpoint exactly for that.

Related branch on website repo: https://github.com/OriginProtocol/origin-website/tree/shahthepro/ogn-dashboard


tomlinton:
Looks fine, you mentioned we need to show the number of investors/team that have locked up. this is just the count, do we need to add a query for lockups by distinct user id?

shahthepro:
Thanks @tomlinton. You were right. Have made that change.

infra/token-transfer-server/src/controllers/user.js
```diff
@@ -93,4 +95,12 @@ router.post(
   })
 )
 
+router.get('/user-stats', async (req, res) => {
+  return res.status(200).send({
+    userCount: await Lockup.count({
```
joshfraser, line 100: Does this count the sum total of locked tokens or the # of distinct people who locked up?  I believe we want the former for the dashboard.

shahthepro: @joshfraser Have added that. Also, noticed that I didn't mention `col` along with `distinct`. Have fixed it too.

tomlinton: There is also an additional confirmed column that will need to be true.

shahthepro: Thanks @tomlinton. Have updated it.

shahthepro:
Merging this now. 

@tomlinton @joshfraser can any of you deploy this to prod?

joshfraser:
Deployed. However, let's remove the `userCount` field as there's no reason to leak that information.

PR #4488, opened 2020-07-14
Update secret and charts for dshop backend Sentry integration

franckc:
Added 2 new env variables to the dshop Rinkeby/Mainnet backend:
 - ENVIRONMENT: either "staging" or "prod". This will be logged as metadata in Sentry so that we can differentiate exceptions coming from staging vs prod.
 - SENTRY_DSN: Sentry secret key

See companion PR on the dshop backend: https://github.com/OriginProtocol/dshop/pull/213


devops/kubernetes/charts/origin-experimental/templates/dshop-backend-mainnet.statefulset.yaml
```diff
@@ -27,6 +27,8 @@ spec:
         image: "{{ .Values.containerRegistry }}/{{ .Release.Namespace }}/{{ .Values.dshopBackendMainnetImage }}:{{ .Values.dshopBackendMainnetImageTag }}"
         imagePullPolicy: Always
         env:
+          - name: ENVIRONMENT
```
mikeshultz, line 30: @franckc   Sorry for belated feedback but was just thinking about when to use env vars and how to reduce them.  Why use this at all?  Why not just use `SENTRY_DSN` to figure out if errors should be sent?

franckc:
@mikeshultz ENVIRONMENT is recorded as metadata with the error in Sentry. And then you use it as a filter in Sentry. Which is super handy to differentiate development vs staging vs prod errors.

PR #4503, opened 2020-09-15
Update Dshop's upload request size limit

shahthepro:
GCP has a [default request limit of 1MB](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#configuration-options). This should increase that to 50MB.


franckc (approved):
Good finding! Is 50MB a bit too generous? I would think images would be at most 5MB, so maybe 10MB is a more realistic limit?

Otherwise LGTM

shahthepro:
@franckc Yeah, 10MB sounds reasonable. Have reduced the limit to that

PR #4515, opened 2020-12-08
Stake banner

ZaHuPro:
This PR requires a new environment variable **STAKE_URL**


infra/token-transfer-client/src/constants/index.js
```diff
@@ -13,7 +13,7 @@ if (process.env.NODE_ENV === 'production') {
     apiUrl = process.env.INVESTOR_API_URL
   }
 } else {
-  apiUrl = 'http://localhost:5000'
+  apiUrl = 'http://103.228.153.6:5000'
```
franckc, line 16: This must be your IP? Let's keep localhost since most developers will be running the stack on their local environment. But we could have an override to make it easier to run the API remotely. Something like:
```
apiUrl = process.env.DEV_API_URL || 'http://localhost:5000'
```

ZaHuPro: @franckc  Sorry, My bad. I will update this

ZaHuPro: @franckc I have changed this

franckc: no problem! thanks for fixing it.

franckc (approved):
LGTM

franckc:
@micahalcorn @tomlinton Could you please do a review on these T3 changes before we merge them in?

micahalcorn:
I'll defer to @tomlinton since this is his baby and he knows how to reproduce all of the relevant states. I read through the changes and they seem to address the known product requirements. 👍

tomlinton:
@ZaHuPro are the changes to the package versions and yarn.lock required? If possible we should revert those as I'd prefer not to change any dependencies.

infra/token-transfer-client/src/components/pages/Dashboard.js
```diff
@@ -28,16 +28,11 @@ const Dashboard = props => {
   const displayLockupCta =
     nextVest && data.config.earlyLockupsEnabled && !data.config.isLocked
   const displayFullWidthLockupCta = displayLockupCta && hasLockups
```
tomlinton, line 30: This or the above condition will need to be changed otherwise the CTA will never show. I think one or the other should just be set to `true` depending on if we want the full width CTA or the square box one.

micahalcorn: My understanding was that the square one would be displayed for people who had no existing lock-ups and the full-width banner would display if the square box was displaying existing lock-up information.

tomlinton: Ok then I think just the `data.config.earlyLockupsEnabled` needs to be removed from that conditional.

ZaHuPro:
@tomlinton  The older versions of SASS are not compatible with the latest macOS version

tomlinton (approved):

PR #4548, opened 2021-05-13
Add CORS_WHITELIST for setting CORS in @origin/token-transfer-server

yujie0505:
### Description:

The whitelist of host names for CORS is hard-coded when deploying to the Heroku. So, I need to have another environment variable to set this whitelist without changing anything in the codebase.

### Checklist:

- [ ] ~[Wrap any new text/strings with `fbt`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#wrapping-text) for translation~
- [ ] ~If there are any new/changed strings to translate, [`npm run translate`](https://github.com/OriginProtocol/origin/tree/master/dapps/marketplace/translation#integrating-translations)~

### DIFF
--- a/infra/token-transfer-server/README.md
+++ b/infra/token-transfer-server/README.md
@@ -9,9 +9,10 @@ export SENDGRID_FROM_EMAIL="Origin Protocol <support@shoporigin.com>"
 export SENDGRID_API_KEY="<SendGrid key>"
 export CLIENT_URL"=http://localhost:3000/#"
 export DATABASE_URL="postgres://origin:origin@localhost/origin"
+export CORS_WHITELIST="https://investor.originprotocol.com,https://team.originprotocol.com"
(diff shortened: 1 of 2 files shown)


tomlinton:
I think you can just make this change locally, we don't need to merge it into our codebase?

yujie0505:
Yes, I could just change it locally to resolve this issue. However, I'd like to submit this PR for the following reasons:

1. I was stuck in the CORS error because I thought `CLIENT_URL` would be used for the accepted hosts in the beginning.
2. After reading `README`, I could not find anything else for me to set the accepted hosts, so I needed to trace code to resolve it.
3. I think that if someone in the community is interested in deploying the `t3-server` for himself, then this PR will save his time not to stuck in the same error and to focus on something interesting.

Many thanks for you~

PR #4557, opened 2021-07-30
Add new env var NEW_SHOP_CREATION_DISABLED for dshop

franckc:
We can use that new env var to enable / disable new shop creation from the dshop deployer.

### DIFF
--- a/devops/kubernetes/charts/origin-experimental/templates/dshop-backend-mainnet.statefulset.yaml
+++ b/devops/kubernetes/charts/origin-experimental/templates/dshop-backend-mainnet.statefulset.yaml
@@ -49,6 +49,11 @@ spec:
               secretKeyRef:
                 name: {{ template "dshopBackendMainnet.fullname" . }}
                 key: SENTRY_DSN
+          - name: NEW_SHOP_CREATION_DISABLED
+            valueFrom:
+              secretKeyRef:
+                name: {{ template "dshopBackendMainnet.fullname" . }}
+                key: NEW_SHOP_CREATION_DISABLED
           - name: REDIS_URL
             value: "redis://{{ template "dshopRedisMainnet.fullname" . }}-0.{{ template "dshopRedisMainnet.fullname" . }}.experimental.svc.cluster.local:6379/0"
           - name: DSHOP_CACHE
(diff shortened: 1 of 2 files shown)


devops/kubernetes/charts/origin-experimental/templates/dshop-backend-mainnet.statefulset.yaml
```diff
@@ -49,6 +49,11 @@ spec:
               secretKeyRef:
                 name: {{ template "dshopBackendMainnet.fullname" . }}
                 key: SENTRY_DSN
+          - name: NEW_SHOP_CREATION_DISABLED
+            valueFrom:
+              secretKeyRef:
+                name: {{ template "dshopBackendMainnet.fullname" . }}
+                key: NEW_SHOP_CREATION_DISABLED
```
mikeshultz, line 56: Does this need to be a secret?

franckc: No, does not need to be a secret.

Is there a simpler way to set an env var? oh, perhaps setting it as a GCP build env var?

mikeshultz: There's a few ways.  Could just use the `kubectl` command to temporarily set an env var but we probably want it to be persistent.  You can do use a literal like this:  

https://github.com/OriginProtocol/origin/blob/c5ad992a99d4147d1826aa0f45a4c1f55b105b46/devops/kubernetes/charts/origin/templates/creator.deployment.yaml#L42-L43

Pretty straight forward, but doesn't let the chart be reuable. What you probably want is to just use the plain values files like this:

https://github.com/OriginProtocol/origin/blob/c5ad992a99d4147d1826aa0f45a4c1f55b105b46/devops/kubernetes/charts/origin-experimental/templates/dshop-backend-mainnet.statefulset.yaml#L56-L57

Then just [set the value in the `values.yaml` file for our deployment](https://github.com/OriginProtocol/origin/blob/c5ad992a99d4147d1826aa0f45a4c1f55b105b46/devops/kubernetes/values/origin-experimental/values.yaml#L7).

franckc:
Going for an alternate approach per Mike's suggestion.

